PR #1, opened 2020-10-05
wcc CLI design proposal

brooksmtownsend:
This PR includes the client libraries for the cli utilities and mocks out the usage of `wcc` as a cli tool. This PR is meant to be a discussion space before merging (unless this is a solid mockup of the commands/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,123 @@
 # wcc
 waSCC Controller - A single CLI to handle all of your waSCC tooling needs
+
+## Using wcc
+```
+wcc <subcommand> [args]
+```
+
+## Subcommands
+
+### keys
+Handles key generation and management, previously handled by the `nk` cli
+```
+USAGE:
+    wcc keys gen <keytype>
+
+FLAGS:
+    -h, --help    Prints help information
+
+ARGS:
+    <keytype>    The type of key pair to generate. May be Account, User, Module, Server, Operator, Cluster
+```
+
+### sign
+Handles signing actor modules, previously handled by the `wascap` cli
+
+```
+USAGE:
+    wcc sign [FLAGS] [OPTIONS] <source> <output> --issuer <issuer-key-path> --name <name> --subject <subject-key-path>
+
+FLAGS:
+    -f, --blob_store     Enable access to the blob store capability
+    -e, --events         Enable access to an append-only event stream provider
+    -z, --extras         Enable access to the extras functionality (random nos, guids, etc)
+        --help           Prints help information
+    -h, --http_client    Enable the HTTP client standard capability
+    -s, --http_server    Enable the HTTP server standard capability
+    -k, --keyvalue       Enable the Key/Value Store standard capability
+    -l, --logging        Enable access to logging capability
+    -g, --msg            Enable the Message broker standard capability
+    -p, --prov           Indicates whether the signed module is a capability provider instead of an actor (the default
+                         is actor)
+
+OPTIONS:
+    -c, --cap <capabilities>...         Add custom capabilities
+    -x, --expires <expires-in-days>     Indicates the token expires in the given amount of days. If this option is left
+                                        off, the token will never expire
+    -i, --issuer <issuer-key-path>      Issuer seed key path (usually a .nk file)
+    -n, --name <name>                   A human-readable, descriptive name for the token
+    -b, --nbf <not-before-days>         Period in days that must elapse before this token is valid. If this option is
+                                        left off, the token will be valid immediately
+    -r, --rev <rev>                     Revision number
+    -u, --subject <subject-key-path>    Subject seed key path (usually a .nk file)
+    -t, --tag <tags>...                 A list of arbitrary tags to be embedded in the token
+    -v, --ver <ver>                     Human-readable version string
+
+ARGS:
+    <source>    File to read
+    <output>    Target output file
+```
+
+### caps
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,2 +1,136 @@
...
+FLAGS:
+    -h, --help    Prints help information
+
+ARGS:
+    <keytype>    The type of key pair to generate. May be Account, User, Module, Server, Operator, Cluster
+```
+
+```
+USAGE:
+    wcc keys jwt <keytype>
```
autodidaddict: Maybe calling this sub-command `claims` is a better, more descriptive option? `keys` above makes sense, but feels like conflation to put the claims and keys functionality in the same sub-command.

README.md
```diff
@@ -1,2 +1,136 @@
...
+
+FLAGS:
+    -h, --help    Prints help information
+
+ARGS:
+    <keytype>    The type of jwt to generate. May be Account, Actor, or Operator.
+```
+
+Note: this combines the functionality of `nkeys` and `wascap` for generating keys & generating JWTs. Depending on the needs of the tool, these could potentially be combined to abstract the need for the user to manage the public key & seeds and instead just provide users with JWTs.
```
autodidaddict: Let's avoid mentioning which individual crates are responsible for this functionality.

README.md
```diff
@@ -1,2 +1,136 @@
...
+
+### sign
+Handles signing actor modules, previously handled by the `wascap` cli
```
autodidaddict: See previous comment -  let's not reference which previous CLI tools or which crates we get the functionality from.

README.md
autodidaddict: let's change `jwt` to `token` , so `wcc claims token ....` produces a new token.

Below, let's change the `caps` command to `claims inspect ...` where we pass the filename of a wasm file to inspect.

README.md
```diff
@@ -1,2 +1,136 @@
...
+    -u, --subject <subject-key-path>    Subject seed key path (usually a .nk file)
+    -t, --tag <tags>...                 A list of arbitrary tags to be embedded in the token
+    -v, --ver <ver>                     Human-readable version string
+
+ARGS:
+    <source>    File to read
+    <output>    Target output file
+```
+
+### caps
```
autodidaddict: Let's move `caps` into the `claims` sub-command and call the individual action `inspect`, e.g. `wcc claims inspect ./foo.wasm`

README.md
```diff
@@ -1,2 +1,136 @@
...
+FLAGS:
+    -h, --help    Prints help information
+    -r, --raw     Extract the raw JWT from the file and print to stdout
+
+ARGS:
+    <file>    The file to read
+```
+
+### lattice
+Handles interrogating the lattice, previously handled by the `latticectl` cli
```
autodidaddict: let's drop the references to previous CLIs/crates

autodidaddict (changes requested):

README.md
```diff
@@ -32,10 +31,21 @@ ARGS:
     <keytype>    The type of jwt to generate. May be Account, Actor, or Operator.
```
autodidaddict: `keytype` should probably be something like `token_type` since we're not generating a key here.

README.md
```diff
@@ -32,10 +31,21 @@ ARGS:
...
+
+FLAGS:
+    -h, --help    Prints help information
+    -r, --raw     Extract the raw JWT from the file and print to stdout
+
+ARGS:
+    <file>    The WASM file to inspect
+```
+
+Depending on the needs of the tool, these could potentially be combined to abstract the need for the user to manage the public key & seeds and instead just provide users with JWTs.
```
autodidaddict: We need the ability to create/manage the seed keys because those are required for signing actor modules.

brooksmtownsend: Is there value in providing the `gen` functionality at all if `wcc` is primarily concerned with creating signed tokens? What if we only provide the `claims token` subcommand and hide the keypair gen from the user, outputting keypairs to disk in order to sign subsequent actors, or allowing for an argument for module / account keys for signing?

autodidaddict: If you're going to construct an operator hierarchy, like you would do for Gantry, then you'll need the ability to create freestanding JWTs that aren't being used for embedding. Gantry requires operator JWTs and account JWTs that aren't embedded in actors.

we should have:
`wcc keys gen` (generate a key for Account, Module, Provider (new), Operator
`wcc claims inspect ...`
`wcc claims token` (required for building operator hierarchies in Gantry)
`wcc sign ...`

We need to allow the ability to explicitly generate keys because when constructing robust hierarchies (provenance chains), we support the notion of using multiple signing keys for a single entity. In other words, operators and accounts have a unique public key, but they also have secondary public keys that correspond to private keys... tl;dr a single account or operator can have multiple private keys for signing. Since keys and JWTs aren't always 1:1, we need this flexibility.

autodidaddict (changes requested):

README.md
brooksmtownsend: Makes sense. I think we also can't fully abstract keys away because when signing subsequent wasm files, you will want to re-use your account / operator to do so and not only use your own key. During implementation I want to do some planning and some real-world tests to figure out how we can make this experience better and easier for a user to generate their own keys, but I'll update README accordingly then

autodidaddict (approved):
LGTM

PR #2, opened 2020-10-07
Continued CLI design

brooksmtownsend:
This PR moves the `sign` subcommand under `wcc claims`, as well as specifies more defaults for `wcc claims sign` to enable an easier initial onboarding to signing `wasm` files in wascc projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,11 +132,11 @@ SUBCOMMANDS:
     watch    Watch events on the lattice
 ```
 
-### oci
+### reg
 
 ```
 USAGE:
-    wash oci <SUBCOMMAND> <artifact>
+    wash reg <SUBCOMMAND> <artifact> [FLAGS]
 
 FLAGS:
     -h, --help       Prints help information
@@ -149,4 +149,18 @@ SUBCOMMANDS:
 
 ARGS:
     <artifact>       URI of the artifact
+```
+
+### up
+Starts an interactive REPL session for waSCC development
+```
+USAGE:
+    wash up [FLAGS]
+
+FLAGS:
+    -h, --help       Prints help information
+    -V, --version    Prints version information
+
+SUBCOMMANDS:
+    help        Prints this message or the help of the given subcommand(s)
 ```
--- a/README.md
+++ b/README.md
@@ -42,12 +42,12 @@ OPTIONS:
     -c, --cap <capabilities>...         Add custom capabilities
     -x, --expires <expires-in-days>     Indicates the token expires in the given amount of days. If this option is left
                                         off, the token will never expire
-    -i, --issuer <issuer-key-path>      Issuer seed key path (usually a .nk file). If this option is left off, `wcc` will attempt to locate an issuer key at `$HOME/.wcc/issuer.nk`, and if it is not found then an issuer key will be generated and placed in `$HOME/.wcc/issuer.nk`. You can override this directory by setting the `WCC_HOME` environment variable.
+    -i, --issuer <issuer-key-path>      Issuer seed key path (usually a .nk file). If this option is left off, `wcc` will attempt to locate an issuer key at `$HOME/.wcc/keys/issuer.nk`, and if it is not found then an issuer key will be generated and placed in `$HOME/.wcc/keys/issuer.nk`. You can override this directory by setting the `WCC_HOME` environment variable.
     -n, --name <name>                   A human-readable, descriptive name for the token
     -b, --nbf <not-before-days>         Period in days that must elapse before this token is valid. If this option is
                                         left off, the token will be valid immediately
     -r, --rev <rev>                     Revision number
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,25 +12,59 @@ wcc <subcommand> [args]
...
+    -k, --keyvalue       Enable the Key/Value Store standard capability
+    -l, --logging        Enable access to logging capability
+    -g, --msg            Enable the Message broker standard capability
+    -p, --prov           Indicates whether the signed module is a capability provider instead of an actor (the default is actor)
+
+OPTIONS:
+    -c, --cap <capabilities>...         Add custom capabilities
+    -x, --expires <expires-in-days>     Indicates the token expires in the given amount of days. If this option is left
+                                        off, the token will never expire
+    -i, --issuer <issuer-key-path>      Issuer seed key path (usually a .nk file). If this option is left off, `wcc` will attempt to locate an account key at `$HOME/.wcc/keys/<project_name>_account.nk`, and if it is not found then an issuer key will be generated and placed in `$HOME/.wcc/keys/<project_name>_account.nk`. You can override this directory by setting the `WCC_HOME` environment variable.
```
autodidaddict: I think we decided instead of using the project name it would be the name of the wasm file being signed, with either an account or module suffix

README.md
```diff
@@ -64,6 +98,31 @@ ARGS:
...
+
+```
+USAGE:
+    wcc keys list [OPTIONS]
+
+FLAGS:
+    -h, --help          Prints help information
+
+OPTIONS:
+    -d, --directory <keysdirectory>     The directory where keys are stored for listing. Defaults to `$HOME/.wcc/keys`, and can be overwritten by setting the WCC_HOME environment variable.
```
autodidaddict: If the directory is either the default or something overridden by an environment variable then we probably don't need a separate flag for this option?

autodidaddict (approved):

README.md
brooksmtownsend: Hm, I want the experience to be consistent across the different commands. I think I see two options here for all the commands:

1. Allow the `WCC_HOME` environment variable to override the directory and document accordingly
2. Provide the `-d, --directory` option on all subcommands that would make use of generated keys

Thoughts on either? I think I lean towards 2

autodidaddict: WWKCD? What would kubectl do? I think it uses an environment variable to choose which context you're working with.

How about this:
Use the WCC_HOME directory if it's supplied 
Use HOME/.wcc if that variable isn't set
Override that location if the directory flag is passed.

So its an order if precedence,  with the -d flag trumping all?

?

thomastaylor312: FWIW in Helm we eventually decided to use the [XDG directory spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) as it is fairly standard across all platforms. You can take a look at how we did it in Helm if you are interested

README.md
thomastaylor312: I feel like that could get cluttered real fast. Is it best practice to have a different key for each wasm file?

thomastaylor312 (approved):
I think this arrangement is much better logical grouping. This looks good to me, but please see the comment in response to @autodidaddict about using the XDG spec

autodidaddict:
I'll take a look at the XDG spec.

Also - yes every wasm file requires two keys to sign - one for the issuer and one for the subject. Every wasm module has a globally unique public key. If you signed multiple modules with the same key, they would have the same identity and would appear as the same entity to a host runtime.

README.md
brooksmtownsend: I can see the argument from the security perspective that we wouldn't want one set of keys to sign many different projects... You could always override it with your own account and module keys if you dislike the clutter but the average user wouldn't need to interact with their `~/.wcc` directory if they didn't want to.

README.md
brooksmtownsend: Do we want to mention the `WCC_HOME` env var in the CLI as well? I don't mind having multiple ways to override the directory and the documentation can reference relevant environment variables, but assuming we want `wcc` to be usable in scripting the `-d` flag is nice to have to not deal with environment

brooksmtownsend:
@autodidaddict Do we want to remove the `gantry` section of the CLI specification? We can replace it with a subcommand that interacts with an OCI compliant registry, but I'm not sure what that would look like atm

autodidaddict:
Remove references to gantry and we can maybe just use oci push and pull

brooksmtownsend:
@autodidaddict please review latest changes for wash and OCI

README.md
```diff
@@ -149,3 +131,22 @@ SUBCOMMANDS:
     stop     Tell a given host to terminate the given actor
     watch    Watch events on the lattice
 ```
+
+### oci
+
+```
+USAGE:
+    wash oci <SUBCOMMAND> <artifact>
```
autodidaddict: Maybe this is more friendly as `registry` or `reg` for short. `oci` is a technical distinction that might add learning friction to developers?

autodidaddict (changes requested):

README.md
brooksmtownsend: Good call, renamed

autodidaddict (approved):

PR #3, opened 2020-10-15
Migrate functionality from existing CLIs

brooksmtownsend:
This PR does not introduce additional functionality or convenience to `wash` commands (with the exception of initial implementations of `wash keys get` and `wash keys list`), rather it is focused on migrating existing functionality from previously used CLIs. 

Some feedback that I'm looking for here is around structure of the `wash` CLI in terms of file separation in `src`, any additional feedback for the README as I'm attempting to base all of the structopt messages off of the README, and the readability of `main.rs`. My goal of this CLI is essentially:

1. Receive command
2. Delegate functionality to each module's `handle_command`
3. Execute the command accordingly

This structure will keep the CLI fairly simple, and each module can be updated as needed without changing the tool itself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,12 @@ edition = "2018"
 
 [dependencies]
 structopt = "0.3.18"
-nkeys = "0.0.9" # TODO: Update to 0.0.11 after updating gantryclient
-wascap = "0.5.0"
-gantryclient = "0.1.0"
-latticeclient = "0.3.1"
+env_logger = "0.7.1"
+serde_json = "1.0.59"
+serde = { version = "1.0.117", features = ["derive"] }
+crossbeam = "0.7.3"
+
+nkeys = "0.0.11"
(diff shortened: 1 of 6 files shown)


brooksmtownsend:
I also included a few tests for the `keys` subcommand, which I'd like to expand to the other modules but I wanted to get feedback first before writing too much code.

brooksmtownsend:
Any thoughts here @autodidaddict @thomastaylor312 @billyoung ?

autodidaddict:
> Any thoughts here @autodidaddict @thomastaylor312 @billyoung ?

My only concern is to avoid testing things that we might already have tests for in the other crates. Essentially in the CLI we want to test that the CLI properly defers command line arguments to the supporting libraries...not that the supporting libraries work as intended.

brooksmtownsend:
> My only concern is to avoid testing things that we might already have tests for in the other crates. Essentially in the CLI we want to test that the CLI properly defers command line arguments to the supporting libraries...not that the supporting libraries work as intended.

Fair point. I'll focus on creating tests for new functionality introduced here rather than executing the logic of supporting libraries.

PR #4, opened 2020-10-21
wash par initial implementation

brooksmtownsend:
Due to a change in my default rust formatter for vscode, this PR looks a lot busier than it actually is.

Changes:
- Pinned dependencies in Cargo.toml to remove git references
- renamed WASH_HOME environment variable to WASH_KEYS to better describe key location
- Added ability for `wash claims token` to generate a `Service` JWT (aka Capability Provider)
- Added `extract_keypair` to keys module, which is responsible for taking a user input key (either file location or seed string) and returning the KeyPair object for it. Also is responsible for looking in $WASH_KEYS directory for the file location, and finally generating a key for the user if not found
- Implemented `par` subcommand, including `create`, `insert` and `inspect` functionality

You can install this version of the `wash` binary by running:
```
cargo install --git https://github.com/brooksmtownsend/wash --branch par-command
```


autodidaddict:
This build is failing... looks like the errors aren't `Sync` and `Send`

src/keys.rs
```diff
@@ -154,21 +155,94 @@ pub fn list(directory: Option<String>) {
...
+            }
+        }
     }
 }
 
 fn determine_directory(directory: Option<String>) -> String {
     match directory {
         Some(d) => d,
-        None => format!("{}/.wash/keys", env::var("HOME").unwrap())
+        None => format!("{}/.wash/keys", env::var("HOME").unwrap()),
```
autodidaddict, line 166: The documentation indicates that there's a `WASH_KEYS` environment variable.. shouldn't we be looking there instead of `HOME/.wash/keys` ? Or maybe I'm not understanding the flow

autodidaddict: Or is `HOME/.wash/keys` the fallback if there is no `WASH_KEYS` var?

src/keys.rs
```diff
@@ -154,21 +155,94 @@ pub fn list(directory: Option<String>) {
...
+            // Default key found
+            Ok(mut f) => {
+                let mut s = String::new();
+                f.read_to_string(&mut s)?;
+                s
+            }
+            // No default key, generating for user
+            Err(_e) => {
+                let kp = KeyPair::new(keypair_type);
+                println!("No keypair found in {}, we will generate one for you and place it there. If you'd like to use your own keys, you can supply them as a flag.", path);
```
autodidaddict: maybe replace `your own` with `alternative` - the keys being used are the user's all the time.. it's just a matter of overriding the default keys

src/keys.rs
brooksmtownsend: Yes, since `structopt` supports the environment variable argument, if a directory is not provided then it will default to the value of `$WASH_KEYS`, if it is defined

autodidaddict:
This is good stuff!

src/keys.rs
brooksmtownsend: Yeah this wording felt a little forced, I'm a fan of this change

autodidaddict (approved):

PR #9, opened 2020-10-29
Support PAR compression and interacting with compressed parJEEzy files

### LINKED ISSUE #7: Support parGEEzy compression
Currently the `create`, `insert` and `inspect` subcommands of `wash` only operate on uncompressed `.par` files. Using the `provider-archive` crate at version `0.2.0` introduces 2 new functions, `compress` and `decompress`, that should allow `wash` to interface with compressed provider archives.

My imagined feature list
1. `create` should have a boolean flag that enables the creation of a compressed `.par.gz` file rather than an uncompressed `.par`
2. `insert` should be able to decompress a parGEEzy, insert another library into the `par`, and output a modified compressed file
3. `inspect` should be able to print metadata information about a compressed file, same as uncompressed, without modifying the file itself

brooksmtownsend:
This PR includes the ability for the `create` command to also create a `parJEEzy` file alongside the `par` file, as well as allowing `insert` and `inspect` to directly interact with compressed `par` files.

I ended up copying logic from `provider-archive` because to me it did not make sense to export the functionality from `provider-archive` to see if an archive is compressed or not. 

Fixes #7


ChrisRx (approved):
lgtm

billyoung (approved):
looks good!

PR #11, opened 2020-11-03
Claims autogeneration, optional autogeneration

### LINKED ISSUE #8: Autogeneration of keys should be optional
Currently, the `keys` module will follow a few steps in order to locate a key that is supplied on the command line:

1. Attempt to open a file at the `input` location (e.g. user provided `/path/to/issuer.nk`)
2. Use input as a seed key (e.g. user provided `SAASDAHSDKJAHSDJKAHSD`)
3. Look in the directory provided as a CLI flag for `directory/modulename_moduletype.nk` e.g. `~/tmp/httpsrver_account.nk`
4. Look in the directory provided as the environment variable `$WASH_KEYS` (same as above, different directory)
5. **Otherwise** the `keys` module will generate a key for you and store it in the directory supplied in step 3 or 4 (respectively)

This autogeneration is a fallback, but it always occurs and is not optional if no keys are found / given. In an attempt to locate keys, the user should be able to pass a boolean flag e.g. `--no-generate` to prevent this step. In the event that `wash` is being used in production, or in a CI/CD step, an autogenerated key will allow for the process to continue without finding a key, but in the case of a failure to find the production keys it could cause issues with actor / capability provider validation when attempting to deploy.

brooksmtownsend:
Fixes #8 

This PR enables autogeneration of keys when signing a wascc actor by using the `claims` subcommand to `wash`.

To sign an actor using `wash`, you simply need to give the `wasm` file to sign, any desired capabilities, and a human readable name of the actor.

```
wash claims sign ./target/wasm32-unknown-unknown/release/myactor.wasm -s -h --name "My Actor"
```

This will result in the following output, assuming that the user does not have keys created already for the `myactor` module stored in `$WASH_KEYS`.
```
No keypair found in "/Users/btownsend/.wash/keys/myactor_account.nk".
We will generate one for you and place it there.
If you'd like to use alternative keys, you can supply them as a flag.

No keypair found in "/Users/btownsend/.wash/keys/myactor_module.nk".
We will generate one for you and place it there.
If you'd like to use alternative keys, you can supply them as a flag.

Successfully signed ./target/wasm32-unknown-unknown/release/kvcounter_s.wasm with capabilities: wascc:http_client,wascc:http_server
```

On subsequent runs of `sign`, `wash` will attempt to locate the keys (if not overridden by command line args) and re-use the same keys for the same module. This does mean that `wash` won't handle the case where a user has multiple actors with the same name.

We are also now allowing the user to supply the `--disable-keygen` flag, which disables autogeneration of keys. This is useful in situations where a key _should_ be found, and if it's not then it's a logical error and the app should not run.
```
➜ wash claims sign ./target/wasm32-unknown-unknown/release/myactor.wasm -s -h --name "My Actor" --disable-keygen
Error: No keypair found in /Users/btownsend/.wash/keys/myactor_account.nk, please ensure key exists or supply one as a flag
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash"
-version = "0.1.3"
+version = "0.1.4"
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
 edition = "2018"
 repository = "https://github.com/wascc/wash"
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -73,15 +73,14 @@ struct SignCommand {
     #[structopt(short = "o", long = "output")]
     output: Option<String>,
 
-    /// Location of key files for signing. Defaults to $WASH_KEYS ($HOME/.wash/keys)
-    #[structopt(
-        short = "d",
-        long = "directory",
-        env = "WASH_KEYS",
-        hide_env_values = true
-    )]
-    directory: Option<String>,
-
+    // /// Location of key files for signing. Defaults to $WASH_KEYS ($HOME/.wash/keys)
+    // #[structopt(
+    //     short = "d",
+    //     long = "directory",
+    //     env = "WASH_KEYS",
+    //     hide_env_values = true
+    // )]
+    // directory: Option<String>,
     /// Disables autogeneration of signing keys
     #[structopt(long = "disable-keygen")]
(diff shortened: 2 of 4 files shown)


billyoung (approved):
looks good!

PR #14, opened 2020-11-10
Add env var input for par subject/issuer

### LINKED ISSUE #12: Support environment variables for signing keys
Currently `wash` supports supplying the seeds for signing keys via command line argument, and it can be supplied either as the seed string (e.g. "SAAK6QHHYMNRDYHYHBFBX4JK5Y5JIXROP5LBDKJMP3IK32W44KXJ43OBRE") or as a file (e.g. `issuer.nk`)

Supplying the seed as a string was meant to cover the scenario where you would like to avoid placing your seed in a file and supply it as an environment variable instead, however if the shell history is not cleaned properly the seed will be viewable in plaintext. 

For commands that request seed keys (e.g. `wash claims token`, `wash par create`, etc) these should have environment variable defaults defined via structopt.

ChrisRx:
Fixes #12

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.1.6"
+version = "0.1.7"
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
 edition = "2018"
 repository = "https://github.com/wascc/wash"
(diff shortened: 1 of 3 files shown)


brooksmtownsend (approved):
LGTM, I signed a `par` using this and environment variables and it worked great (no creds shown in bash history)

Side note, a cool use case for this with autogeneration is, as a developer, we can export our `WASH_ISSUER_KEY` (so our account) and autogeneration can create individual module keys for us for each project that we have. Bazinga!

brooksmtownsend:
Actually @ChrisRx, if you can merge in `main` and bump the semver to `0.1.7` that would be great before you merge

PR #20, opened 2020-11-19
Initial Implementation of OCI push and pull commands

### LINKED ISSUE #19: Support oci registry commands
For our parJEEzy and actor modules, we should be able to `push` and `pull` them to/from OCI compliant registries. The functionality to do this will be encapsulated in the [OCI Distribution crate](https://github.com/deislabs/krustlet/tree/master/crates/oci-distribution).

brooksmtownsend:
This currently depends on a relative path to the `oci-distribution` crate as the `push` PR has not been merged and released yet.

This PR contains an implementation of `wash reg push` and `wash reg pull`, two commands for interacting with OCI compliant registries. As this tool is the `waSCC Shell`, I have restricted `push` and `pull` operations to only work for `waSCC` provider archives (with claims) and `waSCC` signed actors. You can read more about this decision in the proposed PR to our ADR [here](https://github.com/wascc/adr/pull/1/files#diff-caa31b5aee04ae6b39babdc1381aa6fff44281e23408d99838ecb5ac819644f7).

You can test pulling an artifact simply with:
```
wash reg pull wascc.azurecr.io/test/hello:v1
```
This syntax for the tag supports image tag, digest, or both. `latest` is not allowed unless you supply the `--allow-latest` flag.

You can test pushing an artifact (as long as you have credentials) with:
```
export WASH_REG_USER=<username>
export WASH_REG_PASSWORD=<password>
wash reg push wascc.azurecr.io/test/hello:v2 ./hello_signed.wasm
```

Tests forthcoming, I wanted to publish this PR for initial comments before I spent too much time there.

Fixes #19

🚿

### DIFF
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -1,0 +1,348 @@
+extern crate oci_distribution;
+use oci_distribution::client::*;
+use oci_distribution::secrets::RegistryAuth;
+use oci_distribution::Reference;
+use provider_archive::ProviderArchive;
+use spinners::{Spinner, Spinners};
+use std::fs::File;
+use std::io::prelude::*;
+use structopt::clap::AppSettings;
+use structopt::StructOpt;
+use tokio::runtime::*;
+
+const PROVIDER_ARCHIVE_MEDIA_TYPE: &str = "application/vnd.wascc.provider.archive.layer.v1+par";
+const PROVIDER_ARCHIVE_CONFIG_MEDIA_TYPE: &str = "application/vnd.wascc.provider.archive.config";
+const PROVIDER_ARCHIVE_FILE_EXTENSION: &str = ".par.gz";
+const WASM_MEDIA_TYPE: &str = "application/vnd.module.wasm.content.layer.v1+wasm";
+const WASM_CONFIG_MEDIA_TYPE: &str = "application/vnd.wascc.actor.archive.config";
+const WASM_FILE_EXTENSION: &str = ".wasm";
+
+const SHOWER_EMOJI: &str = "\u{1F6BF}";
+
+enum SupportedArtifacts {
+    Par,
+    Wasm,
+}
+
+#[derive(Debug, StructOpt, Clone)]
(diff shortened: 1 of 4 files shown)


src/reg.rs
```diff
@@ -0,0 +1,348 @@
...
+    url: String,
+
+    /// OCI username, if omitted anonymous authentication will be used
+    #[structopt(
+        short = "u",
+        long = "user",
+        env = "WASH_REG_USER",
+        hide_env_values = true
+    )]
+    user: Option<String>,
```
autodidaddict, line 42: If you move the user and password, insecure, and allow latest up to the `RegCliCommand` struct then they'll show up as options for `reg` rather than having to duplicate them for each of the reg commands. It's not a huge deal, but it makes the help output and bad syntax error messages a little easier to read, and saves you a little on destructuring.

src/reg.rs
```diff
@@ -0,0 +1,348 @@
...
+        &config_buf,
+        config_media_type,
+        &auth,
+        None,
+    ))?;
+
+    sp.stop();
+    println!(
+        "\n{} Successfully validated and pushed to {}",
+        SHOWER_EMOJI,
```
autodidaddict, line 331: :heart:

autodidaddict (changes requested):
Just one suggested change to extract the common options from push and pull and move them up to `reg`

brooksmtownsend:
Due to the dependency on `krustlet`, marking this as `do not merge` until the oci-distribution crate has the `push` functionality released. If we merge before then, users will not be able to install wash via `cargo install wash-cli` due to the relative dependency

brooksmtownsend:
@autodidaddict @billyoung @ChrisRx Any other remaining concerns? I'd like to have `wash` pointing to the main krustlet repo, but as `wash up` and other issues get fixed I can see merge conflicts getting harder to diagnose, so I'm settling for pointing to my `krustlet` repo for now

PR #21, opened 2020-11-20
OS Packaging (deb/rpm)

billyoung:
This adds our "MVP" of Deb/RPM packaging.

### DIFF
--- a/build/nfpm.arm64.yaml
+++ b/build/nfpm.arm64.yaml
@@ -1,0 +1,18 @@
+# NFPM: check https://nfpm.goreleaser.com/configuration for detailed usage
+# wash cli [amd64]
+
+name: wash
+arch: arm64
+platform: linux
+version: ${VERSION}
+section: default
+priority: extra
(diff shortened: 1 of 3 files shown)


autodidaddict (approved):

build/nfpm.arm64.yaml
```diff
@@ -0,0 +1,18 @@
+# NFPM: check https://nfpm.goreleaser.com/configuration for detailed usage
+# wash cli [amd64]
```
brooksmtownsend: Not sure if this comment is just for information, or if it's functional, but it should be `wash cli [arm64]` here right?

brooksmtownsend (changes requested):
One change, otherwise LGTM!

build/nfpm.arm64.yaml
billyoung: just information. thanks for catching that. i'll update

brooksmtownsend (approved):

PR #27, opened 2020-12-03
Initial ctl implementation

brooksmtownsend:
This PR implements the `ctl` subcommand which interfaces with the wasmcloud control interface.

The `ctl` subcommand itself has its own subcommands, `get`, `start` and `stop`. The full list of commands is below:
`wash ctl get hosts`
`wash ctl get inventory`
`wash ctl get claims`
`wash ctl start actor`
`wash ctl start provider`
`wash ctl stop actor`
`wash ctl stop provider`

I've tested that these commands work by commenting out the calls to the command interface in this test: https://github.com/wascc/wasmCloud/blob/main/tests/control.rs and instead manually starting / stopping actors and providers while the test waits for the action to complete. 

This PR also does the lifting of renaming `wascc` to `wasmcloud` around the `wash` project, and created a `util` module for helper functions that are shared across the cli.

For `wash ctl start {actor,provider}`, if a host is not provided then an auction will kick off for the actor/provider, then it will be scheduled on the first suitable host

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "wash-cli"
-version = "0.1.8"
+version = "0.1.9"
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
 edition = "2018"
-repository = "https://github.com/wascc/wash"
-description = "waSCC Shell (wash) CLI tool"
+repository = "https://github.com/wasmcloud/wash"
+description = "wasmCloud Shell (wash) CLI tool"
 license = "Apache-2.0"
 readme = "README.md"
-keywords = ["webassembly", "wascc", "wash"]
+keywords = ["webassembly", "wasmCloud", "wasmcloud", "wash"]
 categories = ["wasm", "command-line-utilities"]
 
 [badges]
@@ -21,6 +21,7 @@ serde = { version = "1.0.117", features = ["derive"] }
 crossbeam = "0.7.3"
 tokio = { version = "0.2.0" }
 spinners = "1.2.0"
+nats = "0.8.6"
 
 nkeys = "0.0.11"
 latticeclient = "0.4.0"
@@ -28,6 +29,7 @@ wascap = "0.5.1"
 term-table = "1.3.0"
 provider-archive = "0.3.0"
 oci-distribution = { git = "https://github.com/brooksmtownsend/krustlet", branch = "oci-push" }
+control-interface = { git = "https://github.com/wascc/wasmCloud", branch = "main"}
 
 [[bin]]
 name = "wash"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ wascap = "0.5.1"
 term-table = "1.3.0"
 provider-archive = "0.3.0"
 oci-distribution = { git = "https://github.com/brooksmtownsend/krustlet", branch = "oci-push" }
-control-interface = { git = "https://github.com/wascc/wasmCloud", branch = "main"}
+control-interface = { git = "https://github.com/wasmcloud/wasmcloud", branch = "main"}
 
 [[bin]]
 name = "wash"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 name = "wash-cli"
-version = "0.1.9"
+version = "0.1.11"
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
 edition = "2018"
-repository = "https://github.com/wascc/wash"
-description = "waSCC Shell (wash) CLI tool"
+repository = "https://github.com/wasmcloud/wash"
+description = "wasmCloud Shell (wash) CLI tool"
 license = "Apache-2.0"
 readme = "README.md"
-keywords = ["webassembly", "wascc", "wash"]
+keywords = ["webassembly", "wasmCloud", "wasmcloud", "wash"]
```
autodidaddict: Pretty sure we shouldn't be using the same keyword twice. I think the rust convention is to use all lower-case and searching by keyword on crates[dot]io is case-insensitive.

src/ctl.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    /// RPC Port for connections, defaults to 4222 for local nats
+    #[structopt(short = "p", long = "rpc-port", default_value = "4222")]
+    rpc_port: String,
+
+    /// Namespace prefix for wasmCloud command interface
+    #[structopt(short = "n", long = "ns-prefix")]
+    ns_prefix: Option<String>,
+
+    /// Timeout length for RPC, defaults to 1 second
```
autodidaddict: RPC calls should probably have a slightly higher default timeout. The "scatter gather" options from the control interface take their own independent timeouts because they potentially need to be higher than the single request-response calls.

src/ctl.rs
```diff
@@ -0,0 +1,461 @@
...
+    Stop(StopCommand),
+}
+
+#[derive(Debug, Clone, StructOpt)]
+pub enum GetCommand {
+    /// Query lattice for running hosts
+    #[structopt(name = "hosts")]
+    Hosts(GetHostsCommand),
+
+    /// Query lattice for the inventory of a host
```
autodidaddict: Maybe rephrase this.. it's not querying the entire lattice, it's actually probing a single host for its inventory. Maybe `Query a single host for its inventory` ?

src/ctl.rs
```diff
@@ -0,0 +1,461 @@
...
+    #[structopt(name = "inventory")]
+    HostInventory(GetHostInventoryCommand),
+
+    /// Query lattice for claims present in the lattice
```
autodidaddict: This is a "first responder" query. It asks the entire lattice for the claims cache and the first host that responds satisfies the query.

src/ctl.rs
```diff
@@ -0,0 +1,461 @@
...
+
+pub async fn start_provider(cmd: StartProviderCommand) -> Result<StartProviderAck> {
+    let client = client_from_opts(cmd.opts).await?;
+    client
+        .start_provider(&cmd.host_id, &cmd.provider_ref, cmd.link_name)
+        .await
+        .map_err(convert_error)
+}
+
+pub async fn stop_provider(cmd: StopProviderCommand) -> Result<StopProviderAck> {
```
autodidaddict, line 343: :smile: you were right...these control interface consumer calls are super crisp.

autodidaddict (changes requested):

src/ctl.rs
brooksmtownsend: Do you think 5 seconds is a reasonable timeout length? I'm up for making this as big as we want, though if it gets much longer I may pull in some of the spinner code to show that we're working on it

brooksmtownsend:
Marking as `do not merge` as I'm working to add in the new auction functionality to the `ctl` command before merging

src/ctl.rs
autodidaddict: I think 2 is fine, 3 max. It's a configurable option, so it's just a matter of making the default reasonable.

autodidaddict:
> Marking as `do not merge` as I'm working to add in the new auction functionality to the `ctl` command before merging

When doing an "auction", that should probably just be included in the "start" functionality. For example, if you do `start actor bob` with no further parameters, it should try and auction for a suitable host for bob. If you supply a host name, then it can skip the auction. Similar with a provider, `start provider alfred` with no supplied host name can simply do an auction for a host name and then perform the schedule directly.

Put another way, the 2-phase process of auction-and-start is an implementation detail that we don't need to inflict on our users as additional complexity/friction.

PR #30, opened 2020-12-07
claims and par improvements

### LINKED ISSUE #17: Auto-generated keys to use a single account key for a user
The current behavior for auto-generated keys is to create a new account/service key for every provider added to a provider archive.  I am proposing to instead only generate a single account key for a user, as this aligns more with typical usage of signing keys in wascc.  So this would result in something like `$HOME/.wash/keys/account.nk` vs. `$HOME/.wash/keys/libwascc_fs_account.nk` (with a new one for each capability provider).

### LINKED ISSUE #23: Par Inspect should support OCI URI
Currently `wash par inspect` only works with a relative/absolute filepath to a provider archive on disk. `inspect` should also support an OCI URI as an argument, which will then pull + inspect the par (without persisting on disk, potentially unless a flag is toggled like `--store`?). That way we can verify the claims of a remote parJEEzy without downloading it first.

### LINKED ISSUE #26: claims token only supports seed files for operators, accounts and providers
When generating actor JWTs with `wash claims`, it utilizes the function `extract_keypair` which will attempt to source the keypair from a command line literal, filepath, or environment variable.

The functions for generating operator, account and provider JWTs still use old logic and not the `extract_keypair` function, so those JWTs need seed keys in a file on disk in order to generate the JWT. For uniform functionality, and in the interest of scripting, all of these JWTs should be able to be generated without a seed key on disk, and without exposing a seed key in the bash history.

brooksmtownsend:
Fixes #17 , Fixes #23 , Fixes #26 

Multi-issue fix PR for the three above issues.

- [x] Autogenerated keys now create one account key per-user, and re-use that account key for signing
- [x] `wash par inspect` now supports an OCI URL along with the already supported file path
- [x] `wash claims token` can now generate JWTs for account, operator, provider and actors with autogenerated signing keys (unless the `--disable-keygen` flag is supplied)

### DIFF
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -19,7 +19,7 @@ use nkeys::{KeyPair, KeyPairType};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::fs::{read_to_string, File};
+use std::fs::File;
 use std::io::Read;
 use std::io::Write;
 use std::path::PathBuf;
@@ -73,18 +73,6 @@ struct SignCommand {
     #[structopt(short = "o", long = "output")]
     output: Option<String>,
 
-    // /// Location of key files for signing. Defaults to $WASH_KEYS ($HOME/.wash/keys)
-    // #[structopt(
-    //     short = "d",
-    //     long = "directory",
-    //     env = "WASH_KEYS",
-    //     hide_env_values = true
-    // )]
-    // directory: Option<String>,
-    /// Disables autogeneration of signing keys
-    #[structopt(long = "disable-keygen")]
-    disable_keygen: bool,
-
     #[structopt(flatten)]
     metadata: ActorMetadata,
 }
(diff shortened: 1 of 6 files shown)


src/claims.rs
```diff
@@ -137,9 +107,14 @@ struct GenerateCommon {
     /// Indicates the token expires in the given amount of days. If this option is left off, the token will never expire
     #[structopt(short = "x", long = "expires")]
     expires_in_days: Option<u64>,
+
     /// Period in days that must elapse before this token is valid. If this option is left off, the token will be valid immediately
     #[structopt(short = "b", long = "nbf")]
     not_before_days: Option<u64>,
+
+    /// Disables autogeneration of keys if seed(s) are not provided
```
autodidaddict, line 115: I like this. We should mention that this option is most useful in build pipelines so you don't accidentally generate new keys/require existing keys

autodidaddict: This should default to false (I think structopt does that anyway?) , but again the most common developer path should be the easiest/default.

src/claims.rs
```diff
@@ -148,17 +123,18 @@ struct OperatorMetadata {
     #[structopt(short = "n", long = "name")]
     name: String,
 
-    /// Seed key paths (first seed establishes self-signed identity, others are used for optional valid signers list)
-    #[structopt(short = "s", long = "seed", name = "seed-path")]
-    key_paths: Vec<String>,
+    /// Self signing operator key
+    /// Can either be seed value or path to seed file
```
autodidaddict: We should recommend in the documentation that in build pipelines (especially public) seed values not be passed clear text on the command line, and are either injected via environment variable or use a file.

autodidaddict (approved):
Loving this. We may find usability issues as we move forward and use it in the field and for demos, but this all looks to be on the right track so far.

src/claims.rs
brooksmtownsend: Yeah I think some samples geared towards a build pipeline will be great to illustrate this

src/claims.rs
brooksmtownsend: With a few fixes from my last commit, you can now always pass in the required keys via an environment variable, and any keys through a command line arg as a file. The only option we don't support is supplying additional signers with environment variables, it would need to be provided as an option. That might be something we revisit later if we are designing a key hierarchy that includes additional valid signers

brooksmtownsend: e.g.
```
WASH_ISSUER_KEY=/path/to/operator.nk
# self-sign issuer sourced from env
wash claims token operator --name my_operator --disable-keygen
```

brooksmtownsend: Alternatively, `WASH_ISSUER_KEY=SOA.........` if you don't have bash tracing on and don't want to store on disk

PR #43, opened 2020-12-17
Implemented remote claims inspect

### LINKED ISSUE #34: claims inspect should support OCI link
Just as `wash par inspect` supports an OCI URL for inspecting remote provider archive files, `wash claims inspect` should support an OCI URL for inspecting remote actor modules. In the `par` module, we solve this problem by attempting to open a file with that name, and if it does not exist then we use `crate::reg::pull_artifact` to load the archive into memory before inspecting its claims.

Current functionality:
```
➜ wash claims inspect wascc.azurecr.io/kvcounter:v0.1.1
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/claims.rs:595:39
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Panic in Arbiter thread.
```

Desired functionality:
```
➜ wash claims inspect wascc.azurecr.io/kvcounter:v0.1.1
╔════════════════════════════════════════════════════════════════════════╗
║                           KVCounter - Module                           ║
╠═════════════╦══════════════════════════════════════════════════════════╣
║ Account     ║ ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW ║
╠═════════════╬══════════════════════════════════════════════════════════╣
║ Module      ║ MC6Z57M3HOUPIAMTZDMKFBJ5AZ3UOWZ4WJYO7DZNA6567XSJNBUQP55I ║
╠═════════════╬══════════════════════════════════════════════════════════╣
║ Expires     ║                                                    never ║
╠═════════════╬══════════════════════════════════════════════════════════╣
║ Can Be Used ║                                              immediately ║
╠═════════════╬══════════════════════════════════════════════════════════╣
║ Version     ║                                               v0.1.1 (1) ║
╠═════════════╩══════════════════════════════════════════════════════════╣
║                              Capabilities                              ║
╠════════════════════════════════════════════════════════════════════════╣
║ K/V Store                                                              ║
║ HTTP Server                                                            ║
╠════════════════════════════════════════════════════════════════════════╣
║                                  Tags                                  ║
╠════════════════════════════════════════════════════════════════════════╣
║ None                                                                   ║
╚════════════════════════════════════════════════════════════════════════╝
```

brooksmtownsend:
Fixes #34

### DIFF
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -49,13 +49,7 @@ pub struct ClaimsCli {
 enum ClaimsCliCommand {
     /// Examine the capabilities of a WebAssembly module
     #[structopt(name = "inspect")]
-    Inspect {
-        /// The file to read
-        file: String,
-        /// Extract the raw JWT from the file and print to stdout
-        #[structopt(name = "raw", short = "r", long = "raw")]
-        raw: bool,
-    },
+    Inspect(InspectCommand),
     /// Sign a WebAssembly module, specifying capabilities and other claims
     /// including expiration, tags, and additional metadata
     #[structopt(name = "sign")]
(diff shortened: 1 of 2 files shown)


src/claims.rs
```diff
@@ -591,16 +624,31 @@ fn sign_file(cmd: &SignCommand) -> Result<(), Box<dyn ::std::error::Error>> {
     }
 }
 
-fn render_caps(file: &str, raw: bool) -> Result<(), Box<dyn ::std::error::Error>> {
-    let mut wfile = File::open(&file).unwrap();
-    let mut buf = Vec::new();
-    wfile.read_to_end(&mut buf).unwrap();
+async fn render_caps(cmd: InspectCommand) -> Result<(), Box<dyn ::std::error::Error>> {
```
autodidaddict, line 627: I like this - it's consistent with everywhere else in the new wasmCloud runtime where we check if a ref is a file, and if it isn't, we attempt to treat it as an OCI reference.

autodidaddict (approved):

src/claims.rs
brooksmtownsend: If it ends up occurring again elsewhere in `wash`, I think I'll extract some `bytes_for_reference` function into the `util` crate to prevent rewriting it again 😄

Issue #35, opened 2020-12-15
Catch panics, errors and printlns during REPL execution


brooksmtownsend:
During REPL execution, a panic in a submodule can corrupt the UI of the repl (by constantly showing an error message across the screen). These panics should ideally be logged as errors to the log window and be handled, potentially via something like https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/controlling-panics-with-std-panic.html, so that REPL execution is not interrupted.

One common panic that has affected me in testing the REPL is binding the `http_server` capability provider on a port that is already in use. This is a common mistake and should be something we expect from REPL execution.

brooksmtownsend:
Errors, panics, and `println!` macro calls all display according to the current position rather than the cursor, so those should be minimized in favor of using the appropriate `log`, `error`, `warn` etc calls. We should examine common usages of output throughout the `wash` project and standardize them for parity with the REPL output strategy

brooksmtownsend:
This issue is fairly broad, and other than due diligence to gracefully catch errors we can't catch every error (even unwinding panics is not guaranteed to work)

PR #48, opened 2020-12-25
support claims related commands in repl

tiptop96:


tiptop96:
Here is an attempt at implementing the claims related commands in the REPL. What do you think @brooksmtownsend?

I had a little bit of trouble testing some of the `token` commands, they ended up pancining due to stack overflow when running from cli (`cargo run -- claims token actor --name test -s ../hellohttp/module.nk -i ../hellohttp/account.nk`) and from REPL (`token actor --name test -s ../hellohttp/module.nk -i ../hellohttp/account.nk`). I tired playing around with the params for the command and reading the code, but I have yet to get it too work. Any guidance here would be much appreciated.

Also let me know if you think a draft pull request works for checking in. 👍

brooksmtownsend:
Thanks for the PR @tiptop96! I'm still out on vacation ATM but this is at the top of my backlog for when I get back 🙂 Happy New Years 🎉

tiptop96:
Awesome! Happy new years 🥳

brooksmtownsend:
Three things that I notice right away:
- The commands list is getting long, with just the `ctl` commands it seems okay to have the REPL have a flat list of subcommands, but with the others we may end up needing to hide them behind their respective subcommands like we do for the overarching CLI. e.g. I think we should do `wash> claims inspect` rather than `wash> inspect` to keep the help text brief
- JSON output in the output window might not be count multilines appropriately, so if JSON is at the bottom of the output it doesn't scroll to show all output.
- Errors in command execution aren't displayed in the Output window, unfortunately they end up being displayed inline with the input. I've attached a picture below, I think this is related to #35. I think we should solve that issue before we end up migrating the commands, or we should examine error points throughout `wash` and handle them appropriately rather than bubbling up an error to structopt to display (which works fine for a cli).
image

brooksmtownsend:
Something I completely forgot to mention in my other comment @tiptop96 , the `claims` commands work great for me in the REPL 🚀  Great work there

tiptop96:
Thanks for the feedback! 👍 I will look at this ASAP.

brooksmtownsend:
I have noticed that the `extract_keypair` function sends `println`s even if we're in the REPL, so it seems that not only error messages affect output but also printlns (code here https://github.com/wasmCloud/wash/blob/main/src/keys.rs#L155). I updated this issue https://github.com/wasmCloud/wash/issues/35 to reflect our need to change these `println!` calls, so don't worry about it in this PR, but it might come up in your testing

tiptop96:
Thanks for the heads up, I had missed that.

The par commit `c6bb937` changes some logic for `create`, I hope for the better but I wanted to check anyways. Previously even in case of an error when writing the archive file: https://github.com/wasmCloud/wash/blob/13f0c9d5b9bf887830ebc5eda210f0de1a415d35/src/par.rs#L255-L258
This would still run and print "success": https://github.com/wasmCloud/wash/blob/13f0c9d5b9bf887830ebc5eda210f0de1a415d35/src/par.rs#L261-L268
You should be able reproduce this by passing something like `--destination some_non_existing_folder/xyz`.

This was changed and it prints only one of the two, is that desired?

tiptop96:
I noticed that json output doesn't work for the `ctl` commands in the `REPL`, should I add this as a part of this PR? Or maybe I have been going down the wrong path by including it in the others and should remove it? 😆

brooksmtownsend:
> This was changed and it prints only one of the two, is that desired?

Good catch here, this should definitely be an `if/else` statement rather than printing both in the case of an error. While you're in there feel free to make that change 😄

brooksmtownsend:
>I noticed that json output doesn't work for the ctl commands in the REPL, should I add this as a part of this PR? Or maybe I have been going down the wrong path by including it in the others and should remove it? 😆

Good point, I see now that the `json` output logic is actually contained in the `ctl` module, so it's not accessible by the REPL. Go ahead and file an issue for this so we can keep this PR focused, but that will just need a refactor of the `ctl` module 👍

tiptop96:
Awesome the `if/else` is already fixed by `c6bb937`. ✅

I'll do that. 👍

tiptop96:
I can run all commands, there are a few caveats but they all seem to happen with the CLI and therefore are not related to this issue.

The only thing is I can’t assert is if `par insert` is working. I figured `inspect` would look different after an `insert`. Please take a look if you don't mind @brooksmtownsend.

I tried to reuse a lot of code but ended up with some repetition still (mainly for `reg`). My strategy was basically to try and refactor "display" specific logic to each modules `handle_command`, letting it's callees return `String`s instead of printing them. However this is currently not the case for `reg` and `ctl`. They are both a little bit more complicated to do cleanly or without major refactoring, so I felt it would fall out of the scope of this PR. I think this is something we should consider in order to keep things more maintainable. 🥳

tiptop96:
Oh and also the `ctl call` seems to have problems with it's encoding in the REPL. I failed to run it from the CLI so I don't know if it is REPL specific.

src/keys.rs
```diff
@@ -131,17 +141,10 @@ pub(crate) fn list(directory: Option<String>, output: &Output) {
...
-            for key in keys {
-                println!("{}", key);
-            }
-        }
-        OutputKind::JSON => {
-            println!("{}", json!({ "keys": keys }))
-        }
-    }
+    Ok(match output.kind {
+        OutputKind::Text => format!("====== Keys found in {} ======\n{}", dir, keys.join("\n")),
```
brooksmtownsend, line 145: Nice refactor here, much cleaner

brooksmtownsend (commented):
LGTM, before merge can you please update the version in Cargo.toml to 0.1.14?

brooksmtownsend:
@tiptop96 To answer some of your questions, `par insert` worked just fine for me in the REPL, I was able to create and insert different binaries into a `par` file 😄 . Also, `ctl call` seems to work differently in the REPL than in the command line, I believe it has to do with how we end up parsing the payload. In the REPL, we take a bunch of tokens and join them to create the JSON object while in the command line we are expecting a payload object, e.g. a json inside of quotes. I think at the minimum we will want to include that in the help text of the `call` command, and we should file an issue to make that more intuitive & the same with the REPL and with the cli.

PR #53, opened 2021-01-07
Fix amd64 build

billyoung:
- Update nfpm config with latest syntax changes
- Update GH action package steps


brooksmtownsend (approved):
🤞

Issue #62, opened 2021-01-19
Command output logic should be handled at the CLI / REPL level


brooksmtownsend:
Currently, across the `wash` modules, there are a few different ways of outputting information. For the CLI commands, this is generally accomplished with `println` and for the REPL this is accomplished either with the log level macros (`info`, `debug`, etc) or with the `log_to_output` function. With the latest PR where all commands are available in the REPL, these differences become more clear and in the REPL environment frustrating as `println`s do not display properly in the REPL.

I think that `wash` can be refactored to where the top-level call in each module (e.g. `handle_command`) should return a `String` object that contains any appropriate output for that command. Refactoring in this way will allow the `main.rs` module to `println!` the output, and allow the REPL to call `log_to_output` for the appropriate commands. Throughout the code we can maintain appropriate log level filter macros, as those will be logged to the overall REPL log and should be available in the CLI with the appropriate `RUST_LOG` filter.

A few considerations that come to mind:
- The REPL delegation should get simpler, as each command should just be directly passed to the appropriate `handle_command`
- The spinner logic in `reg` and `ctl` will need to be examined, as we do not want to display a spinner in the REPL. This relates to #57 as well, and can probably be taken care of at the same time
- Since the TUI_LOGGER widget accepts log targets, log targets should be defined as constants to keep the level logs organized
- Because this will involve touching a lot of code, this should be tackled quickly to avoid merge conflicts getting out of hand. If I (Brooks) take this on, I'll comment on this issue to denote that I started.

PR #63, opened 2021-01-20
Fixing small issues in 0.2.0 milestone

### LINKED ISSUE #49: Support initiating actor live update
With this PR merged https://github.com/wasmCloud/wasmCloud/pull/32, the wash control interface module should include an `update` command which initiates a live update of an actor.

### LINKED ISSUE #58: determine_directory panics on Windows
```
fn determine_directory(directory: Option<String>) -> String {
    match directory {
        Some(d) => d,
        None => format!("{}/.wash/keys", env::var("HOME").unwrap()),
   }
}
```

This function, https://github.com/wasmCloud/wash/blob/main/src/keys.rs#L147 at the time of writing this issue, causes a panic in a Windows Powershell environment. Instead the following snippet works in the case of windows:
```
format!("{}{}/.wash/keys", env::var("HOMEDRIVE").unwrap(), env::var("HOMEPATH").unwrap())
```

This should just be an additional statement in the match, if `$HOME` is available use that, otherwise use the HOMEDRIVE + HOMEPATH combination. It may be useful to, instead of just unwrapping, provide an error message stating that we attempted to find a home directory and request that the user set `$WASH_HOME` manually.

### LINKED ISSUE #61: Par create success message always has .par extension, even if compressed
See the following code: https://github.com/wasmCloud/wash/blob/main/src/par.rs#L243

This should have the `.par.gz` extension if the `--compress` flag is set

brooksmtownsend:
Fixes #49
Fixes #58
Fixes #61

This PR fixes 3 different issues, one typo, one adding more verbosity in the case that a user does not define `$HOME` or `$WASH_KEYS` for autogeneration, and adding live actor update command to the command interface. I've issued all three of these together so the refactor of display logic (detailed in #62) can be done in one-go.


billyoung (approved):
looks good!

PR #65, opened 2021-01-21
Display refactor

### LINKED ISSUE #62: Command output logic should be handled at the CLI / REPL level

brooksmtownsend:
Fixes #62

This PR refactors the display logic primarily up to each module's `handle_command` function, which then returns a `String` that can be either printed, logged to a REPL output window, or otherwise handled appropriately. This removes a lot of duplicated code, and no `println!` statements exist outside of `main` and the helper function `print_or_log`. Since `println!` statements can corrupt the display of the REPL, this will ensure that REPL usage will not be affected by print statements in `wash`'s codebase.

A few functions have been added to `util` to accomplish this, namely `output_destination` which returns a variant of an enum based on if we are in REPL mode or CLI mode, `print_or_log` if modules must display information to the user during the execution of the function, which will be `println!`'ed in CLI mode and logged at an `info` level in REPL mode. Lastly, the way that we can actually _tell_ if we are in the REPL is by setting the `WASH_REPL_MODE` environment variable, and retrieving that variable when determining output destination.

Setting an environment variable as a "global" constant was not my first attempt at solving this problem, however I had difficulty imagining a cleaner way to infer the display mode AND avoid passing around a variable to every function. A single `static` variable would've sufficed if I could mutate it during the REPL instantiation, but that introduces unsafe code into an otherwise completely Rust safe codebase. Open to any suggestions here.

Some more minor features that made it into this PR:
- Allowing log level filtering with `wash up`, e.g. `wash up --level trace`. Default mode is now `debug` to avoid incredibly verbose logging
- Fixed an issue where `claims token` would not autogenerate keys for `actor` and `provider` types.
- Addressed clippy warnings
- Updated `wasmcloud-host` dependencies for quicker start actor/provider acks and local insecure registries


brooksmtownsend:
Thanks to a great suggestion by @tiptop96 , I was able to avoid using environment variables here by utilizing the `once_cell` for a safely initialized static variable. Feeling much better about the safety of the REPL_MODE check now

PR #68, opened 2021-01-28
Change push distribution code to 35

billyoung:
Removes specific distribution id's from the push to package cloud.

Distribution ID `35` will allow any debian based distribution:
```
    {
      "display_name": "Any deb-based distribution.",
      "index_name": "any",
      "versions": [
        {
          "id": 35,
          "display_name": "Any version.",
          "index_name": "any",
          "version_number": null
        }
      ]
    }
```


brooksmtownsend (approved):
LGTM

PR #69, opened 2021-01-28
Dependency bump, kvcounter script

brooksmtownsend:
Will merge after #68 to ensure new build kicks off


autodidaddict:
I'm okay with this once the clippy warnings are cleared.

PR #72, opened 2021-01-30
Non blocking commands

tiptop96:
This works as far as I can tell.

However I am not too happy with the result. I think the code got a litle verbose when I had repeat `std::thread::spawn + rt.block_on` 5x. Another downside is the added complexity of wrapping the `OutputState` in `Mutex`. 

I would love some input on how to improve this, I have been banging my head against the wall and couldn't get it cleaner.

Despite what's mentioned above I figured I create the PR and we can discuss if we want to merge, hold off on this or maybe if there is a better way. 😄

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.1.16"
+version = "0.1.17"
 authors = ["Brooks Townsend <brooksmtownsend@gmail.com>"]
 edition = "2018"
 repository = "https://github.com/wasmcloud/wash"
--- a/src/up.rs
+++ b/src/up.rs
@@ -335,7 +335,6 @@ impl WashRepl {
                             }
                             ReplCliCommand::Claims(claimscmd) => {
                                 let output_state = Arc::clone(&self.output_state);
-                                //let output_state = Arc::clone(Arc::clone(&self.output_state));
                                 std::thread::spawn(|| {
                                     let mut rt = actix_rt::System::new("cmd");
                                     rt.block_on(async {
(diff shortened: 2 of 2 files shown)


brooksmtownsend:
Getting to review this now @tiptop96, I think the code looks clean for wrapping the OutputState in an `Arc<Mutex<>>`. I'm looking to test it out with some edge cases before I add my formal review, BTW you may want to pull in the latest from `main` as dependencies for `tui` and `nkeys` has updated.

tiptop96:
Awesome! Do you prefer a rebase or merge on/with main? 😊

brooksmtownsend:
@tiptop96 Preferably rebase as long as it doesn't cause too many conflicts on your end (shouldn't have any I think.) Either one is fine, we squash all PRs going into main so it'll just be added as an extra commit.

brooksmtownsend:
My biggest concern with the Mutex was potential poisoning when a thread panics, e.g. I was wondering what would happen to the OutputState in the case that there was a panic. The only way I know how to induce a panic at the moment is by trying to bind the `httpserver` provider to the same port twice with two actors, and that (though it gives an ugly `Unwrap` error for the address being in use) didn't poison the mutex. (doc for later reference https://doc.rust-lang.org/std/sync/struct.Mutex.html#examples). 

If this becomes an issue in the future we can address this, but for now I think we're good to go. Will review code momentarily

brooksmtownsend (changes requested):
Code looks great, and works great with my testing in the REPL. Can you please bump the patch version to `0.1.17`? Once that's done we'll get this merged 🚀

tiptop96:
Awesome 👏 will `bump n push` in the morning 😴

brooksmtownsend (approved):

PR #75, opened 2021-02-03
refactor how we manage pushing deb/rpm to many distributions

billyoung:
This refactors how we manage pushing to _many_ distributions.

Packagecloud uses "distribution version id's" which means in order to support different versions of ubuntu, debian, fedora, etc, we need to push a package with the specific id. 

Using the matrix for deb and rpm distribution id's, we can simplify this action considerably. If or when packagecloud adds support for a new distribution, we can simply add the ID to the matrix.


brooksmtownsend (approved):
LGTM, should we cut a new release once this gets merged in to re-distribute?

billyoung:
thanks @brooksmtownsend - we can definitely cut a new release after this is merged. lets make sure we use a new tag though, as packagecloud doesn't allow for overriding packages with existing name/version.

PR #78, opened 2021-02-05
Drain

tiptop96:
If anyone could give suggestions for the doc comments that would be greatly appreciated. 🥇

### DIFF
--- a/src/drain.rs
+++ b/src/drain.rs
@@ -62,7 +62,7 @@ pub(crate) fn handle_command(cmd: DrainCliCommand) -> Result<String, Box<dyn ::s
         cleared.push(remove_dir_contents(path)?);
     }
     Ok(match cmd.output.kind {
-        OutputKind::Text => format!("Successfully cleard cahces at: {:?}", cleared),
+        OutputKind::Text => format!("Successfully cleared caches at: {:?}", cleared),
         OutputKind::JSON => json!({ "drained": cleared }).to_string(),
     })
 }
--- a/src/drain.rs
+++ b/src/drain.rs
@@ -1,5 +1,4 @@
-use crate::util::{Output, OutputKind};
-use fs::remove_file;
+use crate::util::{GlobalOutput, Output, OutputKind};
 use serde_json::json;
 use std::env;
 use std::path::Path;
@@ -21,9 +20,9 @@ impl DrainCli {
 #[derive(StructOpt, Debug, Clone)]
 pub(crate) struct DrainCliCommand {
     #[structopt(flatten)]
-    output: Output,
-    #[structopt(flatten)]
     selection: DrainSelection,
+    #[structopt(flatten)]
+    output: GlobalOutput,
 }
 
 impl IntoIterator for DrainSelection {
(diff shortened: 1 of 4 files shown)


src/drain.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+#[derive(StructOpt, Debug, Clone)]
+pub(crate) struct DrainCliCommand {
+    #[structopt(subcommand)]
+    selection: DrainSelection,
+}
+
+// Hack to get around the fact that we can't have a global 'output' arg.
```
brooksmtownsend: We should try and leave out `hack` from our code, maybe something like
```
Propagates output selection from CLI to all commands
```

src/main.rs
```diff
@@ -38,6 +40,9 @@ struct Cli {
 
 #[derive(Debug, Clone, StructOpt)]
 enum CliCommand {
+    /// TODO
```
brooksmtownsend: "Manage contents of local wasmCloud cache"

src/up.rs
```diff
@@ -104,6 +105,10 @@ struct ReplCli {
 #[derive(StructOpt, Debug, Clone)]
 #[structopt(global_settings(&[AppSettings::ColorNever, AppSettings::DisableVersion, AppSettings::VersionlessSubcommands]))]
 enum ReplCliCommand {
+    // TODO
```
brooksmtownsend: "Manage contents of local wasmCloud cache"

src/drain.rs
```diff
@@ -0,0 +1,92 @@
...
+            DrainSelection::All(output)
+            | DrainSelection::Lib(output)
+            | DrainSelection::Oci(output) => output.kind,
+        }
+    }
+}
+
+#[derive(StructOpt, Debug, Clone)]
+pub(crate) enum DrainSelection {
+    /// TODO
```
brooksmtownsend: Remove all cached files created by wasmCloud

src/drain.rs
```diff
@@ -0,0 +1,92 @@
...
+    All(Output),
+    /// TODO
```
brooksmtownsend: Remove cached files downloaded from OCI registries by wasmCloud

src/drain.rs
```diff
@@ -0,0 +1,92 @@
...
+    Oci(Output),
+    /// TODO
```
brooksmtownsend: Remove cached binaries extracted from provider archives

brooksmtownsend (changes requested):
One suggestion and I threw in what I thought appropriate doc comments would be!

src/drain.rs
tiptop96: Yeah, that's a good point 👍

src/up.rs
```diff
@@ -553,6 +567,12 @@ async fn handle_up(cmd: UpCliCommand) -> Result<()> {
     Ok(())
 }
 
+fn handle_drain(drain_cmd: DrainCliCommand, output_state: Arc<Mutex<OutputState>>) -> Result<()> {
+    let output = crate::drain::handle_command(drain_cmd)?;
+    log_to_output(output_state, output);
+    Ok(())
+}
```
brooksmtownsend, line 574: Love how clean the functions are getting now 😄

brooksmtownsend (approved):
Looks great, and works great 😄

PR #82, opened 2021-02-08
Fix build matrix. Adds bionic to deb release versions

billyoung:
@brooksmtownsend this should fix the issue with the build matrix. Decided to go with simple bash `for` loops. It also adds ubuntu bionic `190` to the list of distribution id's.


brooksmtownsend (approved):
LGTM

PR #95, opened 2021-02-19
Allow tui logger to scroll

brooksmtownsend:
With the upgrade of tui_logger to `0.6.0`, we had to modify our terminal backend from crossterm to termion, which accounts for most of the code in this PR.

The important new features are:
- Added ALT+UP/DOWN / PageUp + PageDown handlers to the output pane and tui logger, which will scroll the appropriate pane depending on the current focus
- Added indicator of REPL focus, if focused on the input panel (e.g. typing commands) the REPL and Output window titles will be bold, if focused on the tui panel (e.g. scrolling through logs) the Tui panel will be bold.
- Using channels to send events to asynchronously draw REPL (modification needed as crossterm provided a way to asynchronously poll for events, rather than blocking iteration of events)

Because of the change from crossterm to termion, we no longer have support for SHIFT modified keycodes. Because of this, and changing to ALT, this will change the wash REPL API and we will need to bump to `0.3.0`

Before this is merged in, the following must be tested:
- [x] ALT / PgUp / PgDown commands work on Windows and Linux
- [x] Return (Enter) works on Windows and Linux

### DIFF
--- a/src/up.rs
+++ b/src/up.rs
@@ -5,15 +5,18 @@ use crate::keys::*;
 use crate::par::*;
 use crate::reg::*;
 use crate::util::{convert_error, Result, WASH_CMD_INFO, WASH_LOG_INFO};
-use crossterm::event::{poll, read, DisableMouseCapture, Event, KeyCode, KeyEvent, KeyModifiers};
-use crossterm::terminal::{self, EnterAlternateScreen, LeaveAlternateScreen};
 use log::{error, info, LevelFilter};
-use std::io::{self, Stdout};
+use std::io;
 use std::sync::{Arc, Mutex};
-use std::{cell::RefCell, rc::Rc};
 use structopt::{clap::AppSettings, StructOpt};
+use termion::event::{Event, Key};
+use termion::{
+    input::{MouseTerminal, TermRead},
+    raw::IntoRawMode,
+    screen::AlternateScreen,
+};
 use tui::{
-    backend::CrosstermBackend,
+    backend::TermionBackend,
     layout::{Alignment, Constraint, Direction, Layout, Rect},
     style::{Color, Modifier, Style},
     text::Span,
@@ -23,8 +26,18 @@ use tui::{
 use tui_logger::*;
 use wasmcloud_host::HostBuilder;
 
+type REPLTermionBackend = tui::backend::TermionBackend<
+    termion::screen::AlternateScreen<
+        termion::input::MouseTerminal<termion::raw::RawTerminal<std::io::Stdout>>,
+    >,
+>;
+
 const CTL_NS: &str = "default";
 const WASH_PROMPT: &str = "wash> ";
+/// Option is unsupported for MacOS, the following byte slices correspond
+/// to [1;3A for Option+UP and [1;3B for Option+Down
+const OPTIONUP: &[u8] = &[27_u8, 91_u8, 49_u8, 59_u8, 51_u8, 65_u8];
+const OPTIONDOWN: &[u8] = &[27_u8, 91_u8, 49_u8, 59_u8, 51_u8, 66_u8];
 
 #[derive(Debug, StructOpt, Clone)]
 #[structopt(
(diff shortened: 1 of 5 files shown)


stevelr:
Thank you for working on the log scrolling. I tried wash (0.3.0 from the tui-log-update branch and have some feedback:

- scrolling the Output window with Alt+UP/DOWN and Tui Log with PgUp/PgDown works.

- if you type 'help' in REPL, it pastes help output into the Tui Log, and then it seems to lose the history of logs before the help output, and there is only one line of scrollback buffer.  If you don't type 'help', and the TuiLog has more than a screenful of regular log messages, the messages do scroll back further.

- In wash 0.2.0, I used to be able to select text in the output window but can't now. When you're doing the very first steps in the getting started tutorial, you do 'ctl get hosts', then 'ctl get inventory ...' and that's where I would double-click on the Host Id and paste it into the REPL window to complete the inventory command.  I can't select text anymore (in any of the four panes). This is kind of a big deal for new users (how do you do it btw?). I tried typing the first few letters, to see if it matches an unambiguous prefix, but that didn't seem to work. I did find [this tui crate issue](https://github.com/fdehau/tui-rs/issues/260) where one user couldn't select and the response was 

> "If you use input::MouseTerminal from termion then your application will intercept all mouse events and thus disabling the "normal" behavior of your terminal of choice in response to mouse click or drag events"

Since you're more familiar with the codebase I won't try to debug this further.

brooksmtownsend:
Thank you @stevelr, glad to hear it's working and thank you for doing some more extensive testing on your end. I can reproduce the `help` text and the selection on my end, I changed the terminal backend from crossterm to termion as a part of this upgrade so I bet that the mouse events are being intercepted when they weren't before. Planning to investigate further today and when I have any update's I'll let you know

brooksmtownsend:
Good news @stevelr , I was able to resolve both issues that you mentioned. You should now be able to scroll past the help text and use your mouse to highlight the identifiers, which is a must-have. Let me know if you do further testing here, otherwise I plan to do a few more tests and get this merged 🙂

brooksmtownsend:
Fixes #95

src/up.rs
```diff
@@ -238,78 +247,87 @@ impl WashRepl {
...
                     && self.input_state.input_cursor <= self.input_state.input.len()
                 {
                     self.input_state.input_cursor -= 1;
                     self.input_state.input.remove(self.input_state.input_cursor);
                 };
             }
-            KeyCode::Enter => {
+            Key::Alt(c) if c == 'b' => {
+                //TODO(issue #67): navigate left one word
+                ()
```
autodidaddict: This triggers a clippy warning on an unneeded unit expression.

src/up.rs
```diff
@@ -238,78 +247,87 @@ impl WashRepl {
...
+            }
+            Key::Alt(c) if c == 'f' => {
+                //TODO(issue #67): navigate right one word
+                ()
```
autodidaddict: Same as previous comment.

src/up.rs
```diff
@@ -524,20 +607,32 @@ https://www.wasmcloud.dev/overview/getting-started/#starting-nats for instructio
...
+    std::thread::spawn({
+        let f = {
+            let stdin = io::stdin();
+            move || {
+                for c in stdin.events() {
+                    tx_event.send(c).unwrap();
+                }
+            }
+        };
+        f
```
autodidaddict: clippy: you don't need to assign the block to `f`, you can just return the block value.

src/up.rs
```diff
@@ -705,12 +807,19 @@ fn draw_input_panel(
...
     // Draw REPL panel
     let input_panel = Paragraph::new(display)
-        .block(Block::default().borders(Borders::ALL).title(Span::styled(
-            " REPL ",
-            Style::default().add_modifier(Modifier::BOLD),
-        )))
+        .block(
+            Block::default()
+                .borders(Borders::ALL)
+                .title(Span::styled(format!(" REPL "), style)),
```
autodidaddict: wow.. clippy does not like you. Unnecessary format.

autodidaddict (changes requested):
just make clippy happy.

autodidaddict (approved):

PR #96, opened 2021-02-19
Integration tests

### LINKED ISSUE #88: Integration tests
With our latest milestone release of `0.2.0`, I began thinking about an API promise to ensure that we aren't pushing breaking changes as patch versions. Because we are pre-release, the versioning of wash from here on out will follow a pre-release semver, e.g. as long as our current API is compatible, new features can be added as patch versions. Any breaking API change will be added as minor versions.

Because of this, a full integration test suite would benefit the project. Structopt allows constructing commands from str slices in Rust, so we can fully mock a user input. If we mocked the full commands with multiple (all) command line flags, the tests would fail as soon as our API changes (e.g. flag is removed, renamed). This would be a large-ish undertaking to fully test.

I want to avoid testing the specific _libraries_ that wash depends on (e.g. we don't need to test that `wash key gen module` generates a valid nkey, that's the `nkeys` crate's job). Instead, we should focus on each command that you would type in the CLI / REPL is executed as expected.  Because the REPL re-uses commands from the CLI, I don't believe we will need to write tests for the REPL commands themselves.

brooksmtownsend:
Fixes #88 

The vast majority of code in this PR is the "comprehensive" or "enumerate" tests, which include every flag and option that a wash command can be composed of. These are included as unit tests, however, they do not test any functionality related to the commands _actually running correctly_. The benefit of these tests is to determine API drift, and if any fail then we can examine if we have changed the API of `wash` with a subsequent PR. Once these are in place we can have some confidence knowing that if we change CLI flags, options, or order of required arguments, that tests will fail to indicate that has occurred.

The integration tests in this PR aim to test real-world usage of the `wash` commands to ensure:
1. Our API does not change without us appropriately adjusting the version. API is defined here as the command line interface that users interact with.
2. New features don't break older functionality

I've tested this PR on both of my machines by running `make test-all` where I have a valid docker and docker-compose installation, and it's worked well. If you run into any issues running these tests (I assume because the GH runner works well, that it will work on your machine) then we can address those.


src/drain.rs
```diff
@@ -90,3 +90,28 @@ fn remove_dir_contents<P: AsRef<Path>>(path: P) -> Result<String, Box<dyn ::std:
...
+            DrainSelection::Lib(output) => assert_eq!(output.kind, OutputKind::Text),
+            _ => panic!("drain constructed incorrect command"),
+        }
+        let oci = DrainCli::from_iter_safe(&["drain", "oci", "-o", "json"]).unwrap();
+        match oci.command.selection {
+            DrainSelection::Oci(output) => assert_eq!(output.kind, OutputKind::JSON),
+            _ => panic!("drain constructed incorrect command"),
+        }
+    }
+}
```
autodidaddict, line 117: Does this catch the case where you didn't specify one of the options? for example, if an option is added to `drain`..  I'm assuming that's okay since it won't break any pre-existing scripts that used the drain command... to your point in the PR that this is designed to stop API drift, that drift is specifically breaking changes of existing commands and not the harmless addition of new commands?

src/par.rs
```diff
@@ -428,3 +428,312 @@ fn is_compressed(input: &[u8]) -> Result<bool> {
...
+                assert_eq!(issuer.unwrap(), ISSUER);
+                assert_eq!(subject.unwrap(), SUBJECT);
+                assert_eq!(output.kind, OutputKind::Text);
+                assert_eq!(name, "CreateTest");
+                assert_eq!(vendor, "TestRunner");
+                assert_eq!(destination.unwrap(), "./test.par.gz");
+                assert_eq!(revision.unwrap(), 1);
+                assert_eq!(version.unwrap(), "1.11.111");
+                assert!(disable_keygen);
+                assert!(compress);
```
autodidaddict, line 503: I really like this. It will definitely keep us from making changes to a command or sub-command that we think are harmless but actually cause breaks or worse, no breaks but the command does something the user didn't expect.

autodidaddict (approved):

src/drain.rs
brooksmtownsend: Yeah correct, so if you don't specify one of the options, e.g. `wash drain` then that just prints structopt help text so I didn't design a test around that. Same thing with `wash drain foobar`, we could design a test that provides an unexpected argument, though I think that would be testing structopt more than our code

brooksmtownsend: To answer your original question, none of the tests should fail with another option being introduced, these clauses are more for if we supply `wash drain lib` but it creates a `wash drain foo` command, then we definitely changed the API

PR #102, opened 2021-03-02
Call Alias

### LINKED ISSUE #101: [FEATURE] Support call alias in signing
No problem

Add the ability to specify an actor's call alias while signing the actor

None

**Additional context**
The call alias field is now available in the `Claims<Actor>` struct from `wascap`, so adding this feature would be a matter of just supporting the setting of this field at signing time.

brooksmtownsend:
Fixes #101 

This PR allows developers to use `wash` to sign actors with a call alias, a more human-readable human-typable alternative to the public key.

Currently, a call alias can be comprised of:
- Alphanumeric characters
- Underscores
- Backslashes

A call alias cannot be:
- Empty
- An uppercase 56 character public key (to enforce best practices)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.2.1"
+version = "0.3.1"
 authors = ["wasmcloud Team"]
 edition = "2018"
 repository = "https://github.com/wasmcloud/wash"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.3.1"
+version = "0.2.2"
 authors = ["wasmcloud Team"]
 edition = "2018"
 repository = "https://github.com/wasmcloud/wash"
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -304,6 +304,9 @@ pub(crate) struct ActorMetadata {
     /// Human-readable version string
     #[structopt(short = "v", long = "ver")]
     ver: Option<String>,
+    /// Developer or human friendly unique alias for an alternative identifier for an actor
+    #[structopt(short = "a", long = "call-alias")]
+    call_alias: Option<String>,
 
     /// Path to issuer seed key (account). If this flag is not provided, the will be sourced from $WASH_KEYS ($HOME/.wash/keys) or generated for you if it cannot be found.
     #[structopt(
@@ -423,7 +426,7 @@ fn generate_actor(actor: ActorMetadata) -> Result<String, Box<dyn ::std::error::
         actor.provider,
         actor.rev,
         actor.ver.clone(),
-        None,
+        sanitize_alias(actor.call_alias)?,
     );
 
     let jwt = claims.encode(&issuer)?;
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.2.1"
+version = "0.3.1"
```
autodidaddict: Not to be a pedant here, but semver suggests that if we break an API, we bump from 0.2.1 to 0.3.0

src/claims.rs
```diff
@@ -304,6 +304,9 @@ pub(crate) struct ActorMetadata {
     /// Human-readable version string
     #[structopt(short = "v", long = "ver")]
     ver: Option<String>,
+    /// Developer or human friendly unique alias for an alternative identifier for an actor
```
autodidaddict: Can we be a bit more explicit that this alias is used specifically for invoking actors?

src/claims.rs
```diff
@@ -739,6 +742,14 @@ pub(crate) fn render_actor_claims(
         vec![]
     };
 
+    let call_alias = claims
+        .metadata
+        .as_ref()
+        .unwrap()
+        .call_alias
+        .clone()
+        .unwrap_or("N/A".to_string());
```
autodidaddict: per slack, change `N/A` to something that we know can't pass through normalization, e.g. `(Not set)`

autodidaddict (changes requested):

Cargo.toml
brooksmtownsend: Not pedantic at all, this was done on an assumption that #95 would be merged in before this PR as it's been out there for a bit longer, and that _does_ break an API. This PR does not break an API in my opinion, only adds an optional flag, so it would be a patch version past that PR.

If we merge this PR in first, I think it should be `0.2.2`, unless you think this change breaks the API

src/claims.rs
brooksmtownsend: Changed to `(Not set)`

src/claims.rs
brooksmtownsend: Let me know what you think of the new helptext, I also included our validation rules to remove the guess and check factor if you don't know the expected format

autodidaddict (dismissed):

src/claims.rs
```diff
@@ -848,11 +866,89 @@ where
...
+        {
+            Err("Public key cannot be used as a call alias".into())
+        // Valid aliases contain a combination of lowercase alphanumeric characters, dashes, and backslashes
+        } else if alias
+            .chars()
+            .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '_' || c == '/')
+        {
+            Ok(Some(alias))
+        } else {
+            Err("Call alias contained invalid characters.\nValid aliases are lowercase alphanumeric and can contain underscores and backslashes".into())
```
autodidaddict: s/backslashes/slashes ;) :trollface:

autodidaddict (changes requested):

src/claims.rs
brooksmtownsend: image
darnit

autodidaddict (approved):

Issue #92, opened 2021-02-18
Fully contained REPL host (e.g. no NATS)


brooksmtownsend:
Currently, the wash REPL launches with its own instantiated host. This host can be interacted with via the wasmcloud [control-interface](https://github.com/wasmCloud/wasmCloud/tree/main/crates/wasmcloud-control-interface), and as such requires NATS to be running so that `ctl` commands can be issued to that host.

Requiring NATS to be present for the REPL host, while it is representative of a more "production-ready" architecture for wasmcloud, adds an additional layer of friction to the initial developer experience. I propose that we add the ability, in the event of failing to connect to NATS, to launch a REPL host without the rpc and control clients. This host should have an event collector (mentioned in #91) once that's implemented in `wasmcloud-host`.

With a host that does not have a NATS connection, all `ctl` commands can be instead directly invoked on that host struct. This means that, with our current architecture, the REPL host will need to be able to receive commands instead of just executing in its own thread. An implementation of this might look like using [channels](https://doc.rust-lang.org/rust-by-example/std_misc/channels.html) to pass commands / events to the thread containing the host. Adding a boolean flag denoting the NATS connection to some form of REPL state may be beneficial to easily determine if a command should be published to a control interface or invoked directly on the local host.

brooksmtownsend:
Depends on https://github.com/wasmCloud/wasmCloud/issues/112

brooksmtownsend:
Fixed by #107

PR #107, opened 2021-03-05
No lattice repl

brooksmtownsend:
Addresses #92 

No-nats mode now supports all but the following commands from the standard lattice connected mode:
1. `ctl get claims`
2. `ctl update actor`
3. `ctl get inventory` (partial)
These are all unsupported at the moment because there are not equivalent commands to invoke on the host directly, and some make use of the host controller on the control interface side in order to return the correct information. This will need to be addressed. I'm primarily concerned about implementing `get inventory` for the host inventory, as that contains much needed information, specifically link names and OCI references. @autodidaddict I've added what I can of support for now (see the host inventory implementation for standalone) however we might be better off implementing some of this functionality on the host side.

For this PR, I've added a few additional threads and channels to allow `wash up` to run a REPL without connecting to NATS, as indicated by the `Standalone` title by the REPL. This launches an embedded host, as usual, however, commands will be invoked directly on that host instead of proxying through the control interface.

A few modifications for reviewers:
- `ctl` and `up` are getting to contain large amounts of code, so I've split some large pieces into separate files under a folder of the same name. This is a pretty standard Rust pattern. I didn't split larger pieces of the REPL out into separate files because I made many assumptions about accessing fields, and I'd have to write just as much code as I'd save in getters, setters, and modifications.
- the `output` module under `ctl/output.rs` is used to standardize the output format for control interface operations and embedded host operations, that way what the user sees should be identical for standalone mode and lattice connected mode.
- I edited some integration tests to better check for failures while I was refactoring the output modules.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ nkeys = "0.1.0"
 wascap = "0.6.0"
 provider-archive = "0.4.0"
 wasmcloud-control-interface = "0.1.0"
-wasmcloud-host = "0.15.1"
+wasmcloud-host = "0.15.3"
 
 [dev-dependencies]
 test_bin = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,18 +22,18 @@ tui-logger = "0.6.0"
 tui = { version = "0.14.0", default-features = true }
 log = "0.4.14"
 termion = "1.5"
-actix-rt = "1.1.1"
+actix-rt = "2.2.0"
 spinners = "1.2.0"
 nats = "0.8.6"
 once_cell = "1.5.2"
 term-table = "1.3.1"
-oci-distribution = "0.5.0"
+oci-distribution = "0.6.0"
 
 nkeys = "0.1.0"
 wascap = "0.6.0"
 provider-archive = "0.4.0"
-wasmcloud-control-interface = "0.1.0"
-wasmcloud-host = "0.15.3"
+wasmcloud-control-interface = "0.2.1"
+wasmcloud-host = "0.16.1"
 
 [dev-dependencies]
 test_bin = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.3.2"
+version = "0.4.0"
 authors = ["wasmCloud Team"]
 edition = "2018"
 repository = "https://github.com/wasmcloud/wash"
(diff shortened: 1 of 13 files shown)


autodidaddict:
Safe? You think the weekend makes you safe?
image

autodidaddict:
Maybe the diff isn't showing it, or I'm confused. I was looking for the code that uses channels to send commands either to the control interface or to the managed host.. is that somewhere else?

brooksmtownsend:
@autodidaddict Yeah it's just moved around because I restructured the project a bit to clean up the code, check `src/up/mod.rs` https://github.com/wasmCloud/wash/pull/107/files#diff-dcf780801376c219d3b5d7f2f9a5a2745c5ccb124b8c621281e13d3c775ed9f3

autodidaddict:
Thanks for the link.. though it's still not in the diff, this is what I was looking for:

https://github.com/wasmCloud/wash/blob/dd0540b3e56891689a10ac8274591948794c8920/src/up/mod.rs#L642

autodidaddict (dismissed):

brooksmtownsend:
As of my latest commit:
- Fixed bug where characters would show vertically across output window
- Implemented the above three functions, which now brings standalone mode to full parity with lattice connected mode

There are only a few more areas where I need to implement functionality `wasmcloud-host`:
- [x]  Need a way to get the provider link name for host inventory (possible after https://github.com/wasmCloud/wasmCloud/pull/159)
- [x]  Need a way to update the OCI mapping when an actor is updated (currently if I update `echo:0.2.0` to `echo:0.2.1`, the OCI mapping still contains the echo public key and `echo:0.2.0`)
- [x]  Need to have connection parameters provided for update actor, currently they are just defaults (no latest tags, secure connection, no user/pass) (depends on https://github.com/wasmCloud/wasmCloud/issues/158, not a blocker)

brooksmtownsend:
All functionality has been implemented and `wash` no-nats (standalone) mode has exact parity with nats (lattice) mode. The build is going to fail because of relative path dependencies and that can be remedied once https://github.com/wasmCloud/wasmCloud/pull/163 is ready

brooksmtownsend:
REVIEWER'S GUIDE TO SOME OF THE BIG CHANGES (@autodidaddict )

I've refactored the `ctl` and `up` modules so they appear to have been deleted and written. The `ctl` module has been refactored into `mod` and `output`, where `mod` is the functionality of the `ctl` module and `output` is where all of the helper functions exist to translate the CTL resources into a String for output in the terminal. The neat thing about this is that I'm using these helper functions with the standalone mode as well, that way there will be no difference between what the user sees in either mode.

`up` was refactored as well, mostly to pull some of the large parts of standalone mode types into another file. I'd like to continue to refactor this a bit as `up` and `ctl` are approaching 1000 lines but for now this is enough change.

autodidaddict (changes requested):
LGTM once clippy has been placated.

autodidaddict (approved):

PR #111, opened 2021-03-23
wash: don't panic on invalid URL in claims inspect

l-mb:
Minor tweak; misunderstanding what syntax to pass in to the "claims inspect" function netted me a panic(), I think that could be somewhat nicer ;-)


autodidaddict (approved):

brooksmtownsend (approved):
Small change, big impact! Thank you

PR #115, opened 2021-03-30
properly generate tokens for providers

brooksmtownsend:
Previously, `wash claims token provider` would generate a module key instead of a service key for providers. This is not correct, and any JWTs created this way could be mistaken for actors.

### DIFF
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -530,7 +530,7 @@ fn generate_provider(provider: ProviderMetadata) -> Result<String, Box<dyn ::std
         provider.subject.clone(),
         Some(provider.name.clone()),
         provider.common.directory.clone(),
-        KeyPairType::Module,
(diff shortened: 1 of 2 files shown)


src/claims.rs
```diff
@@ -530,7 +530,7 @@ fn generate_provider(provider: ProviderMetadata) -> Result<String, Box<dyn ::std
         provider.subject.clone(),
         Some(provider.name.clone()),
         provider.common.directory.clone(),
-        KeyPairType::Module,
+        KeyPairType::Service,
```
autodidaddict, line 533: :facepalm:

autodidaddict (approved):

src/claims.rs
brooksmtownsend: Who even wrote this `git blame`

Issue #106, opened 2021-03-03
[FEATURE] Make it easier to distinguish REPL focus


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
When interacting with the `wash` REPL, there are two focus modes that you can alternate between by pressing TAB. The first is the REPL focus where you are typing commands in the input window, and the second is the Tui logger focus where you can adjust the level and kind of logs that you are seeing. This (as of #95) is currently indicated as an `info!` level log message and by bolding the text of the title of each pane. Still, it is difficult to tell the difference as bolding the title is not a big difference.

**Describe the solution you'd like**
We should have a more prominent visual indicator of where the REPL focus is, whether it's a background change, border thickness adjustment, etc.

**Additional context**
This can be done by modifying the `Style` attribute of the input / output / tui_log panels which is passed in via the `.style()` method right before they are rendered. You can see how I'm currently passing the "focused" indicator to the `draw_input_panel`, `draw_output_panel` and `draw_smart_logger` functions already, so it should just be a matter of adjusting the Style object appropriately.

PR #119, opened 2021-04-17
REPL focus - a first pass

VestigeJ:
This PR addresses [#106](https://github.com/wasmCloud/wash/issues/106) to make it easier for the user to distinguish focus while using the REPL. I  think my next pass will try to incorporate the KeyEvents to listen for someone holding the option/fn key prior to paging up/down through the log data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.3.2"
+version = "0.4.0"
 authors = ["wasmCloud Team"]
 edition = "2018"
 repository = "https://github.com/wasmcloud/wash"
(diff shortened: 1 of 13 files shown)


autodidaddict (approved):

brooksmtownsend (changes requested):
Looks like tests are angry for some reason but the distinguished colors look good. One thing I'd like to see is the `Slow Blink` work, as currently it doesn't seem to change my input panel and we would want people in B/W terminals to have an indication as well if possible

VestigeJ:
I tried using terminalizer to record the local version running. Are you not seeing SLOW_BLINK on your end? That one worked fine on mine but noticed Modifier::DIM did nothing visible on the debug build. 

I'll aim to get the fixes done tonight.

PR #120, opened 2021-04-20
A second breakfast first pass to add repl focus cues to the wash repl

### LINKED ISSUE #106: [FEATURE] Make it easier to distinguish REPL focus

VestigeJ:
Intended to close [#106](https://github.com/wasmCloud/wash/issues/106).

I plan to revisit this functionality, I think it would look nice to bind the output panel and the logging panel to when alt or cmd KeyEvents are detected so those key modifiers are more intuitively tied to the output/logging windows for scrolling. 

I ran the tests that failed prior without issue so I expect this one to be in the clear.

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -1093,7 +1093,7 @@ fn draw_input_panel(frame: &mut Frame<ReplTermionBackend>, state: &mut InputStat
     state.input_width = chunk.width as usize - 3;
 
     let style = if state.focused {
-        Style::default().add_modifier(Modifier::BOLD)
+        Style::default().fg(Color::Cyan).add_modifier(Modifier::BOLD | Modifier::SLOW_BLINK)
     } else {
         Style::default()
     };
@@ -1144,7 +1144,7 @@ fn draw_output_panel(
     state.output_width = chunk.width as usize - 1;
 
     let style = if focused {
-        Style::default().add_modifier(Modifier::BOLD)
+        Style::default().fg(Color::Cyan).add_modifier(Modifier::BOLD)
     } else {
         Style::default()
     };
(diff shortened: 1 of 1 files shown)


autodidaddict:
Are we saying there's a test we expect to fail?

autodidaddict (dismissed):

autodidaddict:
Looks like this needs cargo fmt run against it

brooksmtownsend (dismissed):
For some reason I don't see the Slow blink still but it might just be my terminal, the `cyan` is plenty to distinguish 🙂 

Same comment as @autodidaddict , you just need a `cargo fmt` for this to pass. Thank you for doing this!

VestigeJ:
> For some reason I don't see the Slow blink still but it might just be my terminal, the `cyan` is plenty to distinguish slightly_smiling_face
> 
> Same comment as @autodidaddict , you just need a `cargo fmt` for this to pass. Thank you for doing this!

Thanks for letting me learn among all of you. It's good for me.

autodidaddict (approved):

autodidaddict (approved):

PR #123, opened 2021-04-21
Manifest support

### LINKED ISSUE #116: [FEATURE] Ability to consume manifest file directly
**Is your feature request related to a problem?

Linking actors and capabilities via command-line is definitely practical but a bit tedious and hard. Bootstrapping REPL environment via manifest file can save time and reduce friction.

brooksmtownsend:
Fixes #116 

I've refactored a good chunk of the `up` `mod.rs` out into its own file, none of the changes needed to make this feature work took place during that refactor, it's a direct refactor.

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -898,7 +898,7 @@ mod test {
                 assert_eq!(rpc_host, RPC_HOST);
                 assert_eq!(rpc_port, RPC_PORT);
                 assert_eq!(log_level, LogLevel::Info);
-                assert_eq!(manifest, Some("mani.yaml"));
+                assert_eq!(manifest.unwrap().to_str().unwrap(), "mani.yaml");
             }
             cmd => panic!("up generated other command {:?}", cmd),
         }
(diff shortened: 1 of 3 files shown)


src/up/mod.rs
```diff
@@ -656,6 +243,19 @@ async fn handle_up(cmd: UpCliCommand) -> Result<()> {
                     "Host ({}) started in namespace ({})", host.id(), CTL_NS
                 );
             };
+            // If supplied, initialize the host with a manifest
+            if let Some(pb) = cmd.manifest {
+                match HostManifest::from_path(pb.clone(), true) {
+                    Ok(hm) => {
+                        host_output_sender.send("Initializing host from manifest ...".to_string()).unwrap();
+                        host.apply_manifest(hm).await.unwrap();
```
autodidaddict: Probably shouldn't unwrap here. If we fail to apply the manifest, do we want to crash the repl?

autodidaddict (changes requested):
Just worried about one potentialy crash-inducing unwrap, otherwise LGTM.

src/up/mod.rs
brooksmtownsend: Good call, I faintly remember trying to crash the REPL with an unwrap but I don't know if it was this one. Fix incoming.

autodidaddict (approved):

PR #125, opened 2021-04-21
`up` integration tests

brooksmtownsend:
Previously we were not able to test `up` with integration tests because of the way that the REPL hijacks the terminal's stdout. I've been playing around with the REPL and found a configuration that works where we can send commands as stdin to a REPL, the only thing I need to verify first is that it works with the Github runner.

This is becoming more important as the REPL is getting more and more complicated, especially as we have functionality in `Standalone` and `Lattice` mode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ test-all: test test-integration-wasm3 test-integration-wasmtime ## Run all tests
 
 clean:
 	wash drain all
+	docker-compose -f ./tools/docker-compose.yml down
 
 help:  ## Display this help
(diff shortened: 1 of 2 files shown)


brooksmtownsend:
Thinking on this a bit more, it might be best to define some sort of "quiet" or "testing" mode for the REPL where the terminal UI is not displayed, but stdin is still captured for REPL commands. This would be a very minimal environment built moreso for testing than real usage.

If it's unusable for people, then we can define a hidden structopt option to avoid confusion.

PR #126, opened 2021-04-23
Add a Nix Flake

theduke:
This PR adds a `flake.nix` file, which is a package definition for the [Nix](https://nixos.org/learn.html) package management system.

This allows users of Nix to easily run wash with `nix run github:wasmCloud/wash`.

Note that minimal maintenance is required. The only case where `flake.nix` would have to be updated is when new native library dependencies are added.


brooksmtownsend:
@theduke I'm new to `nix`, actually haven't seen it before! I've downloaded it and wanted to give it a go locally before I approve, and I'm not having much luck:
```
nix run -f flake.nix
```
Seems to put me in a subshell, but it doesn't make `wash` available. Am I running the wrong command here?

Thanks for submitting this PR 🎉

theduke:
So two things:

Flakes is the new official package system for Nix, but still unstable, so you need to enable Flakes support with the steps specified here under "Non-NixOS" : https://nixos.wiki/wiki/Flakes 

With that done you can `cd` into the repo and execute `nix run .#` to build and run wash.

theduke:
Side note: Nix is a bit complicated, but it's brilliant for fully repeatable builds, package distribution and also for deterministic development environments.

Thanks to the `devShell` in the `flake.nix` You can just run `nix develop` and get a development environment with all dependencies (in this case rustc, cargo, clang, openssl) set up and available, without installing anything globally.

And, as mentioned, if the flake is in the repo any nix user can just run `nix run github:wasmCloud/wash`, which downloads, builds and executes wash directly without installing anything or worrying about setting up dependencies.

I recommend giving it a closer look. The docs situation for Nix is admittedly not too rosy though.

brooksmtownsend (approved):

brooksmtownsend:
@theduke Got it running great on my machine, just wanted to verify it worked before merging 😄 . I'm a fan of the `nix` stuff, it seems like the "container" of package management solutions. Thanks for contributing this!

PR #129, opened 2021-04-27
Improved cursor position logic

### LINKED ISSUE #90: cursor can be in wrong position after resize
I use a tiling window manager and often move a window from small to full screen and back to small again. In wash, sometimes if I resize the window larger or smaller, the REPL cursor appears a line above or below the line where you would type the next command. It turns out that if the REPL window has had any wrap-around (such as with a 'ctl get inventory -long id-'), where the command wraps onto the second line, and then resize larger so that the command fits on one line, the cursor is drawn one line too low. It's as if it remembered the y position from before the resize and didn't recalculate it after the resize. If you go the other way around - long command on large window, then  shrink the app window so wrap is required, the cursor is one line too high.

After a brief search about tui resizing, I tried adding "terminal.autoresize()?;" at the beginning of draw_ui, but that didn't fix it. If you type "clear" in the REPL window, it does clear and the cursor position is in the correct place.

Not a high priority bug as it doesn't affect functionality, and using the "clear" command is a workaround.

brooksmtownsend:
Fixes #90 

This PR fixes 3 issues (not all documented):
1. #90, where resizing the terminal window does not recompute the position of the cursor
2. Depending on the width of the command, sometimes the cursor vertical position can be off by one
3. At smaller screen widths, the input window would scroll offscreen

In sum, the cursor should now always be in the correct position now.

### DIFF
--- a/src/up/repl.rs
+++ b/src/up/repl.rs
@@ -84,7 +84,7 @@ impl InputState {
                     1_u16
                 }
             })
-            .fold(0_u16, |acc, el| acc + el);
+            .sum();
         self.history_offset
     }
 }
--- a/src/up/repl.rs
+++ b/src/up/repl.rs
@@ -21,7 +21,6 @@ pub(crate) struct InputState {
     pub(crate) history_cursor: usize,
     pub(crate) input: Vec<char>,
     pub(crate) input_cursor: usize,
-    pub(crate) multiline_history: u16, // amount to offset cursor for multiline inputs
     pub(crate) input_width: usize,
     pub(crate) focused: bool,
     pub(crate) title: String,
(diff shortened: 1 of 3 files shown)


src/up/repl.rs
```diff
@@ -57,13 +59,34 @@ impl InputState {
...
+            .map(|h| {
+                let input_length = h.len() + WASH_PROMPT.len();
+                let multilines = input_length / self.input_width;
+                if multilines >= 1 && input_length != self.input_width {
+                    1_u16 + multilines as u16
+                } else {
+                    1_u16
+                }
+            })
+            .fold(0_u16, |acc, el| acc + el);
```
autodidaddict: Fix clippy complaint on this fold.

autodidaddict (dismissed):

src/up/repl.rs
brooksmtownsend: Done 👍🏻

autodidaddict (approved):

Issue #132, opened 2021-04-29
[BUG] Cursor will not compute vertical changes until resized


brooksmtownsend:
**Describe the bug**
An overlooked aspect of an optimization now causes the cursor to reset to line one in the REPL unless you resize the window. https://github.com/wasmCloud/wash/blob/main/src/up/repl.rs#L70-L72

**To Reproduce**
1. run `wash up`
2. type any command, hit `enter`
3. See the cursor return to the first line
4. To calculate the correct position, resize the window

**Expected behavior**
The vertical "history offset" should be calculated anytime the width of the screen changes or another command enters the history. For simplicity, this should just be reverted to calculate it every time the cursor position needs to be recalculated.

PR #135, opened 2021-05-03
Bump wasmcloud-host and control-interface

brooksmtownsend:
Bumping these versions in order to support M1 macbooks with `wasmtime-provider` 0.0.3


brooksmtownsend:
Also updating this to fix #132 , a pretty unfortunate bug that should be patched

ChrisRx (approved):
lgtm

Issue #139, opened 2021-05-06
[BUG] wash drain can panic if directory is not found


brooksmtownsend:
**Describe the bug**
If a user has not cached any actors or providers before, `wash drain` can panic stating that a directory cannot be found.

**To Reproduce**
1. Remove cache directories
1. Run `wash drain all`
1. See error

**Expected behavior**
`wash drain` should be able to gracefully handle the absence temporary cache directories.

brooksmtownsend:
This was fixed in #140

PR #140, opened 2021-05-09
Fix bug in wash drain

FlrnFrmm:
#139 
An error occurred due to an attempt to list a nonexistent dir.

Therefore now first filter the list of dirs to clear for the ones that exist.

Also refactored the function to make it shorter and more expressive.


brooksmtownsend (dismissed):
LGTM once tests pass 👍🏻

brooksmtownsend:
@FlrnFrmm looks like you may just need to run `cargo fmt` on the project

FlrnFrmm:
@brooksmtownsend Done and changed to the solution proposed by `cargo fmt`

brooksmtownsend (approved):

Issue #114, opened 2021-03-29
[FEATURE] Show Actor and Provider name on host inventory screen


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
The current host inventory looks like this:
![image](https://user-images.githubusercontent.com/12040685/112885530-e10ead00-909e-11eb-926b-97afa3a833d3.png)
This inventory looks great, but notice that actors and providers are only human-identifiable by their OCI reference. Actors and providers that are loaded as static plugins or directly from disk do not show OCI references. This causes a frustrating experience for both of those scenarios.

**Describe the solution you'd like**
The host inventory output should also show the `Name` field of the embedded claims, as the provider and actor claims both have this field. Before the host inventory will have access to this information, https://github.com/wasmCloud/wasmCloud/issues/24 must be completed.

**Additional context**
The host inventory output is already exceeding the width of an average terminal. With the addition of the `Name` column, we'll be even further off the screen if someone is developing on a smaller screen, or with a vertical monitor setup for example. This will need to be addressed.

Some potential solutions:
1. In tandem with #74 , we could show the first 5-10 characters of the public keys and allow for tab completion to fill out the public key. In the above example, if you were linking the HTTPSERVER provider, you could type `VAG3Q<tab>` and the rest of the provider key could autofill.
2. Omit certain fields (IDs, OCI reference) by default, but show indicator in UI that implies more information that can be accessed with additional flags

frigus02:
Hi there. I think showing the actor and provider names would be really useful. I thought I might give this a go and would like to clarify a few things.

- All tables have a max column width at the moment, including the host inventory table:
  
  https://github.com/wasmCloud/wash/blob/31b7c49531c2300f2f421955f8cd5a35d7af0cc3/src/ctl/output.rs#L202

  This sets the width per column. This table has 4 columns, so the max table width is 320. Was this intentional or should the total table width default to 80 instead?

- I noticed that the max column width is never actually specified, so it always uses the default of 80.

I tried to see what the tables look like if we would specify the width so that the entire table max width is equal to the available output width. This is the result (columns that exceed the width, break into multiple lines):

![image](https://user-images.githubusercontent.com/3579251/120829830-c3a20780-c555-11eb-8138-148ac312ecb6.png)

Would you be interested in a PR just for that (restraining the table width to the available output width)?

[...]

PR #144, opened 2021-06-05
Restrict tables to available output width

frigus02:
Part of #114

Restrict tables, that are printed to stdout, to the available output width. Currently that is `usize::MAX` for anything printed directly from wash cli and the output pane width inside the wash up REPL.

Restricting tables means columns shrink so all columns fit in one line. If a value inside a column is too long, it breaks into multiple lines inside the column only.

Example before (| is the terminal width):

    | Actor ID                                            Image Refere |
    | nce                                                              |
    | ABCDEFGHIJABCDEFGHIJABCDEFGHIJABCDEFGHIJABCDEFGHIJ  wasmcloud.az |
    | urecr.io/kvcounter:0.2.0                                         |

Example after (| is the terminal width):

    | Actor ID                           Image Reference               |
    | ABCDEFGHIJABCDEFGHIJABCDEFGHIJABC  wasmcloud.azurecr.io/kvcounte |
    | DEFGHIJABCDEFGHIJ                  r:0.2.0                       |

In a later PR we could change the behavior for overflowing column values. For example instead of wrapping we could show the first few characters followed by an ellipsis.

---

It's important for the public key to be easy to copy. This PR should only affect the output pane of `wash up` for now. The public key could already be broken into multiple lines here, making it impossible copy in one go. But this is likely to happen more often now.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -18,6 +18,7 @@ pub(crate) const WASH_LOG_INFO: &str = "WASH_LOG";
 pub(crate) const WASH_CMD_INFO: &str = "WASH_CMD";
 
 thread_local! {
+    /// Currently available output width can change when the user resizes their terminal window.
     static MAX_TEXT_OUTPUT_WIDTH: Cell<usize> = Cell::new(0);
 }
 
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,13 +1,12 @@
 use log::info;
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
-use std::borrow::Cow;
 use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
 use std::str::FromStr;
 use structopt::StructOpt;
-use term_table::{Table, TableStyle};
+use term_table::TableStyle;
 
 pub(crate) type Result<T> = ::std::result::Result<T, Box<dyn ::std::error::Error>>;
 
@@ -135,29 +134,7 @@ pub(crate) fn output_destination() -> OutputDestination {
     }
 }
 
-pub(crate) trait OutputWidthFormat: Send {
-    fn format(&self, output_width: usize) -> Cow<'_, str>;
-}
-
-impl OutputWidthFormat for &str {
-    fn format(&self, _output_width: usize) -> Cow<'_, str> {
-        (*self).into()
-    }
-}
-
-impl OutputWidthFormat for String {
-    fn format(&self, _output_width: usize) -> Cow<'_, str> {
-        self.into()
-    }
-}
-
-pub(crate) fn configure_table_style(table: &mut Table<'_>, columns: usize, max_table_width: usize) {
-    table.max_column_width = (max_table_width - 1) / columns;
-    table.style = empty_table_style();
-    table.separate_rows = false;
-}
-
-fn empty_table_style() -> TableStyle {
+pub(crate) fn empty_table_style() -> TableStyle {
     TableStyle {
         top_left_corner: ' ',
         top_right_corner: ' ',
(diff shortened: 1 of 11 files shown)


brooksmtownsend:
@frigus02 I ran the REPL off of your branch and overall it looks good, agreed that the ID collapse could happen before and this just makes it slightly more likely because it restricts each column. It _feels_ like there should be a better way to get the output width to the table format, in the current state the boxing does seem excessive.

I would like to see if an addition to the `OutputKind` enum could suffice here, e.g.:
```
/// Used for displaying human-readable output vs JSON format
#[derive(StructOpt, Debug, Copy, Clone, Serialize, Deserialize, PartialEq)]
pub(crate) enum OutputKind {
    Text(usize),
    Json,
}
```
where the usize is the width of the terminal 🤔 The usize could default to some value to indicate there is no max width, unless we're in the REPL where we do know the width of the terminal. I think this might be a simpler solution that's also relevant to the information we try to pass around with the Output* structs and enums. JSON should always be the minimized version, so it doesn't need to carry extra information. Thoughts?

frigus02:
Thanks for the review. I definitely agree with the excessive boxing. My first thought was something like you suggest but it seemed hard to access the output width outside the logging functions. That said, I don't think I tried too hard. Let me give this a proper go. I do like the `Text(usize)` in the `OutputKind` enum.

frigus02:
Alright, thanks for the suggestion @brooksmtownsend. This looks so much better. The only awkward piece is the thread local static for the output width. `OutputKind` is created by StructOpt/Clap and quite deeply nested. This was the quickest thing I could think of.

(I just appended commits for now. I usually find that easier to review. Happy to clean them up when we're happy with this, though.)

brooksmtownsend:
@frigus02 I think this looks a lot cleaner! I saw the thread local static, I think it does about the same thing that https://github.com/wasmCloud/wash/blob/main/src/util.rs#L14 does in terms of being available across the program. If you want to give that syntax a go, it feels a bit cleaner and it'll integrate nicely into what we have now (and save you the get/set functions). That should make it ready for review 😄 

Also, appending commits is fine, you're welcome to squash beforehand but we always squash into the repo per-pr, so it'll all condense into one commit anyways.

frigus02:
Yay. The problem with `OnceCell` here is, well, that is can only be written to once 🙂. But I can resize the terminal while the repl is open. In that case it would be nice if the available max width would change, too. With the thread local `Cell` this works.

I can change it to a `OnceCell` and/or add a comment to explain the tradeoff.

Something is still ever so slightly off with the tables. There are certain width where it doesn't seem to fit. It must be off by 1 or something. I'm going to try to find out what's wrong.

brooksmtownsend:
Ah good catch, I wasn't aware it could only be written to once (but it's in the name, makes sense). Go ahead and leave the thread_local, I'd prefer to support resizing over the life of the repl 👍🏻

frigus02:
Another option could be a `OnceCell` which holds a `Mutex`/`RwLock` or something similar. That would enable cross thread synchronization. It doesn't seem required at the moment.

I'm going to add a short comment explaining why this is a thread local and fix the table sizing issue. Then I'm going to mark this as ready for review.

src/util.rs
```diff
@@ -149,3 +180,48 @@ pub(crate) fn empty_table_style() -> TableStyle {
         horizontal: ' ',
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::configure_table_style;
+    use term_table::{row::Row, table_cell::TableCell, Table};
+
```
brooksmtownsend, line 188: 🏅 gold medal for including tests 🎉

autodidaddict:
cc @brooksmtownsend

brooksmtownsend (approved):
LGTM, works well in the terminal too. Should be a good foundation to get #114 fixed.

PR #145, opened 2021-06-14
Truncate id and add name to inventory output

### LINKED ISSUE #114: [FEATURE] Show Actor and Provider name on host inventory screen

frigus02:
Hi again. I started to play with truncating the actor/provider id and adding a name column. It feels tricky to strike the right balance here. I think ideally we don't truncate the id if there is enough space. This  seems okay to me.

- Truncate actor/provider id in host inventory to fit into one column

  - We use the available column width to truncate the id and add an ellipsis
  - Truncation only happens if the output width is restricted, which is currently only the case in the REPL
  - The flag `--output wide` also sets the output width to be unrestricted

- Add actor/provider name column to host inventory

Fixes #114

What do you think?

### DIFF
--- a/src/ctl/output.rs
+++ b/src/ctl/output.rs
@@ -1,5 +1,5 @@
 extern crate wasmcloud_control_interface;
-use crate::util::{format_output, OutputKind, WASH_CMD_INFO};
+use crate::util::{format_ellipsis, format_optional, format_output, OutputKind, WASH_CMD_INFO};
 use log::debug;
 use serde_json::json;
 use term_table::{row::Row, table_cell::*, Table};
@@ -198,6 +198,7 @@ pub(crate) fn hosts_table(hosts: Vec<Host>, max_width: usize) -> String {
 pub(crate) fn host_inventory_table(inv: HostInventory, max_width: usize) -> String {
     let mut table = Table::new();
     crate::util::configure_table_style(&mut table, 4, max_width);
+    let max_id_width = crate::util::get_max_column_width(&table, 0, true);
 
     table.add_row(Row::new(vec![TableCell::new_with_alignment(
         format!("Host Inventory ({})", inv.host_id),
@@ -232,18 +233,20 @@ pub(crate) fn host_inventory_table(inv: HostInventory, max_width: usize) -> Stri
             Alignment::Center,
         )]));
         table.add_row(Row::new(vec![
-            TableCell::new_with_alignment("Actor ID", 2, Alignment::Left),
+            TableCell::new_with_alignment("Actor ID", 1, Alignment::Left),
+            TableCell::new_with_alignment("Name", 1, Alignment::Left),
             TableCell::new_with_alignment("Image Reference", 2, Alignment::Left),
         ]));
         inv.actors.iter().for_each(|a| {
             let a = a.clone();
             table.add_row(Row::new(vec![
-                TableCell::new_with_alignment(a.id, 2, Alignment::Left),
                 TableCell::new_with_alignment(
-                    a.image_ref.unwrap_or_else(|| "N/A".to_string()),
-                    2,
+                    format_ellipsis(a.id, max_id_width),
+                    1,
                     Alignment::Left,
                 ),
+                TableCell::new_with_alignment(format_optional(a.name), 1, Alignment::Left),
+                TableCell::new_with_alignment(format_optional(a.image_ref), 2, Alignment::Left),
             ]))
         });
     } else {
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -165,6 +182,15 @@ pub(crate) fn configure_table_style(table: &mut Table<'_>, columns: usize, max_t
     table.separate_rows = false;
 }
 
+pub(crate) fn get_max_column_width(table: &Table<'_>, column_index: usize, padding: bool) -> usize {
+    let padding_width = if padding { 2 } else { 0 };
```
brooksmtownsend: I think the `padding` argument here could be renamed to something like `paddingOffset` or similar, to avoid this if statement which feels a bit hardcoded. It looks like you're only using it in one place, so you can simply pass 2 in as an adjustment value for that function call, and this lets us potentially change padding later.

src/util.rs
```diff
@@ -66,6 +66,7 @@ impl FromStr for OutputKind {
             "text" => Ok(OutputKind::Text {
                 max_width: get_max_text_output_width(),
             }),
+            "wide" => Ok(OutputKind::Text { max_width: 0 }),
```
brooksmtownsend: I believe your logic further in the code will assign `usize::MAX` in the case where the `max_width` is 0, but I think it would be more clear if you go ahead and assign that value here, in case the logic changes in the `configure_table_style` function

brooksmtownsend (changes requested):
This is a great iteration on the previous PR, good thinking to source the width of the column to the minimum size of the other fields, like the label key. Just a few minor comments.

IMO, the `--output wide` is enough for us to proceed with this PR without having the tab autocomplete, I will just have to take down a note to mention it in the `0.5.0` release.

src/util.rs
frigus02: Yeah, I made `0` mean "no max width". I copied this from clap:

> Sets the max terminal width at which to wrap help messages. Using 0 will ignore terminal widths and use source formatting.
> -- https://docs.rs/clap/2.33.3/clap/struct.App.html#method.max_term_width

It made sense to be because what else would 0 mean.

That said, using `usize::MAX` might make this more obvious. And `term_table` uses `usize::MAX` as their default values as well:

> The maximum width of all columns. Overridden by values in column_widths. Defaults to std::usize::max
> -- https://docs.rs/term-table/1.3.1/term_table/struct.Table.html#structfield.max_column_width

I going to change this to use `usize::MAX` and get rid of `0`.

src/util.rs
frigus02: Yeah. This is a bit awkward. The padding comes from `TableCell::pad_content` in `term_table`, which is a boolean. I wanted to give this parameter the same name (i.e. "pad_content") but messed it up somewhere along the way and used "padding" 🤦.

I wonder if there is a way to make this relation more obvious. Maybe it's best to remove the parameter entirely and do `- 2` where we call the function.

frigus02:
Thanks for the quick review @brooksmtownsend and sorry for the slow turnaround on my side. Please let me know if this looks okay now.

src/ctl/output.rs
```diff
@@ -199,6 +199,9 @@ pub(crate) fn host_inventory_table(inv: HostInventory, max_width: usize) -> Stri
     let mut table = Table::new();
     crate::util::configure_table_style(&mut table, 4, max_width);
 
+    let content_padding_width = 2;
+    let max_id_width = crate::util::get_max_column_width(&table, 0) - content_padding_width;
+
```
brooksmtownsend, line 205: Final request here, perhaps just add a quick note in a comment like "From term_table::TableCell::pad_content" just to show where you got the value from

brooksmtownsend (changes requested):
Only one small request to reflect what we discussed in the PR, otherwise LGTM 👍🏻

src/ctl/output.rs
frigus02: I think it might be nice if `term_table` would support different ways to deal with cell content that is too wide. At the moment it always wraps. If it supported truncation (with ellipsis), we wouldn't have to know about the internal padding calculation here. Maybe that's something for later, though 🙂

brooksmtownsend (approved):

PR #146, opened 2021-06-28
add credsfile argument to wash ctl subcommands

alsuren:
This is required if you want to use wash to control a cluster over NGS without using a leaf node.

It's mostly copy-paste from how wasmcloud does things.

I have not included the equivalent of wasmcloud's `--control-jwt` and `--control-seed` arguments, because I find .creds files easier to work with, and the jwt+seed separation was confusing to me in the beginning. I can add them if you want though.

### DIFF
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -5,7 +5,7 @@ use crate::util::{
 };
 use log::debug;
 use spinners::{Spinner, Spinners};
-use std::time::Duration;
+use std::{fs::File, io::Read, time::Duration};
 use structopt::StructOpt;
 use wasmcloud_control_interface::*;
 mod output;
@@ -43,6 +43,14 @@ pub(crate) struct ConnectionOpts {
     )]
     rpc_port: String,
 
+    /// JWT file for RPC authentication. Must be supplied with rpc_seed.
+    #[structopt(long = "rpc-jwt", env = "WASH_RPC_JWT", hide_env_values = true)]
+    rpc_jwt: Option<String>,
+
+    /// Seed file or literal for RPC authentication. Must be supplied with rpc_jwt.
+    #[structopt(long = "rpc-seed", env = "WASH_RPC_SEED", hide_env_values = true)]
+    rpc_seed: Option<String>,
+
     /// Credsfile for RPC authentication
     #[structopt(long = "rpc-credsfile", env = "WASH_RPC_CREDS", hide_env_values = true)]
     rpc_credsfile: Option<String>,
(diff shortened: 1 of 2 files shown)


brooksmtownsend (changes requested):
Thanks for the PR! This all looks good to me, I would like to see the same process that the wasmcloud binary supports regarding jwt + seed authentication, if you can add that then this looks good to go 😄

alsuren:
> I would like to see the same process that the wasmcloud binary supports regarding jwt + seed authentication

Done. I also expanded the documentation for the credsfile argument, to nudge people in that direction.

src/ctl/mod.rs
```diff
@@ -500,20 +516,54 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
...
+    match File::open(arg) {
+        Ok(mut f) => {
+            let mut value = String::new();
+            f.read_to_string(&mut value)?;
+            Ok(value)
+        }
+        Err(_) => Ok(arg.to_string()),
+    }
+}
+
```
brooksmtownsend: I think this belongs better in `src/util.rs`, since I can see us reusing this in the future for other arguments. To be honest, there are probably places in the code that would benefit from it now, but that's for another day 😄

brooksmtownsend (changes requested):
Just one small change requested for the helper function. Have you tested this with a credsfile / jwt+seed to ensure it properly connects to a nats connection with authentication required?

alsuren:
I did the utils.rs move and then a bit of manual testing fixed and fixed a bug (see also: https://github.com/wasmCloud/wasmCloud/pull/204).

src/util.rs
```diff
@@ -126,6 +128,18 @@ pub(crate) fn format_optional(value: Option<String>) -> String {
...
 
+/// Returns value from an argument that may be a file path or the value itself
+pub(crate) fn extract_arg_value(arg: &str) -> Result<String> {
+    match File::open(arg) {
+        Ok(mut f) => {
+            let mut value = String::new();
+            f.read_to_string(&mut value)?;
+            Ok(value)
+        }
+        Err(_) => Ok(dbg!(arg).trim_end_matches(char::is_whitespace).to_string()),
```
brooksmtownsend: Last request, let's remove the `dbg!` macro here so we don't get extracurricular output.

brooksmtownsend (changes requested):

src/util.rs
alsuren: Oops. I'd reverted it back to how it was in my editor and then forgot to save. Bit distracted today. Sorry about that.

brooksmtownsend (approved):

src/util.rs
brooksmtownsend: No problem 🙂 LGTM

PR #156, opened 2021-08-17
add flags to call actor so it can call a test actor

stevelr:
Signed-off-by: steve <dev@somecool.net>

New options for `wash call actor`:

 `--data`  allows using a json file for the parameter instead of key-value. This flag isn't necessarily test-specific. It could be used for anyone that uses `wash call actor`.
 `--test`  enables attempting to interpret the results as a Vec<TestResult> (defined by the wasmcloud-test-utils crate).
(I initially tried to add a variant to OutputKind, adding `-o test` instead of `--test`, but it would have affected more than six files, since OutputKind is used everywhere, but TestResult only applies to call actor, so it seemed cleaner to not add an enum variant).

### DIFF
--- a/src/call.rs
+++ b/src/call.rs
@@ -7,6 +7,7 @@ use serde_json::json;
 use structopt::clap::AppSettings;
 use structopt::StructOpt;
 use wasmbus_rpc::{core::WasmCloudEntity, Message, RpcClient};
+use wasmcloud_test_util::testing::TestResults;
 
 /// fake key (not a real public key)  used to construct origin for invoking actors
 const WASH_ORIGIN_KEY: &str = "__WASH__";
@@ -28,9 +29,10 @@ impl CallCli {
 
 pub(crate) async fn handle_command(cmd: CallCommand) -> Result<String> {
     let output_kind = cmd.output.kind;
+    let is_test = cmd.test;
     let res = handle_call(cmd).await;
     //TODO: Evaluate from_utf8_lossy and use of format here
-    Ok(call_output(res, &output_kind))
+    Ok(call_output(res, is_test, &output_kind))
 }
 
 #[derive(Debug, Clone, StructOpt)]
@@ -93,6 +95,14 @@ pub(crate) struct CallCommand {
     #[structopt(flatten)]
     pub(crate) output: Output,
 
+    /// optional json file data
+    #[structopt(short, long)]
+    pub(crate) data: Option<std::path::PathBuf>,
+
+    /// when invoking a test actor, interpret the response as TestResults
+    #[structopt(long)]
+    pub(crate) test: bool,
+
     /// Public key or OCI reference of actor
     #[structopt(name = "actor-id")]
     pub(crate) actor_id: String,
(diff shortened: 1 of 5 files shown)


src/call.rs
```diff
@@ -93,6 +95,14 @@ pub(crate) struct CallCommand {
     #[structopt(flatten)]
     pub(crate) output: Output,
 
+    /// optional json file data
+    #[structopt(short, long)]
+    pub(crate) data: Option<std::path::PathBuf>,
```
brooksmtownsend: I think we should indicate that the data is used for the operation, e.g. 
```
Optional JSON file to send as the operation payload
```

brooksmtownsend (changes requested):
small suggestion for help text, otherwise LGTM

src/util.rs
```diff
@@ -120,13 +120,19 @@ pub(crate) fn labels_vec_to_hashmap(constraints: Vec<String>) -> Result<HashMap<
...
+    let json = serde_json::from_str::<serde_json::Value>(payload)?;
     let payload = wasmbus_rpc::serialize(&json)?;
     Ok(payload)
 }
 
+/// transform msgpack bytes into a json string
+pub(crate) fn msgpack_to_json_val(msg: &[u8]) -> Result<serde_json::Value> {
+    let val: serde_json::Value = wasmbus_rpc::deserialize(msg)?;
+    Ok(val)
+}
```
brooksmtownsend: Does this... really work? I'm not accusing you of not testing it, I'm just really glad if you found the right way to deserialize arbitrary msgpack bytes since @autodidaddict and I spent hours trying to find the right way to do that in Rust

brooksmtownsend: edit: I tried it out, unfortunately looks like it's the same issue as we were running into before where the nested byte array messed with deserialization (this is the echo actor.)
```
➜ cargo run -- call MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 HandleRequest '{"method": "GET", "path": "/mycounter", "body":"", "queryString":"", "header":{}}' -o json
    Finished dev [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/wash call MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 HandleRequest '{"method": "GET", "path": "/mycounter", "body":"", "queryString":"", "header":{}}' -o json`
{"error":"deserialization: invalid type: byte array, expected any valid JSON value"}
```

Unless you have some ideas for the deserialization, I think you may have to revert to the lossy response for `json` as well

brooksmtownsend (changes requested):

src/util.rs
stevelr: fixed - see update

brooksmtownsend (approved):
Someone get @stevelr a 🏅 for the fix of the century

autodidaddict:
This is that moment from a sci-fi movie when everyone says it can't be done and @stevelr whips something together with bubble gum and a pine cone and saves the planet.

stevelr:
Those kinds of problems are my specialty!

PR #157, opened 2021-08-17
Support for application of manifest files

### LINKED ISSUE #158: Add support for manifest file application to wash
for automated deployments, we need manifest file support.

The most useful implementation, IMO, would be to implement the manifest file reader within wash, and have wash implement the control-interface commands to process it. This would be useful for developers that are using an already-running host and just created a new actor. An automated deployment of the host would require two commands - one to start the host, and one to start the wash-manifest processing. (or, to save one command, the host could spawn a process to do that). 

Theoretically, this could be implemented as a bash script calling wash repeatedly, but that wouldn't be a good way to teach people that manifest files are declarative.

autodidaddict:


src/ctl/mod.rs
```diff
@@ -569,6 +605,114 @@ pub(crate) async fn update_actor(cmd: UpdateActorCommand) -> Result<CtlOperation
...
+    let mut results = vec![];
+
+    for ld in hm.links.iter() {
+        match client
+            .advertise_link(
+                &ld.actor,
+                &ld.provider_id,
+                &ld.contract_id,
+                ld.link_name.as_ref().unwrap_or(&"default".to_string()),
+                ld.values.clone().unwrap_or(HashMap::new()),
```
brooksmtownsend: Clippy wants you to use `unwrap_or_default` here

brooksmtownsend (changes requested):
all looks good, only request is a clippy linting thing

brooksmtownsend (approved):

PR #160, opened 2021-08-26
Support querying, deleting link definitions

brooksmtownsend:
Depends on https://github.com/wasmCloud/control-interface/pull/9

This PR includes support for the following commands, and breaks the API for `ctl link` (we have not published `0.6.0` yet, so this is okay)
1. `wash ctl link put <args>` to put a link definition
2. `wash ctl link del <args>` to delete a link definition
3. `wash ctl link query` to retrieve the list of current link definitions

Tests will fail as the control interface is not available yet, the above PR must be merged and then tests can be re-run

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -227,6 +227,12 @@ dependencies = [
 
 [[package]]
 name = "base64"
+version = "0.12.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


autodidaddict (dismissed):
This build is broken because of the relative path reference. Everything else looks fine, but that'll need to be fixed prior to release.

autodidaddict (approved):

PR #165, opened 2021-09-09
Issue warning message for 'wash reg push' to localhost without --insecure

ChrisWhealy:
Signed-off-by: Chris Whealy <chris.whealy@red-badger.com>

If the `wash reg push localhost:nnnn ...` command is issued without the `--insecure` option, then the following, somewhat cryptic, error message is received
```
Error: error sending request for url (https://localhost:5000/v2/): error trying to connect: record overflow
```
This PR simply adds an explanatory warning message in such cases

### DIFF
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -1,6 +1,6 @@
 extern crate oci_distribution;
 use crate::util::{format_output, Output, OutputKind};
-use log::{debug, info};
+use log::{debug, info, warn};
 use oci_distribution::client::*;
 use oci_distribution::secrets::RegistryAuth;
 use oci_distribution::Reference;
@@ -304,6 +304,10 @@ fn validate_provider_archive(
 }
 
 pub(crate) async fn handle_push(cmd: PushCommand) -> Result<String, Box<dyn ::std::error::Error>> {
+    if cmd.url.starts_with("localhost:") && !cmd.opts.insecure {
+        warn!(" Unless an SSL certificate has been installed, pushing to localhost without the --insecure option will fail")
+    }
+
     let spinner = match cmd.output.kind {
         OutputKind::Text => Some(Spinner::new(
             &Spinners::Dots12,


src/reg.rs
```diff
@@ -304,6 +304,10 @@ fn validate_provider_archive(
 }
 
 pub(crate) async fn handle_push(cmd: PushCommand) -> Result<String, Box<dyn ::std::error::Error>> {
+    if cmd.url.starts_with("localhost:") && !cmd.opts.insecure {
+        warn!(" Unless an SSL certificate has been installed, pushing to localhost without the --insecure option will fail")
+    }
+
```
brooksmtownsend, line 310: I like this to cover the common use case of the local docker registry, perhaps in the future with some more testing we can find a solution to give a better warning message. Since it comes from `oci-distribution`, I think this will work well for now.

brooksmtownsend (approved):
LGTM

src/reg.rs
alsuren: Yeah, this is a "cover the common use case for now" PR. We had a go at patching out the error in `oci-distribution` and it ended up looking a bit ugly.

We also looked into how `docker push` does things. It creates a list of urls to try. If the hostname resolves to 127.*.*.* then it adds an http:// url to the front of the list.

(relevant bits of code for my own future reference:
https://github.com/moby/moby/blob/1e71c6cffedb79e3def696652753ea43cdc47b99/registry/config.go#L150-L155
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/config.go#L258
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/config.go#L258
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/service.go#L264
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/service.go#L268
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/service_v2.go#L10
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/registry/service.go#L283
https://github.com/moby/moby/blob/5176095455642c30642efacf6f35afc7c6dede92/distribution/push.go#L52 
)

I think that this logic could probably be added into `oci-distribution`, but a warning is good enough for now.

Issue #163, opened 2021-09-04
add 'gen' command to wash to trigger code generation


stevelr:
wash needs an additional cli command to trigger code generation. For rust interface crates, there's a build.rs that invokes  libraries to do generation during a cargo build, but there isn't a way from the command line for someone to trigger code generation. The use cases for this are:
- generate html documentation
- generate rust source file in some situations when it's been deleted and build.rs dosn't create it
- override default parameters during rust code generation
- generate source files for other target languges

stevelr:
fixed in https://github.com/wasmCloud/wash/pull/169

PR #169, opened 2021-09-15
Gen command

stevelr:
Add 'gen' command - fixes wash #163 
Update dependencies: upgraded oci_distribution from 0.6.0 to 0.7.0, cargo from 0.55 to 0.56
Bumped wash crate to 0.6.1

### DIFF
--- a/src/smithy.rs
+++ b/src/smithy.rs
@@ -1,7 +1,9 @@
 //! smithy model lint and validation
 //!
+use crate::generate::emoji;
 use anyhow::anyhow;
 use atelier_core::model::Model;
+use console::style;
 use std::path::PathBuf;
 use structopt::StructOpt;
 use weld_codegen::{
@@ -28,7 +30,7 @@ pub(crate) struct ValidateCli {
     opt: ValidateOptions,
 }
 
-/// Generate using codegen.toml
+/// Generate code from smithy IDL files
 #[derive(Debug, StructOpt, Clone)]
 #[structopt(name = "gen")]
 pub(crate) struct GenerateCli {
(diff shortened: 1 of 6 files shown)


src/smithy.rs
```diff
@@ -185,3 +231,65 @@ fn select_config(opt_config: &Option<PathBuf>) -> Result<CodegenConfig, anyhow::
...
+    let mut config = select_config(&opt.config)?;
+    if !opt.lang.is_empty() {
+        config.output_languages = opt.lang.clone()
+    }
+    let mut input_models = Vec::new();
+    std::mem::swap(&mut config.models, &mut input_models);
+    let model = build_model(opt.input, input_models, config.base_dir.clone(), verbose)?;
+
+    let templates = if let Some(ref tdir) = opt.template_dir {
+        println!("importing templates from {}", &tdir.display());
```
autodidaddict: this looks like a debug statement. Does this still belong in the final code?

src/smithy.rs
```diff
@@ -27,6 +28,14 @@ pub(crate) struct ValidateCli {
     opt: ValidateOptions,
 }
 
+/// Generate using codegen.toml
```
autodidaddict: Generate what? I think this help line (which shows up when you just type `wash` with no parameters) should indicate at least the category of things that can be generated here.

autodidaddict (dismissed):
Just a recommendation on a doc change and removing what looks like a debug output line

src/smithy.rs
stevelr: it's useful output, but I changed it to improve the formatting

src/smithy.rs
stevelr: updated to "Generate code from smithy IDL files"

brooksmtownsend (approved):
LGTM

PR #171, opened 2021-09-17
Use longer timeout or start actor and provider

### LINKED ISSUE #159: Use multiple of timeout delay for start actions
This issue was created because our original plan of doing an "early ack" on the start actor and start provider handlers in the OTP host won't work (because the first action that takes place in both of those handlers is the download of OCI bytes). As such, the best alternative is to make it so that wash is aware of this potential delay and uses a multiple of the RPC timeout when invoking start actor and start provider.

brooksmtownsend:
Fixes #159 
This PR separates the timeouts for starting an actor and a provider into `timeout` and `auction_timeout`. This allows quick response from auctions as well as waiting long enough for actors and providers to start.

Contrary to my recommendation in the issue, I did find occasionally that providers can take longer than 30s to download, validate, and start. By default, I set that to 60s to address that. 
Signed-off-by: Brooks Townsend <brooksmtownsend@gmail.com>

### DIFF
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -302,6 +302,10 @@ pub(crate) struct StartActorCommand {
     /// Constraints for actor auction in the form of "label=value". If host-id is supplied, this list is ignored
     #[structopt(short = "c", long = "constraint", name = "constraints")]
     constraints: Option<Vec<String>>,
+
+    /// Timeout to await an auction response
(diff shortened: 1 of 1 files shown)


autodidaddict (dismissed):

autodidaddict (dismissed):

brooksmtownsend:
Going for a new record of approvals on this PR

autodidaddict (approved):

PR #172, opened 2021-09-17
Update wasmbus rpc

stevelr:
- update wasmbus-rpc to 0.4
- change git2 to use vendored-libgit2 feature to address bug seen here: (which I encountered on my machine while running 'wash new ...') https://github.com/killercup/cargo-edit/issues/487
- update dirs dependency
-


brooksmtownsend (approved):
LGTM

PR #173, opened 2021-09-18
Feature: add reg ping to test if oci url is valid

stevelr:
Add reg ping command to test if an oci url contains an artifact.
We can use this for build automation to see if a version has been published. It's lighter weight than than reg pull, since it only downloads the manifest, not the entire artifact.

I did not bump the crate version in this PR, in case this is merged at the same time as PR #172 

Ping does not include the -o output format option because when json is enabled (for reg pull, for example), json is not used to encode error outputs anyway. The simplest way to use in scripts is to check exit status: 0 for success, 1 for failure.


stevelr:
When I created this PR it wasn't originally based on main correctly (it was a branch off pr 172). I corrected that by reversing those changes in this pr. So when/if this is merged, do a squash merge to clean up the commit history.

autodidaddict (approved):

PR #176, opened 2021-09-23
Require cluster seed for rpc invocations

brooksmtownsend:
Fixes #176 
Tested this using the latest built tarball, works great. One caveat is for the Host, you need to supply both the `WASMCLOUD_CLUSTER_SEED` environment variable and the `WASMCLOUD_CLUSTER_ISSUERS` which is just the issuer that matches the seed. I'm going to file an issue for this if this is required and we can discuss there on the OTP repo https://github.com/wasmCloud/wasmcloud-otp/issues/224

### DIFF
--- a/.github/workflows/release_snap.yml
+++ b/.github/workflows/release_snap.yml
@@ -29,7 +29,7 @@ jobs:
           args: --all-features
 
   snap_release:
-    depends_on: [cargo_check, clippy_check]
+    needs: [cargo_check, clippy_check]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


autodidaddict (dismissed):

autodidaddict:
@brooksmtownsend looks like this failed a format check.

Also, it's worth remembering that the list of valid issuers is different than the seed parameter on the host. In other words, each host signs with _one_ key, but can accept _one or more_ valid issuers. This gives operations the ability to use 1 seed key or multiple.

autodidaddict (commented):
hah, rust fmt wont' let you leave trailing whitespace on the end of a line.

brooksmtownsend:
@autodidaddict Yeah it took a bit of looking around and I understand that now, I think this is something tangental but it is actually a bug, filed the issue here: https://github.com/wasmCloud/wasmcloud-otp/issues/224

autodidaddict (approved):

PR #183, opened 2021-09-28
fix(generator): Fixes windows path handling

thomastaylor312:
When creating a new project, a subdirectory would be pushed with its
default `/` instead of the operating system specific slash. This fixes
the issue and I did double check that it should work on *nix systems
too


autodidaddict (approved):
I've always defaulted to using `PathBuf` when building CLI apps in Rust. Now I know why it's the defacto over string.

brooksmtownsend (approved):

PR #196, opened 2021-11-04
[FEATURE] wash Context

### LINKED ISSUE #194: [FEATURE] Give wash the ability to select and manipulate host configuration contexts
Developers may wish to easily and rapidly switch between various lattice connection and security settings. If they want to switch keys/issuers or switch from being configured to run against `localhost` versus running directly against a specific host, then using this feature will enable that functionality. The ultimate goal is to improve the developer experience by making it easy to manage multiple lattice/cluster configurations.

When a context is select/made the default, then it will copy whatever values are stored for that context into the `~/.wash/host_config.json` file. An index of available contexts can be found in the `~/.wash/contexts/index.json` file, and each context can have its own separate file (possibly named the same as the context name, e.g. `myconfig.json`).

**Nice to Have** - if everything you need to manage a list of contexts is easily found in a single sub-directory under `~/.wash/` then developers can easily copy the entire contents of that directory to other machines to take their context configurations with them.

The current version of the wasmCloud host runtime examines the `~/.wash/host_config.json`. If the file exists, then it will load its contents into configuration, which can then be overridden by a "local" `host_config.json` file and then subsequently overridden via environment variables.

If a developer does nothing, and just manages contexts, then the OTP host will _always_ be able to find and read the currently selected lattice and host settings.

## New Commands
* `wash ctx` - Get help for this new set of subcommands. 
* `wash ctx list` - Show all stored contexts, highlights the default/current context
Example output:
```
Bob
Alice (current)
Fred
```
* `wash ctx del {name}` - Delete a named context
* `wash ctx new {name}` - Create a new context (name can also be prompted rather than supplied)
* `wash ctx default {name}` - Sets the default/current context
* `wash ctx edit {name}` - Opens the developer's favorite editor (possibly obtained via env var?) to edit the context. If no editor is configured, then we can just display the full path to the file name the developer needs to edit.

## Additional Context
It's likely that `wash ctx new` will supersede the other potential feature of having a `wash new host` command. This is more technically appropriate as creating a new "host" implies that the host ID is fixed, which is no longer possible in the current version of wasmCloud OTP.

brooksmtownsend:
Fixes #194 

I've implemented the following commands:
1. `new`
2. `default`
3. `edit`
4. `del`
5. `list`

TODO list:
- [x] Ensure context can be loaded correctly from other modules in the CLI
- [x] Use context values to connect with `ctl` if present
- [x] Use context values to connect with `call` if present
- [x] Add prompts to `new` to allow user to supply values interactively (?)
- [x] Write some tests
- [x] modify the `host_config.json` to respond to default context changes
- [x] Ensure this works on windows
- [x] Ensure this handles `contexts` dir not existing


src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    } else {
+        None
+    }
+}
+
+/// Takes an absolute path to a JSON file and returns an IndexJson object
+fn get_index(context_dir: PathBuf) -> Result<IndexJson> {
+    let mut buf = vec![];
+    let mut f = std::fs::File::open(context_dir.join(INDEX_JSON))?;
+    f.read_to_end(&mut buf)?;
```
thomastaylor312: Nit: `serde_json` has a Reader type you can use rather than reading into memory: https://docs.rs/serde_json/1.0.68/serde_json/fn.from_reader.html

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    };
+
+    if let Ok(choice) = prompt_for_choice(&entry, prompt) {
+        choices.get(choice).map(|c| c.to_string())
+    } else {
+        None
+    }
+}
+
+/// Takes an absolute path to a JSON file and returns an IndexJson object
```
thomastaylor312: Nit: This doc comment should say that it takes a path to a _directory_ containing an index.json

src/ctx/context.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// Name of the default context
+    default: String,
+}
+
+impl IndexJson {
+    pub fn new(default: String) -> Self {
+        IndexJson { default }
+    }
+
+    pub fn default(&self) -> &String {
```
thomastaylor312: Rather than this being an accessor (that shares a name with a "reserved" function in Rust), I'd just recommend making the `default` data member `pub`

src/ctx/context.rs
```diff
@@ -0,0 +1,83 @@
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+
+#[derive(Clone, Deserialize, Serialize, Debug)]
+pub(crate) struct IndexJson {
```
thomastaylor312: Ok, this name is suuuuuuuper confusing. It took me a couple of minutes to figure out what this was doing (and that it is related to a specific file name). Maybe something like `DefaultContext`

src/ctx/context.rs
```diff
@@ -0,0 +1,83 @@
...
+            rpc_port: default_nats_port(),
+            rpc_jwt: None,
+            rpc_seed: None,
+            rpc_credsfile: None,
+            rpc_timeout: default_timeout(),
+        }
+    }
+}
+
+fn default_nats_host() -> String {
```
thomastaylor312, line 130: These should all be `const` unless you actually need to do some lookups or manipulation. The `String`s you need can be converted from the `&str` when used in the default

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    #[structopt(flatten)]
+    pub(crate) output: Output,
+}
+
+#[derive(StructOpt, Debug, Clone)]
+pub(crate) struct DelCommand {
+    /// Location of context files for managing. Defaults to $WASH_CONTEXTS ($HOME/.wash/contexts)
+    #[structopt(long = "directory", env = "WASH_CONTEXTS", hide_env_values = true)]
+    directory: Option<PathBuf>,
```
thomastaylor312, line 84: Can structopt do global flags? This is in every subcommand

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    /// Name of the context to edit, if not supplied the user will be prompted to select a context
+    #[structopt(name = "name")]
+    pub(crate) name: Option<String>,
+
+    /// Your terminal text editor of choice. This editor must be present in your $PATH, or an absolute filepath.
+    #[structopt(short = "e", long = "editor", env = "EDITOR")]
+    pub(crate) editor: String,
+}
+
+/// Lists all JSON files found in the context directory, with the exception of `index.json`
```
thomastaylor312, line 141: This help text should definitely be part of the command

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    if let Ok(choice) = prompt_for_choice(&entry, prompt) {
+        choices.get(choice).map(|c| c.to_string())
+    } else {
+        None
+    }
+}
+
+/// Takes an absolute path to a JSON file and returns an IndexJson object
+fn get_index(context_dir: PathBuf) -> Result<IndexJson> {
```
thomastaylor312: Nit: I'd recommend `&Path` so callers don't have to clone. You end up allocating a new `PathBuf` anyway with `.join` below

thomastaylor312: Pretty much anywhere you take a `PathBuf` you can use this, as all of the `io` functions take `AsRef<Path>`

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    #[structopt(short = "e", long = "editor", env = "EDITOR")]
+    pub(crate) editor: String,
+}
+
+/// Lists all JSON files found in the context directory, with the exception of `index.json`
+/// Being present in this list does not guarantee a valid context
+fn handle_list(cmd: ListCommand) -> Result<String> {
+    let dir = context_dir(cmd.directory)?;
+    let index = get_index(dir.clone()).ok();
+    let contexts: Vec<String> = context_filestems_from_path(get_contexts(dir.clone())?);
```
thomastaylor312: Super Nit: shouldn't need a hardcoded type here

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    let output_contexts = match cmd.output.kind {
+        OutputKind::Text if index.is_some() => contexts
+            .iter()
```
thomastaylor312: Nit: `IntoIter` to consume the data that you are dropping anyway in here

src/ctx/mod.rs
```diff
@@ -0,0 +1,385 @@
...
+    serde_json::to_writer(&File::create(context_path)?, &index).map_err(|e| e.into())
+}
+
+/// Loads the default context, according to index.json, into a WashContext object
+pub(crate) fn get_default_context(context_dir: PathBuf) -> Result<WashContext> {
+    let index = get_index(context_dir.clone())?;
+    load_context(context_path_from_name(context_dir, index.default()))
+}
+
+/// Load a file and return a Result containing the WashContext object. The `context_path` arg must be an absolute path
```
thomastaylor312: Does it have to be absolute? I don't think `open` requires that

thomastaylor312 (commented):
Functionality and structure looks pretty good with my first pass. Left random, hopefully helpful, notes along the way as I read through the code

src/ctx/mod.rs
brooksmtownsend: Wow I was looking for this to avoid the `f.to_owned()` there, nice.

src/ctx/mod.rs
brooksmtownsend: I'm not exactly sure what I was thinking here after re-reading it. I might've been confused about passing the name of a context rather than the path to it...

Regardless, I'll remove that :)

src/ctx/mod.rs
brooksmtownsend: I've looked at putting it on the parent `Struct` `CtxCli` before, since the command is a field there, however this requires you to specify the flag _before_ the subcommand.

e.g. instead of `wash ctx list --directory $HOME/.wash/contexts` you'd have to do `wash ctx --directory $HOME/.wash/contexts list`

brooksmtownsend: And IMO one of the biggest pain points of CLIs are positional arguments or flags that need to be in the correct place, so I figured a bit of copy-paste saves the user the pain

thomastaylor312: Yeah, that sounds fine to me

brooksmtownsend:
With the 2 recent commits, `wash ctx` now attempts to find the host configuration file in `~/.wash/host_config.json` and transform it into a context named `host_config`. This context will be set as the default if there is not already one present _and_ will be updated each time the user runs a control interface/RPC command to ensure that `wash` always connects to the local host by default and responds to any configuration changes.

So, to recap, the user launches an OTP host with custom configuration using `./bin/wasmcloud_host start`. Since this is the first run, a `host_config.json` file is created in the same directory the user ran that command, as well as in `~/.wash/` for later usage.

Now, once the host is running, the user runs `wash ctl get hosts` in order to find the host ID. Because no contexts exist, we find the `host_config.json` in `~/.wash/host_config.json` and create a context from the values there and immediately use it for that command. The net result here that anytime a local user changes the host configuration and launches a local host, `wash` is immediately configured to interact with that host.

As mentioned above, `wash ctx` also has the ability to manage different contexts for remote clusters of hosts, and those can be managed manually with the other subcommands.

brooksmtownsend:
~Converting this back to draft as there are some more kinks to work out~
Ready for review with the caveat that the implementation is based on the assumptions in https://github.com/wasmCloud/wasmcloud-otp/pull/264 are valid and accepted

stevelr:
will this work in the use case where you want to run more than one host on the same machine - simultaneously?

stevelr:
Based on the order of precedence defined in https://github.com/wasmCloud/wasmcloud-otp/issues/245 multiple hosts should work. If they are read only.

If the home dir file is overwritten, it doesn't work.

brooksmtownsend:
> will this work in the use case where you want to run more than one host on the same machine - simultaneously?

@stevelr It will work if you want to run more than one host on the same machine, if they have the same config. There's nothing stopping you from launching multiple hosts with different configurations, however the `wash ctx` managed `host_config` context will always be pointing to the _latest_ launched host configuration. In this case I'd recommend the user managing contexts on their own rather than using the managed one

autodidaddict (changes requested):
Just to avoid deserialization warnings, can you add the `js_domain` field to your context struct?

brooksmtownsend:
> Just to avoid deserialization warnings, can you add the `js_domain` field to your context struct?

@autodidaddict I certainly can, it won't be needed for any connectivity configurations will it? I did test that the Context struct would be able to deserialize from the `host_config.json` even if it included extra fields, e.g. `allowed_insecure`

autodidaddict:
Ok if it deserializes fine then no need. That field is not needed to contact hosts

autodidaddict (approved):

PR #198, opened 2021-11-11
Adopt common naming convention with OTP host

### LINKED ISSUE #195: [FEATURE] Create parity in environment variables with OTP host
`wash`, as a command-line interface, requires a variety of arguments that accept environment variables as alternatives. Things like `WASH_ISSUER_SEED` allow the user to pass in sensitive values from the environment rather than from a shell, and `WASH_RPC_HOST` prevent you from having to specify `--rpc-host 127.0.0.1` for each operation. These values are specific to `wash`.

Ideally, the environment variables that we use for wash should be the same that are used for the OTP wasmCloud host (and future hosts.) If you were able to specify `WASMCLOUD_RPC_HOST` and use it for both the host and `wash`, then you could start NATS and the host in the background and immediately issue `wash` commands to that host.

This is a list of what I could find in `wash` that could be changed to better match the host:
|  wash | wasmcloud-otp  | Notes|
|---|---|---|
|  WASH_RPC_HOST | WASMCLOUD_RPC_HOST  | |
|  WASH_RPC_PORT | WASMCLOUD_RPC_PORT  | |
|  WASH_RPC_JWT | WASMCLOUD_RPC_JWT | |
|  WASH_RPC_SEED | WASMCLOUD_RPC_SEED | |
| WASH_RPC_TIMEOUT  |  WASMCLOUD_RPC_TIMEOUT_MS | |
| WASH_RPC_NSPREFIX  |  WASMCLOUD_LATTICE_PREFIX | `nsprefix` should be renamed to `lattice_prefix` in my opinion around the codebase to better fit our naming scheme |
|  WASH_CTL_HOST | WASMCLOUD_CTL_HOST  | |
|  WASH_CTL_PORT | WASMCLOUD_CTL_PORT  | |
|  WASH_CTL_JWT | WASMCLOUD_CTL_JWT  | |
|  WASH_CTL_SEED | WASMCLOUD_CTL_SEED | |
| WASH_CTL_NSPREFIX  |  WASMCLOUD_LATTICE_PREFIX | `nsprefix` should be renamed to `lattice_prefix` in my opinion around the codebase to better fit our naming scheme |
| _no variable ATM_ | WASMCLOUD_CTL_TIMEOUT_MS | Though neither of these variables are present, there are some command line arguments in `wash` that should be configured with an environment variable `WASMCLOUD_CTL_TIMEOUT_MS` |

This work may be essentially encapsulated in #194, if environment variables are scarcely used. However, in the case that the user does not want to use wash contexts to manage configuration, this would still be beneficial.

**Additional context**
Many of these 
This feature could cause confusion in the case that you desire to start local wasmCloud hosts with different settings than what you use for `wash`. Off the top of my head, this would mostly apply when launching local hosts and using `wash` to administer different hosts, which would be solved again by #194 which would let you select different sets of context for `wash`.

byblakeorriver:
Changes the environment variable names in the wash CLI to match the ones that would be set in OTP host. 
Also, variable names with more than a prefix change had their underlying struct fields changed to match their new names.
Also changed the timeout conventions to use milliseconds instead of seconds.

Should resolve: https://github.com/wasmCloud/wash/issues/195

### DIFF
--- a/src/call.rs
+++ b/src/call.rs
@@ -48,7 +48,7 @@ pub(crate) struct ConnectionOpts {
         short = "r",
         long = "rpc-host",
         default_value = "0.0.0.0",
-        env = "WASH_RPC_HOST"
+        env = "WASMCLOUD_RPC_HOST"
     )]
     rpc_host: String,
 
@@ -57,16 +57,16 @@ pub(crate) struct ConnectionOpts {
         short = "p",
         long = "rpc-port",
         default_value = "4222",
-        env = "WASH_RPC_PORT"
+        env = "WASMCLOUD_RPC_PORT"
     )]
     rpc_port: String,
 
     /// JWT file for RPC authentication. Must be supplied with rpc_seed.
-    #[structopt(long = "rpc-jwt", env = "WASH_RPC_JWT", hide_env_values = true)]
+    #[structopt(long = "rpc-jwt", env = "WASMCLOUD_RPC_JWT", hide_env_values = true)]
     rpc_jwt: Option<String>,
 
     /// Seed file or literal for RPC authentication. Must be supplied with rpc_jwt.
-    #[structopt(long = "rpc-seed", env = "WASH_RPC_SEED", hide_env_values = true)]
+    #[structopt(long = "rpc-seed", env = "WASMCLOUD_RPC_SEED", hide_env_values = true)]
     rpc_seed: Option<String>,
 
     /// Credsfile for RPC authentication. Combines rpc_seed and rpc_jwt.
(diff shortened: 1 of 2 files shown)


src/call.rs
```diff
@@ -57,40 +57,40 @@ pub(crate) struct ConnectionOpts {
...
     /// See https://docs.nats.io/developing-with-nats/security/creds for details.
     #[structopt(long = "rpc-credsfile", env = "WASH_RPC_CREDS", hide_env_values = true)]
     rpc_credsfile: Option<String>,
 
-    /// Namespace prefix for wasmcloud command interface
+    /// Lattice prefix for wasmcloud command interface
     #[structopt(
-        short = "n",
-        long = "ns-prefix",
+        short = "x",
```
byblakeorriver, line 79: Used `x` here. I still need to verify that `l` cannot be used. I saw that it was used here: https://github.com/wasmCloud/wash/blob/main/src/ctl/mod.rs#L193

brooksmtownsend (changes requested):
This code LGTM, though you'll have to sign the DCO for checks to pass. @byblakeorriver the instructions for how to do this can be found in the `Details` section of the DCO check

autodidaddict:
LGTM as well, just waiting for the DCO check

byblakeorriver:
Cool, got that part fixed. @brooksmtownsend, are you all ok with using x as the short code for lattice-prefix?

brooksmtownsend:
> Cool, got that part fixed. @brooksmtownsend, are you all ok with using x as the short code for lattice-prefix?

Yeah X is fine with me, as far as I know it's only used on the "claims" subcommand so you should be good there

brooksmtownsend (approved):

Issue #192, opened 2021-10-12
[FEATURE] Give a hint for commonly mistaken inputs


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
Because we have a general idea about our inputs for wash, e.g. an actor ID, provider ID, and host ID should all be 56 character nkeys, we can give the user a warning if the input seems misshapen. Because command line tools have positional arguments it's also common for users to input arguments in the wrong order, we can detect if a nkey is provided in place of a contract ID, for example.

**Describe the solution you'd like**
We know the most about our inputs that are nkeys, and we should be able to provide warnings if the wrong key is provided (e.g. we expect an actor key, but we're provided with an nkey starting with VADK... which we know is a provider key)

We can also verify other inputs placed positonally around nkeys, for example if a contract ID is provided as a 56 character string, we can warn the user that it looks like an nkey (of course, it's just a warning, and if some user creates a 56 character contract ID that's fine)

brooksmtownsend:
Thanks to #199 we got some great progress here. One other place we can build off of that PR is the cluster seed arg for `wash call`

StuartHarris:
Oh yeah :-) forgot than one. Will add

brooksmtownsend:
With the vast number of improvements in the linked PRs above, I'm marking this as complete. Thanks all for the contributions!

PR #202, opened 2021-11-19
Cluster seed validation for CLI args (and interactive)

StuartHarris:
This PR implements some more of https://github.com/wasmCloud/wash/issues/192 by introducing `ClusterSeed` for the `wash call` command args and interactive input.

Note there is a remaining question about what a default ClusterSeed should be (if anything), as this was previously an empty string (which won't parse now).

- [x] introduce `ClusterSeed`
- [x] write some unit tests for `Id` and `Seed` parsing
- [x] revisit `ClusterSeed::default()` (wraps `String::default()`)

Also added a `.rustfmt.toml` to allowing merging imports, but happy to revert this if you think the imports are now "too merged" :-)


StuartHarris:
Accidentally pulled some older commits onto this branch. Ooops! Just trying to fix it now :-)

StuartHarris:
@brooksmtownsend opening this up for review, but would love to get your opinion on what `ClusterSeed::default()` should be. Before, it was an empty string (although not in a happy code path), so it may not matter what it is in reality (but I had to make it something that would now parse, so I made is `SC` + 54 x `0` :-)

autodidaddict (dismissed):
LGTM, though I'd like @brooksmtownsend to double-check this to make sure it works fine in conjunction with the automatic configuration reading `~/.wash/host_config.json`

src/id.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self(parse(s, PREFIX, true)?))
+    }
+}
+
+fn parse(value: &str, prefix: char, is_seed: bool) -> Result<String, IdParseError> {
+    let count = value.chars().count();
+    if count != 56 {
+        return Err(IdParseError::WrongLength(count));
+    }
```
brooksmtownsend: I did a bit of investigating while testing, looks like the seed keys are 58 characters and the public keys are 56 characters. You may need to define the count as separate constants and create that after reading the prefix

src/id.rs
```diff
@@ -0,0 +1,151 @@
...
+    } else {
+        Err(IdParseError::WrongKeyType {
+            found: value.chars().take(prefix.chars().count()).collect(),
+            expected: prefix,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
brooksmtownsend, line 79: You'll probably need to add 2 more `0`s here for the proper test after making the above change

src/call.rs
```diff
@@ -288,7 +293,7 @@ async fn rpc_client_from_opts(
                     error!(
                         "No cluster seed provided and no context available, this RPC will fail."
                     );
-                    "".to_string()
+                    ClusterSeed::default()
```
brooksmtownsend, line 296: Weirdly enough, this makes actor calls without a cluster seed give me this error:
```
Error invoking actor: Invalid seed length: Bad seed length: 56
```

This is an error from the NATS client when providing the seed, I don't have a particularly strong preference if you want to handle that as an error or simply use the blank string instead.

brooksmtownsend (changes requested):
As far as I can tell this looks good with `wash ctx`. Turns out seed keys are 58 characters so valid keys won't parse with this iteration. After doing some testing I found a few more requests.

src/id.rs
StuartHarris: Thanks for catching this @brooksmtownsend! This should be fixed now.

src/call.rs
StuartHarris: This should be 58 now, so I'd be interested in what the NATS error becomes (I wasn't sure how to test this).

brooksmtownsend: The error shown now is:
```
Error invoking actor: Codec failure: Data encoding failure: invalid symbol at 2
```

Which is unfortunately not very descriptive, and I believe due to the `0` in the default cluster seed 🤔 
Why don't we just change the `ClusterSeed::default()` to be a legitimate cluster seed, that doesn't really correspond to anything. That way we can get down to handling and displaying actual errors, and if the user disables antiforgery checks (e.g. if they are writing their own host runtime) then it shouldn't matter anyways.

Here's a legitimate, but just generated by my CLI and unused, cluster seed:
```
SCAMSVN4M2NZ65RWGYE42BZZ7VYEFEAAHGLIY7R4W7CRHORSMXTDJRKXLY
```

I'd be interested to know if @autodidaddict thinks differently about this problem

StuartHarris: Also interested in your opinion @autodidaddict. I don't think there's a nice solution. Ideally, since it wraps a String, the Seed's default should wrap String's default. But that won't parse, and if we went that way we'd be losing some of the safety that the type provides. I've made the change that @brooksmtownsend suggested, but we can revert that commit if you think there's a better way.

autodidaddict: I think `ClusterSeed::default()` should be the same as string default.. and then code elsewhere can/should validate empty/improperly formed seeds.

autodidaddict (approved):

brooksmtownsend (approved):

PR #204, opened 2021-11-21
feat(*): fetch from wasmcloud cache on inspect commands

### LINKED ISSUE #138: [FEATURE] Use wasmcloud cache for inspect commands
Currently, both `wash claims inspect` and `wash par inspect` do not refer to the wasmcloud cache before downloading an actor or provider from an OCI registry. This causes the `inspect` command to take longer than it should, as OCI references are intended to be immutable.

Following the same process as `wasmcloud-host` does, `wash claims inspect` and `wash par inspect` should read and write from/to the cache stored in the user's temp directory. This logic can be found in the [cached_file](https://github.com/wasmCloud/wasmCloud/blob/main/crates/wasmcloud-host/src/oci.rs#L68-L80) function.

matthewtgilbride:
Closes #138

### DIFF
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -104,6 +104,10 @@ pub(crate) struct InspectCommand {
     #[structopt(long = "insecure")]
     insecure: bool,
 
+    /// skip the local OCI cache
+    #[structopt(long = "no-cache")]
+    no_cache: bool,
+
     #[structopt(flatten)]
     pub(crate) output: Output,
 }
@@ -674,27 +678,19 @@ fn sign_file(cmd: SignCommand) -> Result<String, Box<dyn ::std::error::Error>> {
 async fn get_caps(
     cmd: &InspectCommand,
 ) -> Result<Option<Token<Actor>>, Box<dyn ::std::error::Error>> {
-    let module_bytes = match File::open(&cmd.module) {
-        Ok(mut f) => {
-            let mut buf = Vec::new();
-            f.read_to_end(&mut buf).unwrap();
-            buf
-        }
-        Err(_) => {
-            crate::reg::pull_artifact(
-                cmd.module.to_string(),
-                cmd.digest.clone(),
-                cmd.allow_latest,
-                cmd.user.clone(),
-                cmd.password.clone(),
-                cmd.insecure,
-            )
-            .await?
-        }
-    };
+    let artifact_bytes = crate::reg::get_artifact(
+        cmd.module.to_string(),
+        cmd.digest.clone(),
+        cmd.allow_latest,
+        cmd.user.clone(),
+        cmd.password.clone(),
+        cmd.insecure,
+        cmd.no_cache,
+    )
+    .await?;
 
     // Extract will return an error if it encounters an invalid hash in the claims
-    let claims = wascap::wasm::extract_claims(&module_bytes);
+    let claims = wascap::wasm::extract_claims(&artifact_bytes);
     match claims {
         Ok(token) => Ok(token),
         Err(e) => Err(Box::new(e)),
@@ -971,6 +967,7 @@ mod test {
             "--allow-latest",
             "--insecure",
             "--jwt-only",
+            "--no-cache",
         ])
         .unwrap();
 
@@ -984,6 +981,7 @@ mod test {
                 password,
                 insecure,
                 output,
+                no_cache,
             }) => {
                 assert_eq!(module, SUBSCRIBER_OCI);
                 assert_eq!(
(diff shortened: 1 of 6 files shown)


src/claims.rs
```diff
@@ -681,15 +681,21 @@ async fn get_caps(
...
-                cmd.digest.clone(),
-                cmd.allow_latest,
-                cmd.user.clone(),
-                cmd.password.clone(),
-                cmd.insecure,
-            )
-            .await?
+            if let Ok(mut from_cache) = File::open(cached_file(&cmd.module)) {
+                let mut buf = Vec::new();
+                from_cache.read_to_end(&mut buf).unwrap();
```
brooksmtownsend: Definitely a nit, but I think I'd prefer this variable to be called something like `cached_artifact` since `from` is a Rust keyword

src/par.rs
```diff
@@ -284,15 +284,22 @@ pub(crate) async fn handle_inspect(cmd: InspectCommand) -> Result<String> {
...
-                cmd.digest,
-                cmd.allow_latest,
-                cmd.user,
-                cmd.password,
-                cmd.insecure,
-            )
-            .await?;
+            let artifact = if let Ok(mut from_cache) = File::open(cached_file(&cmd.archive)) {
+                let mut buf = Vec::new();
+                from_cache.read_to_end(&mut buf).unwrap();
```
brooksmtownsend: Same `from` nit here

brooksmtownsend (commented):
Great stuff @matthewtgilbride, really only had one nit comment here.

matthewtgilbride:
> Great stuff @matthewtgilbride, really only had one nit comment here.

Thanks @brooksmtownsend - renamed things to "local_artifact" and "cached_artifact - along moving the local/cache/remote helper to `reg.rs`.  

I'm wondering what the best way to test this is?  I was thinking of jamming the `echo.wasm` file into the cache directory similarly to how it's added to the local directory [here](https://github.com/wasmCloud/wash/blob/a9f980c71a9f9e62ab53ab031ddf2a3a34889b53/tests/integration_claims.rs#L16).  Then I was going to add a call to get the remote actor before the `push` command  to see if it would get pulled from the cache.  Feels a bit awkward...thoughts?

brooksmtownsend:
> Thanks @brooksmtownsend - renamed things to "local_artifact" and "cached_artifact - along moving the local/cache/remote helper to `reg.rs`.
> 
> I'm wondering what the best way to test this is? I was thinking of jamming the `echo.wasm` file into the cache directory similarly to how it's added to the local directory [here](https://github.com/wasmCloud/wash/blob/a9f980c71a9f9e62ab53ab031ddf2a3a34889b53/tests/integration_claims.rs#L16). Then I was going to add a call to get the remote actor before the `push` command to see if it would get pulled from the cache. Feels a bit awkward...thoughts?

I think that's a good idea, a good baseline would be a unit test just to verify:
```
should produce a file like wasmcloud_azurecr_io_kvcounter_v1.bin
```

One great way you could test this is exactly what you suggested, throw the `echo.wasm` in the temp directory under a name like `fakeoci_io_echo_v1.bin` and then try to inspect `faceoci.io/echo:v1` and even though that doesn't exist, the cached file should be retrieved instead

brooksmtownsend (changes requested):
This LGTM @matthewtgilbride, just need the DCO fix 😄 If you check the workflow details it'll tell you the command you can run

matthewtgilbride:
👍 think I'm ready.  Added the `--no-cache` option so take a look before approving.

brooksmtownsend (changes requested):
I thought of one more request... could you also add a test for the provider so that we test that works as well? I figure having a `wash claims inspect` + `wash par inspect` would be great

matthewtgilbride:
> I thought of one more request... could you also add a test for the provider so that we test that works as well? I figure having a `wash claims inspect` + `wash par inspect` would be great

🦃 done

Interestingly the `wash par inspect` integration tests are _really_ slow.  Way slower than it should take to download/push the httpclient provider a couple of times.  🤔

brooksmtownsend:
> Interestingly the `wash par inspect` integration tests are _really_ slow.  Way slower than it should take to download/push the httpclient provider a couple of times.  🤔 

That's a fun one actually that we worried about as we started, turns out the whole compress/decompress part of the gzip library we use is SUPER slow in debug mode, so hopefully if we try with release mode that doesn't reproduce

brooksmtownsend (approved):
This LGTM, and should be ready to merge as soon as the tests run and pass 🚀

PR #190, opened 2021-10-04
GitHub Action update + Brew automation

jordan-rash:
Update action to build OS specific artifacts while supporting automated Brew releases for OSX installation

Requires: https://github.com/wasmCloud/homebrew-wasmcloud/pull/6

PR #205, opened 2021-11-21
Draft: Update Github Actions

### LINKED ISSUE #189: Scrub GHA actions
I have noticed that the actions are using deprecated or out dated `uses:` stanzas.  Will use this issue to note them as I find them all

`release.yml`
https://github.com/wasmCloud/wash/blob/main/.github/workflows/release.yml#L23
https://github.com/wasmCloud/wash/blob/main/.github/workflows/release.yml#L36
https://github.com/wasmCloud/wash/blob/main/.github/workflows/release.yml#L50

ghost:
Fixes #189.

Updated all occurrences of `actions/checkout` to use the v2 version.

Changed the `release` job to use `softprops/action-gh-release@v1`, as used in https://github.com/wasmCloud/wasmcloud-otp.
Wasmcloud-otp creates tarballs for each platform and puts them in the Github Release though, and I wasn't sure if that was something that should be done here. I'm leaning towards no because `wash` is already distributed through the various package managers, but interested to hear any feedback on that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,24 +3,24 @@ name: Release
 on:
   push:
     tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
 
 jobs:
   cargo_check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: make test-all
-    - name: Check fmt
-      run: cargo fmt -- --check
+      - uses: actions/checkout@v2
+      - name: Build
+        run: cargo build --verbose
+      - name: Run tests
+        run: make test-all
+      - name: Check fmt
+        run: cargo fmt -- --check
 
   clippy_check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - run: rustup component add clippy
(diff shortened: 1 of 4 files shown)


ghost:
Ah didn't see #190 before opening this. That one doesn't appear to update the checkout actions though.

autodidaddict:
@brooksmtownsend @jordan-rash we should rationalize between this Pr and #190 .

jordan-rash:
Agree.  I think, at this point, I will close #190 and we move forward with this one.  The primary focus on #190 was to get brew automation working and not the critical updates that @mattwilkinsonn did here.  

Also, in my getting more familiar with brew, it turns out, deep in the bowels of their documentation, there is a better way than what I proposed in #190.  So I will leave the branch there, but we can get rid of the PR for now.

jordan-rash:
One thing that sticks out here is the linting.  We should probably have a common action that does a linting check on the yaml for all our GHAs.  

Screen Shot 2021-11-22 at 7 47 14 AM

ghost:
> One thing that sticks out here is the linting. We should probably have a common action that does a linting check on the yaml for all our GHAs.
> 
> Screen Shot 2021-11-22 at 7 47 14 AM

Yeah I noticed that as well. The YAML formatter I have on VSCode auto fixed it, but could be useful to have it lint in the CI itself.

Some google searching led me here https://github.com/marketplace/actions/yaml-schema-validator and https://github.com/marketplace/actions/yaml-lint

brooksmtownsend (approved):
This looks great! I agree with you that there's no need to compile+attach the `wash` binary since we already have other mechanisms for distribution. As long as the actions pass this LGTM

PR #207, opened 2021-11-23
ref(link): Return better errors when provider id is given

thomastaylor312:
It is relatively common for first time users to pass the provider id rather
than the contract ID when doing `wash ctl link del`. This adds a block
that tests if it is a provider ID and returns a hint to the user. I also
noticed we weren't returning an error code on link commands if they failed,
so I added some logic to fix it

### DIFF
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -463,6 +463,20 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
                 .link_name
                 .clone()
                 .unwrap_or_else(|| "default".to_string());
+            // Check if the contract ID parameter is a 56 character key and suggest that the user
+            // give the contract ID instead
+            //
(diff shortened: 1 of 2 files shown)


src/ctl/mod.rs
```diff
@@ -463,6 +463,20 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
                 .link_name
                 .clone()
                 .unwrap_or_else(|| "default".to_string());
+            // Check if the contract ID parameter is a 52 character key and suggest that the user
```
autodidaddict: nit: you say 52 in the comment but it's 56 for realz

autodidaddict (dismissed):

src/ctl/mod.rs
thomastaylor312: Ha, I updated it in one place and not the other

brooksmtownsend (approved):
LGTM, this is gonna save a few developers a lot of time

PR #214, opened 2021-12-08
feat(ctl): warning if contract IDs look like an nkey

### LINKED ISSUE #188: [FEATURE] validation of `wash link put` arguments
it's really easy to mistype a contract id, or get the arguments to `wash ctl link put` mixed up, and put an id in there instead.

It would be nice for wash to do some basic validation 
- [ ] force contract id to contain a `:`
- [ ] any contract id that starts `wasmcloud:` should match against a list of well-known first-party contract ids
  - [ ] maybe a flag to skip this validation if you're deploying a new capability provider on an old version of `wash`?

An alternative would be to avoid using positional arguments entirely, and force people to use option flags instead (this is the approach that the azure CLI takes, if my memory serves correctly). This would also avoid the problem that the `wash reg push` arguments are in a confusing order.

matthewtgilbride:
closes #188

### DIFF
--- a/src/call.rs
+++ b/src/call.rs
@@ -1,6 +1,7 @@
 use crate::{
     ctx::{context_dir, get_default_context, load_context},
     id::ClusterSeed,
+    id::ModuleId,
     util::{
         convert_rpc_error, extract_arg_value, format_output, json_str_to_msgpack_bytes,
         msgpack_to_json_val, nats_client_from_opts, Output, OutputKind, Result,
@@ -121,7 +122,7 @@ pub(crate) struct CallCommand {
 
     /// Public key or OCI reference of actor
     #[structopt(name = "actor-id")]
-    pub(crate) actor_id: String,
+    pub(crate) actor_id: ModuleId,
 
     /// Operation to invoke on actor
     #[structopt(name = "operation")]
@@ -322,8 +323,10 @@ async fn rpc_client_from_opts(
 #[cfg(test)]
 mod test {
     use super::{CallCli, CallCommand};
+    use crate::id::ModuleId;
     use crate::util::Result;
     use std::path::PathBuf;
+    use std::str::FromStr;
     use structopt::StructOpt;
 
     const RPC_HOST: &str = "127.0.0.1";
@@ -395,7 +398,7 @@ mod test {
                 );
                 assert!(test);
                 assert_eq!(bin, '2');
-                assert_eq!(actor_id, ACTOR_ID);
+                assert_eq!(actor_id, ModuleId::from_str(ACTOR_ID).unwrap());
                 assert_eq!(operation, "HandleOperation");
                 assert_eq!(payload, vec!["{ \"hello\": \"world\"}".to_string()])
             }
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -1,23 +1,27 @@
 extern crate wasmcloud_control_interface;
+
+use std::{
+    path::{Path, PathBuf},
+    time::Duration,
+};
+
+use spinners::{Spinner, Spinners};
+use structopt::StructOpt;
+use wasmcloud_control_interface::{
+    Client as CtlClient, CtlOperationAck, GetClaimsResponse, Host, HostInventory,
+    LinkDefinitionList,
+};
+
+pub(crate) use output::*;
+
 use crate::{
     ctl::manifest::HostManifest,
     ctx::{context_dir, get_default_context, load_context},
     id::{ModuleId, ServerId, ServiceId},
     util::{
-        convert_error, labels_vec_to_hashmap, Output, OutputKind, Result, DEFAULT_LATTICE_PREFIX,
-        DEFAULT_NATS_HOST, DEFAULT_NATS_PORT, DEFAULT_NATS_TIMEOUT,
+        convert_error, labels_vec_to_hashmap, validate_contract_id, Output, OutputKind, Result,
+        DEFAULT_LATTICE_PREFIX, DEFAULT_NATS_HOST, DEFAULT_NATS_PORT, DEFAULT_NATS_TIMEOUT,
     },
-};
-pub(crate) use output::*;
-use spinners::{Spinner, Spinners};
-use std::{
-    path::{Path, PathBuf},
-    time::Duration,
-};
-use structopt::StructOpt;
-use wasmcloud_control_interface::{
-    Client as CtlClient, CtlOperationAck, GetClaimsResponse, Host, HostInventory,
-    LinkDefinitionList,
 };
 
 mod manifest;
@@ -479,20 +483,9 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
                 .link_name
                 .clone()
                 .unwrap_or_else(|| "default".to_string());
-            // Check if the contract ID parameter is a 56 character key and suggest that the user
-            // give the contract ID instead
-            //
-            // NOTE: `len` is ok here because keys are only ascii characters that take up a single
-            // byte.
-            if cmd.contract_id.len() == 56 && cmd.contract_id.starts_with('V') {
-                link_del_output(
-                    &cmd.actor_id,
-                    &cmd.contract_id,
-                    link_name,
-                    Some("It looks like you used a Provider ID (e.g. VABC...) instead of a contract ID (e.g. wasmcloud:httpserver)".into()),
-                    &cmd.output.kind,
-                )?;
-            }
+
+            validate_contract_id(&cmd.contract_id)?;
+
             sp = update_spinner_message(
                 sp,
                 format!(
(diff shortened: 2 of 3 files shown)


src/ctl/mod.rs
```diff
@@ -652,6 +657,13 @@ pub(crate) async fn get_claims(cmd: GetClaimsCommand) -> Result<GetClaimsRespons
 }
 
 pub(crate) async fn link_del(cmd: LinkDelCommand) -> Result<CtlOperationAck> {
+    if let (Ok(_), Ok(_), Ok(_)) = (
+        ModuleId::from_str(&cmd.contract_id),
+        ServiceId::from_str(&cmd.contract_id),
+        ServerId::from_str(&cmd.contract_id),
+    ) {
+        log::warn!("{} has been passed as a contract ID but it looks like an Actor/Provider key.  You sure about that my friend?", cmd.contract_id)
+    }
```
brooksmtownsend: I like the idea of this, though I think that this will never actually work because one string couldn't be `Ok(_)` parsed to be a module ID, service ID, or serverId

You could restructure this just a bit though, you could do a simple `if ModuleId::from_str().is_ok() || ServiceId::from_str().is_ok() || ...` in order to just check if any of them parse correctly and that would work. Though honestly, I think a fine amount of rigor here is checking if the contract ID is a 56 character alphanumeric uppercase string. Contract IDs are a convention to be something like `wasmcloud:httpserver`, and I think it's highly unlikely anyone seriously uses a contract ID of `THISISASQLDBPROVIDERANDITSTHEFIRSTVERSIONOFMANYTOCOMEHAH` as a contract ID.

TL;DR, I think you'd be fine just asserting that it's not a 56 length string with all uppercase letters and numbers.

brooksmtownsend (changes requested):

src/ctl/mod.rs
matthewtgilbride: Agreed.  Take a look now...sorry for the delay 👍

matthewtgilbride: Added a validation of the `ctl stop` command as well...two questions:

1. Am I handling sending output to the user how we want to?  The `link_del_output` and `link_put_output` commands format the json message under an `error` key...maybe it should be `warning`?
2. Are there any other validations we wanted to handle as part of #192?

src/ctl/mod.rs
```diff
@@ -1401,3 +1420,19 @@ mod test {
...
+// Check if the contract ID parameter is a 56 character key and suggest that the user
+// give the contract ID instead
+//
+// NOTE: `len` is ok here because keys are only ascii characters that take up a single
+// byte.
+fn validate_contract_id(contract_id: &str) -> Option<&str> {
+    if contract_id.len() == 56
+        && contract_id
+            .chars()
+            .all(|c| char::is_alphanumeric(c) && char::is_uppercase(c))
```
brooksmtownsend: I think here we probably _technically_ want [is_ascii_alphanumeric](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_alphanumeric), though I'm not sure if we'd ever hit that edge case.

src/ctl/mod.rs
```diff
@@ -1401,3 +1420,19 @@ mod test {
         Ok(())
     }
 }
+
+// Check if the contract ID parameter is a 56 character key and suggest that the user
+// give the contract ID instead
+//
+// NOTE: `len` is ok here because keys are only ascii characters that take up a single
+// byte.
+fn validate_contract_id(contract_id: &str) -> Option<&str> {
```
brooksmtownsend: I'd suggest making the return type of this function a `Result<()>`, using the `crate::util::Result`. That way, you can either return `Ok(())` if it's a valid contract ID, and then `Err("It looks like you ...")` if not.

Doing this lets you do simple things like:
```
validate_contract_id("wasmcloud:httpserver")?;
// do a link put
```

which will then bubble up the error if it exists, or otherwise validate the contract ID and continue execution

src/ctl/mod.rs
```diff
@@ -796,6 +808,13 @@ pub(crate) async fn start_provider(cmd: StartProviderCommand) -> Result<CtlOpera
 }
 
 pub(crate) async fn stop_provider(cmd: StopProviderCommand) -> Result<CtlOperationAck> {
+    validate_contract_id(&cmd.contract_id).map(|message| {
+        format_output(
+            message.to_string(),
+            json!({ "error": message }),
+            &cmd.output.kind
+        )
+    });
```
brooksmtownsend: See suggestion about `validate_contract_id` returning `Result<()>`, that way you can just call the function and use `?` to bubble up the error. You don't actually need the `format_output` thing here, we don't currently consistently return json output for errors and should make another issue for that

src/ctl/mod.rs
```diff
@@ -513,6 +516,15 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
...
         }
         Link(LinkCommand::Put(cmd)) => {
+            validate_contract_id(&cmd.contract_id).map(|message| {
+                link_put_output(
+                    &cmd.actor_id,
+                    &cmd.provider_id,
+                    Some(message.to_string()),
+                    &cmd.output.kind,
+                )
+            });
```
brooksmtownsend: See suggestion about validate_contract_id returning Result<()>, that way you can just call the function and use ? to bubble up the error. No need for `link_put_output` here as it only returns a formatted string

src/ctl/mod.rs
```diff
@@ -479,20 +485,17 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
...
                 .clone()
                 .unwrap_or_else(|| "default".to_string());
-            // Check if the contract ID parameter is a 56 character key and suggest that the user
-            // give the contract ID instead
-            //
-            // NOTE: `len` is ok here because keys are only ascii characters that take up a single
-            // byte.
-            if cmd.contract_id.len() == 56 && cmd.contract_id.starts_with('V') {
+
+            validate_contract_id(&cmd.contract_id).map(|message| {
```
brooksmtownsend: Same comments as above, this can just be a `validate_contract_id` call without the `link_del_output` call

brooksmtownsend (changes requested):

src/ctl/mod.rs
brooksmtownsend: https://github.com/wasmCloud/wash/issues/217 (other issue that I just wrote up)

src/ctl/mod.rs
brooksmtownsend: 1. I think I addressed this in the comments, but I don't believe you need to use the `link___output` functions for your error formatting and a `?` will suffice just fine
2. I think that the only other change I'd want to see for #192 is validating the actor ID for `wash call`

src/ctl/mod.rs
stevelr: is_uppercase('1') would return false
so you probably need something like
```contract_id.chars().all(|c| c.is_ascii_digit() || c.is_ascii_uppercase());```

**Comment by matthewtgilbride** [comment]
@brooksmtownsend apologies for the holiday delay 🌲 - have a look when you have another chance.

File: src/call.rs [hunk 7]
```diff
@@ -121,7 +122,7 @@ pub(crate) struct CallCommand {
 
     /// Public key or OCI reference of actor
     #[structopt(name = "actor-id")]
-    pub(crate) actor_id: String,
+    pub(crate) actor_id: ModuleId,
```
matthewtgilbride, line 125: Is this all that's necessary?  I'm presuming the `structopt` crate will throw an error if the argument can't be parsed to a `ModuleId`? Clever indeed...

autodidaddict: Yeah this works because of trait magic. Since `ModuleId` implements `from_str`, structopt automatically knows how to parse it from the command line: https://github.com/wasmCloud/wash/blob/cf6126436301f943a58a6dbc778014c093988307/src/id.rs#L22

autodidaddict:
Current build failure (test fail):
```
    --> src/call.rs:399:17
    |
399 |                 assert_eq!(actor_id, ACTOR_ID);
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `Id`, found `&str`
    |
    = note: expected struct `Id<'M'>`
            found reference `&str`
    = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

matthewtgilbride:
> Current build failure (test fail):
> 
> ```
>     --> src/call.rs:399:17
>     |
> 399 |                 assert_eq!(actor_id, ACTOR_ID);
>     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `Id`, found `&str`
>     |
>     = note: expected struct `Id<'M'>`
>             found reference `&str`
>     = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
> ```

fixed

autodidaddict (approved):

brooksmtownsend (approved):

brooksmtownsend:
Thanks @matthewtgilbride !

PR #216, opened 2021-12-12
fix/feat: add `cfg` module to get or create the `.wash` directory

### LINKED ISSUE #215: [BUG] `.wash` folder not created, context commands error out
**Describe the bug**

The `.wash` folder was not created on my computer, and when I try to use the `wash ctx` commands they fail on a missing directory.

**To Reproduce**
1. Install wash from `cargo`
2. Attempt to follow wasmCloud [Getting Started](https://wasmcloud.dev/overview/getting-started/) guide, using Docker Compose
3. Try to create a context using `wash ctx`.
4. Error `Error: No such file or directory (os error 2)`

**Expected behavior**
The `.wash` folder should be created, either when the tool is installed, or all commands that use it should have the ability to create the folder on the fly.

**Screenshots**

Screen Shot 2021-12-12 at 3 01 56 PM

 - OS: macOS Big Sur 11.6.1
 - Shell: zsh
 - wash Version: 0.7.2

ghost:
Fixes #215.

Currently if the `.wash` folder does not already exist, the `ctx` commands fail. This PR adds a `cfg` module along with a `cfg_dir` function that gets the `.wash` directory, and creates it if it does not exist.

This will need to be added to all of the functions that use the `.wash` directory. That looks like `keys.rs` and a few other ones in `ctx`. I'll implement these after getting a review on this design and any feedback you guys have.


src/ctx/mod.rs
```diff
@@ -406,11 +407,12 @@ pub(crate) fn context_dir(cmd_dir: Option<PathBuf>) -> Result<PathBuf> {
     let dir = if let Some(dir) = cmd_dir {
         dir
     } else {
-        home_dir()?.join(CTX_DIR)
```
brooksmtownsend, line 409: Only request: now that `cfg_dir` exists, let's go ahead and remove the `home_dir` function and all references to it to avoid any unexpected behavior in the future

brooksmtownsend (changes requested):

ghost:
Follow up question @brooksmtownsend  - Why do we have separate environment variables for each part of the config directory? Like `WASH_CONTEXTS` etc. Could we just have a `WASH_CONFIG` variable so they can customize where the config is stored, but not the layout inside the config folder?

My thinking is that if we then add more configuration options, users would have to maintain tons of environment variables if they wanted to customize where the config is stored, vs just setting a single one and calling it a day.

brooksmtownsend:
> Why do we have separate environment variables for each part of the config directory? Like WASH_CONTEXTS etc. Could we just have a WASH_CONFIG variable so they can customize where the config is stored, but not the layout inside the config folder?

The main reason here was that someone could feasibly set `WASH_KEYS=/tmp/mykeys` where they mounted a folder containing keys, and it wouldn't have to be called `/path/to/folder/keys`. 

I think there's three possible scenarios for someone configuring wash
1. User is fine with just using `$HOME/.wash`, no configuration needed (most common)
2. User wants to use a custom directory for wash things, e.g. `$HOME/github/wash_cfg/`, in which case a `WASH_CONFIG` or `WASH_CFG_DIR` environment variable is best
3. User wants to individually use a custom directory, e.g. a filesystem mounted folder, for `keys` or `contexts` like `/home/washctxs`

We currently support all three options with `WASH_KEYS` and `WASH_CONTEXTS`, and if we used a `WASH_CFG_DIR` variable we couldn't support option 3. For now I lean towards keeping _KEYS and _CONTEXTS even though it's a little more verbose just because it's more configurable, but I'm certainly be open to more discussion

ghost:
> We currently support all three options with `WASH_KEYS` and `WASH_CONTEXTS`, and if we used a `WASH_CFG_DIR` variable we couldn't support option 3. For now I lean towards keeping _KEYS and _CONTEXTS even though it's a little more verbose just because it's more configurable, but I'm certainly be open to more discussion

Ok that makes sense! Perhaps could revisit this if it becomes an issue later.

ghost:
I've removed the `home_dir` function and made all references to `.wash` use the `cfg_dir` one. This PR should be ready now @brooksmtownsend

brooksmtownsend (approved):
LGTM, thank you again @mattwilkinsonn !

PR #220, opened 2021-12-16
fix: replace `.unwrap()` with `?` in image reference parsing

### LINKED ISSUE #219: [BUG] crash when using http:// before registry address
**Describe the bug**

**To Reproduce**
1. Start a local private registry
2. Push an actor or provider to the local registry
3. Use wash to ping the registry using http or https before the address

Eg.
```
wash reg ping --insecure http://localhost:5000/v2/fakepay_provider:0.1.0
```

**Expected behaviour**
Ok

**Actual behaviour**

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ReferenceInvalidFormat', /home/mats/.cargo/registry/src/github.com-1ecc6299db9ec823/wash-cli-0.7.2/src/reg.rs:276:44
```

 - OS: Linux/Ubuntu 20.04
 - Shell: bash
 - wash Version 0.7.2

ghost:
Fixes #219.

Using `http://` in the reference should crash as that isn't supported by OCI, as far as I can tell. But bubbling up the error results in a cleaner error message instead of a panic trace.

Old error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ReferenceInvalidFormat', src/reg.rs:276:44
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

New error:
```
Error: invalid reference format
```

### DIFF
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -149,7 +149,7 @@ pub(crate) async fn handle_command(
 }
 
 pub(crate) async fn handle_pull(cmd: PullCommand) -> Result<String, Box<dyn ::std::error::Error>> {
-    let image: Reference = cmd.url.parse().unwrap();
+    let image: Reference = cmd.url.parse()?;
     let spinner = match cmd.output.kind {
         OutputKind::Text => Some(Spinner::new(
             &Spinners::Dots12,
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
It's the simple fixes that are really satisfying 😉

PR #221, opened 2021-12-16
Refactor output messages to better support JSON

### LINKED ISSUE #217: [FEATURE] Return properly formatted JSON on error
`wash` strives to support an `-o json` option wherever possible for command output to best support scripting. This is usually handled well for success output, but errors can often be unwrapped with `?` and simply printed to the console in the `main` module.

Ideally, we should have a methodology to output success messages and failures. The solution I'd like is that if you use the `-o json` flag with `wash`, every possible command output should be able to be piped into `jq` and interpreted. For example, a failure to connect to NATS in text mode and in JSON mode (currently) outputs:
```
Error: Connection refused (os error 61)
```
But in JSON mode, it should be:
```
{"error": "Connection refused (os error 61)"}
```

Scripting with `wash` shouldn't fail with the inability for `jq` to parse the errors.

My proposal is to ensure we properly format successful commands with JSON output, which should already be done. For error messages, they should simply be returned as Err(string) to the `main` module and turned into the proper output format there. This is for simplicity as there are many possible sources of errors, primarily stemming from using `?` to unwrap or explicitly returning errors.

We could simply change every `?` into a `map_err` call that uses `format_output` to properly format the error as a JSON string when appropriate. This will require a good bit of code modification and has a greater potential to be forgotten, though it is a viable option.

ghost:
Fixes #217.

Right now the `-o json` support is spotty across the different commands, and is nonexistent for errors. I'm hoping to solve this with this PR.

My idea is to have a new type, `CommandOutput`, that is returned from every command. It includes a JSON map and a string, and one of those is selected to be output in `main.rs`. This forces every command to offer a JSON output.

For errors with JSON output, I'm just transforming the error into a json map with the error message under the `error` key, so we can still just bubble up errors to the top and then print them.

I've refactored `call` to use this new type as an example. I'll refactor the others after hearing what you all think.


src/main.rs
```diff
@@ -40,6 +44,9 @@ A single CLI to handle all of your wasmCloud tooling needs
             name = "wash",
             about = ASCII)]
 struct Cli {
+    #[structopt(flatten)]
+    output: Output,
+
```
brooksmtownsend: When we do this here, does this affect the positioning of the arguments at all? As long as the following command works, for example, then I'm fine with this (and pulling the `Output` struct out of all the commands)
```
wash call -o json MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 HttpServer.HandleRequest '{"method": "GET", "path": "/echo", "body": "", "queryString":"","header":{}}'
```
and
```
```
wash call -o json MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5 HttpServer.HandleRequest '{"method": "GET", "path": "/echo", "body": "", "queryString":"","header":{}}' -o json
```
```

brooksmtownsend (commented):
I like this refactoring a _lot_. This is exactly the kind of solution I was envisioning where we don't need to worry about `?` unwrapping across the project and it's easy to remember to provide both formats with the `Result<CommandOutput>` return type.

I had one question that I'd like to know the answer to (in this review) but other than that I'd put my ✅ of approval on this

autodidaddict:
Per title (and breaking build status), converting this to a draft for now.

ghost:
@brooksmtownsend - is there a reason we don't use [anyhow](https://docs.rs/anyhow/latest/anyhow/) instead of rolling our own Result type?

brooksmtownsend:
@mattwilkinsonn There's not a reason why we don't use `anyhow`, that's something that would be useful it just wasn't really necessary when we were writing `wash`. I think it would make error logic cleaner, so I'd happily accept some refactoring to use `anyhow` 😄

thomastaylor312:
Just a heads up that `anyhow` sometimes causes some error swallowing when returning as an error in a CLI application. I love `anyhow`, but it _could_ be a little gnarly. For example, this is what we had to do in Bindle: https://github.com/deislabs/bindle/blob/main/bin/client/main.rs#L38-L46

autodidaddict:
The original idea behind not using anyhow was just keeping things simple and not taking any crate dependencies we didn't absolutely need.

ghost:
I finished changing everything to `anyhow` and `CommandOutput`, and got all of the tests passing. This should now be ready for review.

src/main.rs
ghost: Using the `global = true` flag in the structopt macro lets the `-o` flag not care about what position it is in, so we're all good here!

autodidaddict: TIL that structopt has a global option 🤯

src/call.rs
```diff
@@ -231,7 +224,7 @@ async fn rpc_client_from_opts(
     let ctx = if let Some(context) = opts.context {
         load_context(&context).ok()
     } else if let Ok(ctx_dir) = context_dir(None) {
-        get_default_context(&ctx_dir).ok()
+        get_default_context(ctx_dir.as_path()).ok()
```
brooksmtownsend, line 228: nit: I don't particularly mind if it's `&` or `.as_path()` but let's change `225` to do the same thing or revert this part

src/ctl/mod.rs
```diff
@@ -519,104 +468,117 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
...
-                &format!("Provider {} started successfully", provider_ref),
-                &ack.error,
-                &output.kind,
+            if !ack.accepted {
+                bail!("Operation failed: {}", ack.error);
+            }
+
+            CommandOutput::from_key_and_text(
+                "result",
+                format!("Provider {} started successfully", provider_ref),
```
brooksmtownsend: This is actually a change that went in later, but when you rebase you'll see the success message changed here. This should say something about how the provider start request was received

src/ctl/mod.rs
```diff
@@ -635,7 +597,9 @@ pub(crate) async fn get_hosts(cmd: GetHostsCommand) -> Result<Vec<Host>> {
         (None, None) => Duration::from_millis(DEFAULT_NATS_TIMEOUT),
     };
     let client = ctl_client_from_opts(cmd.opts).await?;
-    client.get_hosts(timeout).await.map_err(convert_error)
+    Ok(client.get_hosts(timeout).await.map_err(convert_error)?)
+    // .map_err(convert_error)
+    // .into()
```
brooksmtownsend: We can remove these 2 lines of comment

src/ctl/mod.rs
```diff
@@ -519,104 +468,117 @@ pub(crate) async fn handle_command(command: CtlCliCommand) -> Result<String> {
...
-                &format!("Actor {} started successfully", actor_ref),
-                &ack.error,
-                &output.kind,
+            if !ack.accepted {
+                bail!("Operation failed: {}", ack.error);
+            }
+
+            CommandOutput::from_key_and_text(
+                "result",
+                format!("Actor {} started successfully", actor_ref),
```
brooksmtownsend: Same comment re: "request received"

src/ctl/output.rs
```diff
@@ -1,116 +1,95 @@
...
+            map.insert("link_name".to_string(), json!(link_name));
+            Ok(CommandOutput::new(
+                format!(
+                    "Deleted link for {} on {} ({}) successfully",
+                    actor_id, contract_id, link_name
+                ),
+                map,
+            ))
+        }
+        Some(f) => Err(anyhow!(format!("Error deleting link: {}", f),)),
```
brooksmtownsend: I think here we can insert the failure into the command output as key `failure` and just omit `Error deleting link`

src/ctl/output.rs
```diff
@@ -1,116 +1,95 @@
...
+            map.insert("provider_id".to_string(), json!(provider_id));
+            Ok(CommandOutput::new(
+                format!(
+                    "Published link ({}) <-> ({}) successfully",
+                    actor_id, provider_id
+                ),
+                map,
+            ))
+        }
+        Some(f) => bail!("Error advertising link: {}", f),
```
brooksmtownsend, line 73: Here we can use the command output map and just use `f` as the failure value

src/ctx/mod.rs
```diff
@@ -281,11 +271,14 @@ fn handle_edit(cmd: EditCommand) -> Result<String> {
...
         }
-        Command::new(editor)
+        let status = Command::new(editor)
             .arg(&context_path_from_name(&dir, &ctx_name))
-            .status()
-            .map(|_exit_status| "Finished editing context successfully".to_string())
-            .map_err(|e| e.into())
+            .status()?;
+
+        match status.success() {
```
brooksmtownsend, line 278: nit/personal preference: since we don't use status anymore we can just match against the `Command::new` statement

src/reg.rs
```diff
@@ -287,14 +285,14 @@ pub(crate) async fn handle_ping(cmd: PingCommand) -> Result<String, Box<dyn ::st
         _ => RegistryAuth::Anonymous,
     };
     let (_, _) = client.pull_manifest(&image, &auth).await?;
-    Ok("Ok".to_string())
+    Ok(CommandOutput::from("Ok"))
```
brooksmtownsend: Ping should "Pong" back, right? 😄

src/smithy.rs
```diff
@@ -128,15 +129,14 @@ pub(crate) async fn handle_lint_command(
     )
     .map_err(|e| anyhow!("lint error: {}", e.to_string()))?;
 
+    // TODO: make this return the report instead of printing directly
```
brooksmtownsend, line 132: @stevelr does `report_action_issues` print directly to stdout or is there an alternative function to return a string?

brooksmtownsend (changes requested):
A couple of comments and suggestions, and you will need to rebase on main...

Overall, this is fantastic! There's a ton of code cleanup here, increased clarity, and overall a much better experience. Really love the addition of adding json output for errors everywhere, as we can now easily script wash commands with error handling in mind. Love when a new feature removes more code than it adds 😄

src/ctl/output.rs
ghost: The only problem I see with this is that we're then actually returning an `Ok` when it really should be an `Err`.

 I'm not really sure why I wrote it so ugly though. I've changed it to just use the `bail!` macro. Does that work?

src/ctl/output.rs
ghost: Same comment as above - I have reservations about using `Ok` for something that is actually an `Err`.

src/ctx/mod.rs
ghost: My thinking was to use the `?` operator there so I can avoid another level of nesting just to match the `Result` - I can match on the `Result` instead if you'd prefer.

src/reg.rs
ghost: Lol, changed

src/call.rs
ghost: Changed above statement to match. I personally find the `&` confusing when it's can coerce something into a different type (`PathBuf` -> `Path`). I know that they are kind of the same, but definitely threw me for a loop at first glance.

src/smithy.rs
mattwilkinsonn: > A couple of comments and suggestions, and you will need to rebase on main...
> 
> Overall, this is fantastic! There's a ton of code cleanup here, increased clarity, and overall a much better experience. Really love the addition of adding json output for errors everywhere, as we can now easily script wash commands with error handling in mind. Love when a new feature removes more code than it adds 😄

Thanks! I enjoyed getting to clean it up a bit. Also helped me learn the internals of what exactly everything is doing.

And yeah, seeing the delete count higher than the add count was definitely my favorite part of making this PR, lol.

I've made most of the changes you've requested @brooksmtownsend, and left comments on the ones I have questions about. I also rebased on main to pick up the other changes. I'll look out for your feedback on those remaining items!

ghost:
> A couple of comments and suggestions, and you will need to rebase on main...
> 
> Overall, this is fantastic! There's a ton of code cleanup here, increased clarity, and overall a much better experience. Really love the addition of adding json output for errors everywhere, as we can now easily script wash commands with error handling in mind. Love when a new feature removes more code than it adds 😄

Thanks! I enjoyed getting to clean it up a bit. Also helped me learn the internals of what exactly everything is doing.

And yeah, seeing the delete count higher than the add count was definitely my favorite part of making this PR, lol.

I've made most of the changes you've requested @brooksmtownsend, and left comments on the ones I have questions about. I also rebased on main to pick up the other changes. I'll look out for your feedback on those remaining items!

src/smithy.rs
ghost: `report_action_issues` definitely just prints to stdout directly ATM. I just left it as a todo as I suspect this will need to be handled in the upstream crate first

ghost:
The integration tests hate the JSON output for some reason. I'm working on refactoring the tests to be more reliable.

ghost:
I've refactored all of the integration tests to parse the outputted JSON, and then use [assert-json-diff](https://crates.io/crates/assert-json-diff) to check it, so they more reliably test the JSON coming out of each command. They are all passing locally for me now.

src/ctx/mod.rs
brooksmtownsend: It's only a personal preference so I'm fine with it, it is a few levels of method chaining and this reads well

src/ctl/output.rs
brooksmtownsend: Ah I might not have been specific enough, I do like this being an `Err` here, but I was mostly just referring to changing the error message to just be `f` instead of `Error deleting link: {f}`. After reflecting a bit further I think this is fine, your `bail!` change is just fine

brooksmtownsend (approved):
This all LGTM after the latest commits, just looking for a secondary approval from @autodidaddict / @thomastaylor312 / @stevelr

autodidaddict (approved):

brooksmtownsend:
Thanks @mattwilkinsonn !

Issue #218, opened 2021-12-13
[BUG] Auto key generation writes keys to disk as world readable


thomastaylor312:
**Describe the bug**
When using `wash key gen`, the keys written to disk have permissions of `644`, which means that anyone on the system can read them.

**To Reproduce**
1. Run a `wash claims sign` command to sign an actor or provider
1. Run `ls -la ~/.wash/keys` 
1. Observe the global read permission for the directory and generated key

**Expected behavior**
When generating keys, the permissions on *nix-like systems should be set to `600`. Also, when creating the directory for keys, that directory permission should be set to `700`.

**Environment (please complete the following information) **
 - OS: MacOS
 - Shell: zsh
 - wash Version: 0.7.1

thomastaylor312:
Please note that for windows, it is real hard to manage permissions like this is Rust. The permissions logic will need to be separate for the different OS families anyway

protochron:
I started working on this, and I'm done with the *nix side. Windows of course is a lot more difficult, since it looks like you need to call the Windows API directly using something like https://crates.io/crates/winapi.

thomastaylor312:
Yeah, that sounds right. And even that API is a little unclear on what is the proper way to set permissions

PR #225, opened 2022-01-06
Set permissions on generated keys to be user-only on Unix systems

protochron:
Set the permissions on generated keys to `0600` and the directory to
`0700` on Unix systems. Windows permissions are a no-op due to having to
call the Windows APIs directly which is not very straightforward in
Rust.

Partial fix for #218.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -1,6 +1,6 @@
 use crate::{
     cfg::cfg_dir,
-    util::{CommandOutput, OutputKind},
+    util::{set_permissions_keys, CommandOutput, OutputKind},
 };
 use anyhow::{bail, Context, Result};
 use nkeys::{KeyPair, KeyPairType};
@@ -206,9 +206,12 @@ pub(crate) fn extract_keypair(
 
                 let kp = KeyPair::new(keygen_type);
                 let seed = kp.seed()?;
-                fs::create_dir_all(Path::new(&path).parent().unwrap())?;
-                let mut f = File::create(path)?;
+                let key_path = Path::new(&path).parent().unwrap();
+                fs::create_dir_all(key_path)?;
+                set_permissions_keys(key_path)?;
+                let mut f = File::create(path.clone())?;
                 f.write_all(seed.as_bytes())?;
+                set_permissions_keys(&path)?;
                 seed
             }
             _ => {
--- a/src/util.rs
+++ b/src/util.rs
@@ -2,7 +2,13 @@ use anyhow::{anyhow, bail, Result};
 use nats::asynk::Connection;
 use serde::{Deserialize, Serialize};
 use std::{
-    collections::HashMap, env::temp_dir, error::Error, fmt, fs::File, io::Read, path::PathBuf,
+    collections::HashMap,
+    env::temp_dir,
+    error::Error,
+    fmt, fs,
+    fs::File,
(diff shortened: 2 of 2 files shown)


protochron:
I can leave this as is and add in Windows as a follow-up PR or leave this up until that work is finished. Let me know what you'd prefer!

brooksmtownsend (dismissed):
```
~/.wash ➜ ll
Permissions Size User   Date Modified Name
drwxr-xr-x     - brooks 29 Nov  2021  contexts/
drwx------     - brooks  6 Jan 09:36  keys/
drwxr-xr-x     - brooks  6 Jan 09:29  prekeys/
.rw-r--r--   634 brooks  5 Jan 13:00  host_config.json

~/.wash ➜ ll prekeys
Permissions Size User   Date Modified Name
.rw-r--r--    58 brooks 29 Nov  2021  brooks_account.nk
.rw-r--r--    58 brooks 30 Nov  2021  clinicapi_module.nk
...

~/.wash ➜ ll
Permissions Size User   Date Modified Name
drwxr-xr-x     - brooks 29 Nov  2021  contexts/
drwx------     - brooks  6 Jan 09:36  keys/
drwxr-xr-x     - brooks  6 Jan 09:29  prekeys/
.rw-r--r--   634 brooks  5 Jan 13:00  host_config.json

~/.wash ➜ ll keys
Permissions Size User   Date Modified Name
.rw-------    58 brooks  6 Jan 09:36  brooks_account.nk
.rw-------    58 brooks  6 Jan 09:36  echo_module.nk
```

LGTM! Cool thing is that this will also retroactively fix the permissions of the `keys` directory too (not existing keys) so this will improve security regardless.

I don't have any reservation with accepting the partial fix and accepting Windows later, especially for something that's non-breaking

brooksmtownsend:
@protochron looks like you might just need a rebase since we merged another PR

src/keys.rs
```diff
@@ -218,9 +218,12 @@ If you'd like to use alternative keys, you can supply them as a flag.\n",
 
                 let kp = KeyPair::new(keygen_type);
                 let seed = kp.seed()?;
-                fs::create_dir_all(Path::new(&path).parent().unwrap())?;
-                let mut f = File::create(path)?;
+                let key_path = Path::new(&path).parent().unwrap();
+                fs::create_dir_all(key_path)?;
```
thomastaylor312, line 210: Nit: I feel like this might be a bit cleaner if we create an `ensure_key_dir` function that makes sure the directory exists with the correct permissions and a `write_key` function that takes the bytes and does everything to create the file with the right permissions. To be clear, this is only a nit though

thomastaylor312 (dismissed):
This is totally fine to merge now if you want to handle windows in a follow up. I couldn't figure out the windows stuff myself in another project, so if you know how, I'll probably end up copying it!

src/keys.rs
protochron: Agreed, but I might tackle that in a follow-up PR, since I may end up reworking that function anyway

protochron:
@brooksmtownsend rebased! I think someone needs to approve running the other github actions.

brooksmtownsend (dismissed):
Just approved the test re-run, still LGTM!

thomastaylor312 (approved):

PR #226, opened 2022-01-13
use lattice controller client & wasmbus linked to nats-aflowt

stevelr:


autodidaddict (dismissed):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.7.2"
+version = "0.7.3"
```
brooksmtownsend: Let's go ahead and bump this to `0.8.0`. This itself isn't really a breaking change, BUT there have been breaking changes since `0.7.2` and we need to release `0.8.0` to compensate

src/ctl/mod.rs
```diff
@@ -932,7 +913,7 @@ async fn ctl_client_from_opts(opts: ConnectionOpts) -> Result<CtlClient> {
     // Determine connection parameters, taking explicitly provided flags,
     // then provided context values, lastly using defaults
     //TODO: Deprecate the `opts.timeout` in `v0.8.0`
-    let timeout = match (opts.timeout_ms, opts.timeout) {
+    let timeout_ms = match (opts.timeout_ms, opts.timeout) {
```
brooksmtownsend: we're bumping to `v0.8.0` here, I don't have a preference if we push the deprecation back to `v0.9.0` or just remove it now

brooksmtownsend (dismissed):
just a request to bump to `0.8.0` as we've had breaking changes make it into main

Cargo.toml
stevelr: done

src/ctl/mod.rs
stevelr: removed deprecated timeout flags

brooksmtownsend (approved):
Big fan of the structopt -> clap changes with subcommand

PR #230, opened 2022-01-18
Renamed windows set_keys util function

brooksmtownsend:
Renamed a util function for windows that caused a compilation error. Bumped a patch version to compensate, and also added a matrix of OS'es so we can ensure we don't run into this again.


autodidaddict (dismissed):

autodidaddict (dismissed):

brooksmtownsend:
@autodidaddict apologies for the continued requested reviews, keeps dismissing the approval when I change a minor thing (I thought the tests would pass last time)

autodidaddict (approved):

PR #234, opened 2022-01-22
fix: use globs to copy lib files to docker image

### LINKED ISSUE #233: [BUG] Docker CI build failing on missing lib files
**Describe the bug**

In the past few commits to `main`, the Docker CI jobs have all failed.

It appears the `rust:alpine` image is missing two files that our `Dockerfile` tries to copy over into the final image:

- /usr/lib/libcrypto.so.46
- /usr/lib/libssl.so.48

**To Reproduce**
1. Merge a PR into main.
2. Observe the failed Docker builds.

**Expected behavior**
The Docker builds not to fail due to missing files.

**Environment (please complete the following information)**
N/A, CI problem

**Additional context**
I tried pulling down the `rust:alpine` image we use, and then installed the packages from the first line of the Dockerfile:
`apk add --no-cache clang clang-dev libressl-dev ca-certificates musl-dev llvm-dev clang-libs curl gcompat`

When I enter the `/lib` directory these libs are present:
Screen Shot 2022-01-22 at 3 15 57 PM

It appears the ending number on `libcrypto` and `libssl` have changed to `.1.1` from `.46` and `.48` respectively. The dumb fix is to just change the file names, but then we'd just run into this problem again if another name changed.

ghost:
Fixes #233

Just changing the COPY commands to use globs to grab the right file regardless of the numbers at the end. This builds without error on my machine now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,11 +29,11 @@ COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=builder /etc/passwd /etc/passwd
 COPY --from=builder /etc/group /etc/group
 
-COPY --from=builder /lib/ld-musl-x86_64.so.1 /lib/ld-musl-x86_64.so.1
-COPY --from=builder /usr/lib/libssl.so.48 /usr/lib/libssl.so.48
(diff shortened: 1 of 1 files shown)


autodidaddict (approved):

brooksmtownsend (approved):
Works for me 👍🏻 🎉

PR #236, opened 2022-01-31
Support provider configuration json for start

brooksmtownsend:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4294,7 +4294,7 @@ dependencies = [
 
 [[package]]
 name = "wash-cli"
-version = "0.8.1"
+version = "0.9.0"
 dependencies = [
  "anyhow",
  "assert-json-diff",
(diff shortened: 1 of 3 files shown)


autodidaddict (approved):

thomastaylor312:
Is this a breaking change? Seems more like a new feature

autodidaddict:
The feature has been part of the control interface for quite some time... wash just didn't support specifying it (it was explicitly forcing provider startup configuration to `None`)

thomastaylor312:
Ok, going to remove the breaking label then

stevelr:
Is this a machine-generated config or hand-generated? for hand-generated, some of us prefer toml over json. Capability provider configs and codegen are usually toml files. It it's machine generated, slight preference for json.

autodidaddict:
it depends entirely on the provider. It's an opaque **json** string by convention. Each provider has the power to do with this string as they please. But, on the wire, we just treat it like a string.

PR #239, opened 2022-02-17
Update wash docker deploy pattern and add alpine image

jordan-rash:
In order to support the tutorials in the vendor environment, they require containers have `/bin/bash`. This PR is an attempt to add an `alpine` tag to dockerhub to satisfy that requirement.  Also, I have ripped out some of the bespoke tagging steps in favor of the docker published tagging method.

The if statements are in place to help me test with my fork.  Can remove if needed

Signed-off-by: Jordan Rash <15827604+jordan-rash@users.noreply.github.com>


Dockerfile
```diff
@@ -23,6 +23,9 @@ RUN adduser \
 ENV RUSTFLAGS=-Ctarget-feature=-crt-static
 RUN cargo build --release
 
+RUN mv /build/target/release/wash /usr/local/bin
```
protochron: Instead of releasing the build image with wash on the path, it seems like it might make more sense to build out an entirely separate alpine-based image as a new build target and point to that instead.
Ex
```docker
FROM alpine as release-alpine
RUN apk add bash libressl-dev ca-certificates && \
         rm -rf /var/cache/apk/*
COPY --from=builder /build/target/release/wash /usr/local/bin
...
```

jordan-rash: Sounds like a good plan to me....updating now

jordan-rash:
Successful run of new GHA file: https://github.com/jordan-rash/wash/runs/5251808902?check_suite_focus=true

brooksmtownsend (dismissed):

Dockerfile
```diff
@@ -23,6 +23,21 @@ RUN adduser \
 ENV RUSTFLAGS=-Ctarget-feature=-crt-static
 RUN cargo build --release
 
+FROM alpine as release-alpine
+COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
+COPY --from=builder /etc/passwd /etc/passwd
+COPY --from=builder /etc/group /etc/group
+
+COPY --from=builder /lib/ld-musl-x86_64.so.* /lib/
```
protochron: You don't need any of these .so files, or really any of these COPY statements other than the wash one since you're making an alpine image and should already be there.

protochron: If the .so files are missing, then I'd install them from the corresponding packages

jordan-rash: When I did that, I got a bunch of GCC errors.  I need to drop the user down (good reminder) since these will be used in a training env...so I will probably need to keep the passwd && group for now

I recognize that I could just apt install these all again, but since the builder image is doing it, it just faster to COPY them.  But if we would prefer just installing them again, i can change it.

```
~/L/M/c/g/w/wash >  docker run -it --rm cjrash/wash wash -V
Error loading shared library libssl.so.50: No such file or directory (needed by /usr/local/bin/wash)
Error loading shared library libcrypto.so.47: No such file or directory (needed by /usr/local/bin/wash)
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /usr/local/bin/wash)
Error relocating /usr/local/bin/wash: EVP_PKEY_set1_DSA: symbol not found
Error relocating /usr/local/bin/wash: EVP_get_digestbyname: symbol not found
Error relocating /usr/local/bin/wash: EC_KEY_generate_key: symbol not found
Error relocating /usr/local/bin/wash: PEM_read_bio_ECPrivateKey: symbol not found
< snip >
```

When going back and adding them one at a time, this was the minimum set of libs to make `wash -V` work. 
```
 28 COPY --from=builder /usr/lib/libssl.so.* /usr/lib/
 29 COPY --from=builder /usr/lib/libcrypto.so.* /usr/lib/
 30 COPY --from=builder /usr/lib/libgcc_s.so.* /usr/lib/
 31 COPY --from=builder /lib/ld-musl-x86_64.so.* /lib/
```

protochron: Sorry, that's on me for saying earlier that you could (and should) get all of those .so files by installing the corresponding packages. You can replace all of those `COPY` instructions with

`RUN apk --no-cache bash libgcc libressl-dev ca-certificates`

which accomplishes the same thing, albeit with a few more dependencies. Installing the packages also has the side benefit of not breaking `apk`, which I think would be surprising to someone if they were to pull the alpine-based container and then try and do any work it in after overriding the entrypoint.

stevelr: +1 on installing packages over copying `.so` files.

jordan-rash: > if they were to pull the alpine-based container and then try and do any work it in after overriding the entrypoint.

So this is the exact use case we are trying to avoid with a image for training.  We dont really want to offer much outside of running `wash`.  The concern is that if someone logs into the training platform, they can start running something like a miner on infrastructure we have to pay for.  There is a timeout native to the training environment, but its long enough that this is something we have to think about.  Maybe the `alpine` tag isnt the best for what is being done here...

I will switch the COPYs to another install, as the side effect is only adding a few seconds to the build

brooksmtownsend (commented):
No additional feedback, this looks good to me once it looks good to @protochron

protochron (approved):

ghost:
@jordan-rash it looks like the push to Docker hub failed from these changes. Do you know what could have caused that? The error is:
```
#20 ERROR: denied: requested access to the resource is denied
------
 > pushing wasmcloud/wash:alpine with docker:
------
error: denied: requested access to the resource is denied
Error: buildx failed with: error: denied: requested access to the resource is denied
```

I'm guessing it is a credentials problem?

jordan-rash:
Sorry @mattwilkinsonn , yea I caught this and spoke to @brooksmtownsend out of band about it.  I havent been able to recreate the issue in my fork, so we might have to revert this until we figure out whats going on.  The action seems to suggest the login to docker hub succeeded. 🤷🏼

PR #240, opened 2022-02-20
feat: Add git initialization for `wash new actor` and flag to disable it

### LINKED ISSUE #232: [FEATURE] add 'git init' for new project creation
The `wash new actor` command creates a project folder containing a `.cargo` subdirectory. The subdirectory might be overlooked by a developer when checking into git. We may be able to prevent that problem by running `git init` as part of the project creation process, in a way that ensures the `.cargo` subdirectory is included in the repo.

ghost:
Fixes #232.

This PR adds a flag to the `wash new` commands, `--no-git-init`, which disables the git repository initialization. Without that passed, the `make_project` function will use `git2::Repository::init()` to create a new repo, which is the same as [Cargo](https://github.com/rust-lang/cargo/blob/master/src/cargo/util/vcs.rs#L34-L36).

There doesn't seem to be any tests for this command atm. Should I add some?

### DIFF
--- a/src/generate/mod.rs
+++ b/src/generate/mod.rs
@@ -158,6 +158,10 @@ pub(crate) struct NewProjectArgs {
     /// template name - name of template to use
     #[clap(short, long)]
     pub(crate) template_name: Option<String>,
+
+    /// Don't create a git repository. Will create one if this is not passed.
+    #[clap(long)]
+    pub(crate) no_git_init: bool,
 }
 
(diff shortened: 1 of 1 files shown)


stevelr (approved):
Thanks Matt! I tested your PR for new actor/interface/provider and it looks good.
I think it would be good to add a test case for this new flag in the project-templates repo in `test/generate-all-templates.sh`

PR #245, opened 2022-03-06
feat: Implement wait for `ctl` commands

### LINKED ISSUE #91: Implement wait flag for ctl commands
Imperatively interacting with a wasmcloud host, or a lattice of interconnected hosts, it's important to be able to know when certain events take place. For example, when you issue this command to a wasmcloud host:
```
wash ctl start actor wasmcloud.azurecr.io/subscriber:0.2.0
```
You receive an acknowledgment that the actor start event has been _received_. All this means is that your command was successfully delivered to the host and that the host is going to proceed with pulling the actor from our OCI registry and starting it. What you don't get from this acknowledgment is the assurance that the actor actually _was_ started.

I propose a `--wait` flag that can be appended to the `start`, `stop`, `link` and `update` subcommands under `ctl`. These are all commands that return with an acknowledgment but not a confirmation of the command executing. When the `--wait` flag is supplied, the command will instead block until the appropriate event can be observed (e.g. ActorStarted for `start actor`).

In order to monitor the event stream, there are two possibilities:
- When connected to hosts running in a lattice via NATS, events are published as NATS messages that can be monitored. This can be used to monitor events when connected via NATS.
- When connected to a host without NATS (which will be a possibility in the REPL after #92 ) the host can be built with an event sender that can be monitored for the same events that would be sent across NATS. 

Both of these possibilities are important. When using `wash` as a command-line interface, we'll want to monitor events that are flowing over NATS messages, as the control interface operates on a lattice that's connected to via NATS. We'll want to do the same thing when `wash up` is launched with a NATS connection. When the REPL is launched without a NATS connection (after #92 ) we should directly build a host with an event receiver that can be used to monitor these events.

ghost:
Fixes #91.

In this PR I am implementing a way for all `ctl` commands to wait for the outcome of the command, instead of just returning after acknowledgement. This will modify all of these commands to wait by default, with a `--skip-wait` flag that can be passed to mimic the current behavior.

So far I've done this for `ctl start actor`. I'd like to hear feedback on this implementation before I go on to the rest of the commands.

### DIFF
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -380,7 +380,7 @@ pub(crate) struct StartProviderCommand {
     #[clap(long = "skip-wait")]
     skip_wait: bool,
 
-    /// Timeout to await the provider start, defaults to 3000 milliseconds.
+    /// Timeout to await the provider start, defaults to 15000 milliseconds.
     #[clap(long = "start-timeout-ms", default_value_t = 15000)]
     wait_timeout_ms: u64,
 }
@@ -1235,7 +1235,7 @@ mod test {
                 assert_eq!(host_id.unwrap(), HOST_ID.parse()?);
                 assert_eq!(provider_ref, "wasmcloud.azurecr.io/provider:v1".to_string());
                 assert!(skip_wait);
-                assert_eq!(wait_timeout_ms, 3000);
+                assert_eq!(wait_timeout_ms, 15000);
             }
             cmd => panic!("ctl start provider constructed incorrect command {:?}", cmd),
         }
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -23,7 +23,7 @@ use wasmcloud_control_interface::{
     LinkDefinitionList,
 };
 
-use self::wait::{find_event, find_start_actor_result_event};
+use self::wait::{wait_for_actor_start_event, FindEventOutcome};
 
 mod manifest;
 mod output;
@@ -710,19 +710,22 @@ pub(crate) async fn start_actor(mut cmd: StartActorCommand) -> Result<CommandOut
         ));
     }
 
-    let start_actor_result_checker =
-        find_start_actor_result_event(host.to_string(), cmd.actor_ref.clone());
-
-    find_event(
+    let event = wait_for_actor_start_event(
         &receiver,
         Duration::from_millis(cmd.start_timeout_ms),
-        start_actor_result_checker,
-    )??;
+        host.to_string(),
+        cmd.actor_ref.clone(),
+    )?;
 
-    Ok(CommandOutput::from_key_and_text(
-        "result",
-        format!("Actor {} started on host {}", cmd.actor_ref, host),
-    ))
+    match event {
+        FindEventOutcome::Success(_) => Ok(CommandOutput::from_key_and_text(
+            "result",
+            format!("Actor {} started on host {}", cmd.actor_ref, host),
+        )),
+        FindEventOutcome::Failure(err) => {
+            bail!("{}", err);
+        }
+    }
 }
 
 pub(crate) async fn start_provider(mut cmd: StartProviderCommand) -> Result<CtlOperationAck> {
--- a/src/ctl/mod.rs
+++ b/src/ctl/mod.rs
@@ -23,7 +23,10 @@ use wasmcloud_control_interface::{
     LinkDefinitionList,
 };
 
-use self::wait::{wait_for_actor_start_event, wait_for_provider_start_event, FindEventOutcome};
+use self::wait::{
+    wait_for_actor_scale_event, wait_for_actor_start_event, wait_for_provider_start_event,
+    FindEventOutcome,
+};
 
 mod manifest;
 mod output;
@@ -282,6 +285,15 @@ pub struct ScaleActorCommand {
     /// For example, autonomous agents may wish to “tag” scale requests as part of a given deployment
     #[clap(short = 'a', long = "annotations")]
     pub annotations: Vec<String>,
+
+    /// By default, the command will wait until the actor has been started.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the actor to start.
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await an actor start, defaults to 3000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 3000)]
+    start_timeout_ms: u64,
 }
 
 #[derive(Debug, Clone, Parser)]
@@ -598,16 +610,7 @@ pub(crate) async fn handle_command(
                 " Scaling Actor {} to {} instances ... ",
                 cmd.actor_id, cmd.count
             ));
-
-            let ack = scale_actor(cmd.clone()).await?;
-            if !ack.accepted {
-                bail!("Operation failed: {}", ack.error);
-            }
-
-            CommandOutput::from_key_and_text(
-                "result",
-                format!("Actor {} scaled to {} instances", cmd.actor_id, cmd.count),
-            )
+            scale_actor(cmd.clone()).await?
         }
     };
 
@@ -723,9 +726,7 @@ pub(crate) async fn start_actor(mut cmd: StartActorCommand) -> Result<CommandOut
             "result",
             format!("Actor {} started on host {}", cmd.actor_ref, host),
         )),
-        FindEventOutcome::Failure(err) => {
-            bail!("{}", err);
-        }
+        FindEventOutcome::Failure(err) => bail!("{}", err),
     }
 }
 
@@ -814,21 +815,54 @@ pub(crate) async fn start_provider(mut cmd: StartProviderCommand) -> Result<Comm
     }
 }
 
-pub(crate) async fn scale_actor(cmd: ScaleActorCommand) -> Result<CtlOperationAck> {
+pub(crate) async fn scale_actor(cmd: ScaleActorCommand) -> Result<CommandOutput> {
     let client = ctl_client_from_opts(cmd.opts, None).await?;
 
     let annotations = labels_vec_to_hashmap(cmd.annotations)?;
 
-    client
+    let receiver = client.events_receiver().await.map_err(convert_error)?;
+
+    let ack = client
         .scale_actor(
             &cmd.host_id.to_string(),
             &cmd.actor_ref,
             &cmd.actor_id.to_string(),
             cmd.count,
-            Some(annotations),
+            // Some(annotations),
+            None,
         )
         .await
-        .map_err(convert_error)
+        .map_err(convert_error)?;
+
+    if !ack.accepted {
+        bail!("Operation failed: {}", ack.error);
+    }
+
+    if cmd.skip_wait {
+        return Ok(CommandOutput::from_key_and_text(
+            "result",
+            format!(
+                "Request to scale actor {} to {} instances recieved",
+                cmd.actor_id, cmd.count
+            ),
+        ));
+    }
+
+    let event = wait_for_actor_scale_event(
+        &receiver,
+        Duration::from_millis(cmd.start_timeout_ms),
+        cmd.host_id.to_string(),
+        cmd.actor_id.to_string(),
+        cmd.actor_ref.to_string(),
+    )?;
+
+    match event {
+        FindEventOutcome::Success(_) => Ok(CommandOutput::from_key_and_text(
+            "result",
+            format!("Actor {} scaled to {} instances", cmd.actor_id, cmd.count),
+        )),
+        FindEventOutcome::Failure(err) => bail!("{}", err),
+    }
 }
 
 pub(crate) async fn stop_provider(cmd: StopProviderCommand) -> Result<CtlOperationAck> {
@@ -1420,6 +1454,8 @@ mod test {
                 actor_ref,
                 count,
                 annotations,
+                skip_wait,
+                start_timeout_ms,
             })) => {
                 assert_eq!(&opts.ctl_host.unwrap(), CTL_HOST);
                 assert_eq!(&opts.ctl_port.unwrap(), CTL_PORT);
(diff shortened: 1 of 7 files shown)


src/ctl/mod.rs
```diff
@@ -328,6 +331,15 @@ pub(crate) struct StartActorCommand {
     /// Timeout to await an auction response, defaults to 2000 milliseconds
     #[clap(long = "auction-timeout-ms", default_value_t = default_timeout_ms())]
     auction_timeout_ms: u64,
+
+    /// By default, the command will wait until the actor has been started.
+    /// If this flag is passed, the command will return immedately after acknowledgement from the host, without waiting for the actor to start.
```
brooksmtownsend: nit: typo in `immediately`

src/ctl/mod.rs
```diff
@@ -328,6 +331,15 @@ pub(crate) struct StartActorCommand {
...
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await an actor start, defaults to 10000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 10000)]
+    start_timeout_ms: u64,
```
brooksmtownsend: I feel like this is a little long, but I think we can hone in on a proper "normal" timeout as we test this out a bit. If something like 3000ms covers the 99% of actor starts that should be plenty, and since we listen for actor start / actor start failed it should still be very quick.

src/ctl/mod.rs
```diff
@@ -715,10 +720,37 @@ pub(crate) async fn start_actor(mut cmd: StartActorCommand) -> Result<CtlOperati
...
+        .map_err(convert_error)?;
+
+    if !ack.accepted {
+        bail!("Operation failed: {}", ack.error);
+    }
+
+    if cmd.skip_wait {
+        return Ok(CommandOutput::from_key_and_text(
+            "result",
+            format!("Starting actor {} on host {}", &cmd.actor_ref, &host),
```
brooksmtownsend: To avoid confusion, I prefer the `Start actor request received:` language in the original, that way it's explicit that the host received and acknowledged the request. I like the idea of including the host ID like you did here.

src/ctl/mod.rs
```diff
@@ -715,10 +720,37 @@ pub(crate) async fn start_actor(mut cmd: StartActorCommand) -> Result<CtlOperati
...
+    }
+
+    let start_actor_result_checker =
+        find_start_actor_result_event(host.to_string(), cmd.actor_ref.clone());
+
+    find_event(
+        &receiver,
+        Duration::from_millis(cmd.start_timeout_ms),
+        start_actor_result_checker,
+    )??;
```
brooksmtownsend: I feel like it would be more ergonomic to do something like
```
let start_actor_result = wait_for_actor_start_event(receiver, timeout, host, actor_ref)?;
```

And instead wrap the `find_event` logic within the `wait_for_actor_start_event` function. See other comment below for more

brooksmtownsend: Of course, feel free to rename the function as you see fit, it's long but descriptive

src/ctl/wait.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        let outcome = check_function(event)?;
+
+        if let Some(result) = outcome {
+            return Ok(result);
+        }
+    }
+}
+
+pub(crate) fn find_start_actor_result_event(
```
brooksmtownsend: As mentioned above, I'd rather see this function take the receiver and the timeout and call find_event internally. I'm fine with this function passing this closure to `find_event`

src/ctl/wait.rs
```diff
@@ -0,0 +1,80 @@
...
+                    .ok_or(anyhow!("No image_ref found in data"))?
+                    .as_str()
+                    .ok_or(anyhow!("image_ref is not a string"))?
+                    .to_string();
+
+                if image_ref == actor_ref {
+                    return Ok(Some(Ok(())));
+                }
+            }
+            "com.wasmcloud.latticeactor_start_failed" => {
```
brooksmtownsend: This needs to be `com.wasmcloud.lattice.actor_start_failed`

src/ctl/wait.rs
```diff
@@ -0,0 +1,80 @@
...
+        if let Some(result) = outcome {
+            return Ok(result);
+        }
+    }
+}
+
+pub(crate) fn find_start_actor_result_event(
+    host_id: String,
+    actor_ref: String,
+) -> impl Fn(Event) -> Result<Option<Result<()>>> {
```
brooksmtownsend: So here's the scenarios that I see in this function
1. Event received that isn't relevant
2. Error coercing event types or unwrapping its fields
3. Matching event received (success)
4. Matching event received (failure)

Because of this, I think we could simplify this using an enum where the variants are
1. Success
1. Failed
1. NotApplicable

That way we return an `Err` when we get unexpected things when deserializing or unwrapping, and `Ok` with the enum variant that describes that event. Curious what you think about simplifying this Return type a bit

brooksmtownsend (changes requested):
This is a great start, and handles the various cases well for starting an actor. I had some code ergonomics suggestions and some minor nits

src/ctl/wait.rs
ghost: Woops. Will fix

src/ctl/wait.rs
ghost: So you are suggesting something like `Result<ResponseEnum>`? Where `ResponseEnum` has those three variants. Note ResponseEnum is a working title, I'll come up with a better one

src/ctl/mod.rs
ghost: Yeah I like that approach. Will work on implementing it

src/ctl/mod.rs
ghost: Yeah I wasn't sure what the best timeout was, just stuck 10sec in there to be safe to start with. I can drop it to 3000

src/ctl/wait.rs
brooksmtownsend: That's what I could think of to simplify the cases a little bit, yes

ghost:
@brooksmtownsend I've made all of the changes from your feedback. Could you take a look when you have a bit? Thanks!

brooksmtownsend (dismissed):
Changes look great! I'd give a 👍🏻 to proceed with provider start too

ghost:
I've finished adding the wait parameters to the commands that can support them right now (that have `CloudEvents` being sent over the `wasmbus.evt.*` channel), and rebased to pick up the spinner changes. However the CI is failing on a weird fmt error:

```
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
```

Is this new? Can we just turn that off in the `rustfmt.toml`?

ghost:
@brooksmtownsend this is ready for review

src/ctl/mod.rs
```diff
@@ -399,6 +432,15 @@ pub(crate) struct StopProviderCommand {
...
     #[clap(name = "contract-id")]
     pub(crate) contract_id: String,
+
+    /// By default, the command will wait until the provider has been stopped.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the provider to stop.
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await the provider stop, defaults to 3000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 3000)]
```
stevelr: the command is `stop provider` but this arg is `start-timeout-ms`. I think it would be simpler and less for people to remember if we use `timeout-ms` for all timeout args (unless there's a conflict within the same sub-command and a prefix is needed to disambiguate between two separate timeouts)

src/ctl/mod.rs
```diff
@@ -377,6 +401,15 @@ pub(crate) struct StopActorCommand {
...
     #[clap(long = "count", default_value = "1")]
     pub(crate) count: u16,
+
+    /// By default, the command will wait until the actor has been stopped.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the actor to stp[].
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await rhe actor stop, defaults to 3000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 3000)]
```
stevelr: change to `long = "timeout-ms"`
(should have been `stop-timeout-ms` but I think dropping the start/stop prefix is preferable)

src/ctl/mod.rs
```diff
@@ -359,6 +374,15 @@ pub(crate) struct StartProviderCommand {
...
     #[clap(long = "config-json")]
     config_json: Option<PathBuf>,
+
+    /// By default, the command will wait until the provider has been started.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the provider to start.
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await the provider start, defaults to 3000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 3000)]
```
stevelr: change to long = "timeout-ms"

src/ctl/mod.rs
```diff
@@ -328,6 +334,15 @@ pub(crate) struct StartActorCommand {
...
     #[clap(long = "auction-timeout-ms", default_value_t = default_timeout_ms())]
     auction_timeout_ms: u64,
+
+    /// By default, the command will wait until the actor has been started.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the actor to start.
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await an actor start, defaults to 3000 milliseconds.
+    #[clap(long = "start-timeout-ms", default_value_t = 3000)]
```
stevelr: change to long = "timeout-ms"

src/ctl/mod.rs
```diff
@@ -377,6 +401,15 @@ pub(crate) struct StopActorCommand {
     /// Number of actors to stop
     #[clap(long = "count", default_value = "1")]
     pub(crate) count: u16,
+
+    /// By default, the command will wait until the actor has been stopped.
+    /// If this flag is passed, the command will return immediately after acknowledgement from the host, without waiting for the actor to stp[].
+    #[clap(long = "skip-wait")]
+    skip_wait: bool,
+
+    /// Timeout to await rhe actor stop, defaults to 3000 milliseconds.
```
brooksmtownsend: nit: typo `await rhe actor`

src/ctl/wait.rs
```diff
@@ -0,0 +1,307 @@
...
+/// with the `FindEventOutcome` enum containing the success or failure state of the event.
+///
+/// If the timeout is reached or another error occurs, the `Err` variant of the `Result` will be returned.
+///
+/// You can use the generics in `EventCheckOutcome` and `FindEventOutcome` to return any data from the event out of your `check_function`.
+fn find_event<T>(
+    receiver: &Receiver<Event>,
+    timeout: Duration,
+    check_function: impl Fn(Event) -> Result<EventCheckOutcome<T>>,
+) -> Result<FindEventOutcome<T>> {
```
brooksmtownsend, line 20: One thing I'm considering would be useful here, especially for the actor start/stop events, is a `count` parameter where you can wait for a specified number of events. 

For example, if you do a `wash ctl start actor <echo> --count 5`, the ideal scenario would be to wait for 5 `ActorStarted` events. 

I'm personally willing to file that as a feature request though and it doesn't need to be done in this PR. I think I lean towards the feature request only because waiting for multiple events introduces a few questions that I don't have a good answer to, e.g.
1. What if you receive 3/5 successful events before the timeout, is that a mixed success or a fail?
2. What if you receive 99 successful events and 1 failure?
3. What if you receive 1 successful event and 99 failures?

TL;DR I think we feature request awaiting multiple as there are some nuances

brooksmtownsend (commented):
Overall this looks awesome and I'm installing it to test now... just left one nit for a typo and a larger comment meant for some discussion

src/ctl/mod.rs
brooksmtownsend: For starting a provider, this timeout may need to be a good bit larger as it's a larger file to download. I'd probably set the default at 15 seconds, in the general case this will be enough time and in the constrained device case we can always set it higher.

brooksmtownsend (changes requested):
image

Works like a charm! Requesting one change regarding the start provider timeout, and agree with Steve's suggestions.

src/ctl/mod.rs
ghost: The problem is that the `ConnectionOps` struct already has a `timeout_ms` value, which indicates the timeout for receiving the ack from the lattice: https://github.com/wasmCloud/wash/blob/main/src/ctl/mod.rs#L60-L67

I can fix it so it is 'start' and 'stop' in the correct places though

src/ctl/wait.rs
ghost: Yeah that would be a good feature, I agree it should probably be handled in another PR though

src/ctl/mod.rs
stevelr: Then I would make it so timeout-ms is the timeout for the operation, and ack-timeout-ms (or connect-timeout-ms) is timeout for the connection Ack. 

More than 99% use case would be timeout for the operation, so that should be the simpler form

brooksmtownsend (dismissed):
My feedback was addressed, @mattwilkinsonn you can go ahead and bump this to `0.10.0-alpha.2` so we can release an alpha version once this merges 👍🏻 

@stevelr any other feedback?

ghost:
Last thing - is there a reason we disable the help flag on these ctl commands? Seems rather user unfriendly. I see that there's a few flags with a `-h` short option, could we remove those and enable all of the helps?

Can do this in another PR though

brooksmtownsend:
> Last thing - is there a reason we disable the help flag on these ctl commands? Seems rather user unfriendly. I see that there's a few flags with a -h short option, could we remove those and enable all of the helps?

I wasn't aware we did this... we can keep that separate from this PR but every command should support `-h` and `--help`

src/ctl/mod.rs
ghost: Fixed, made the action timeout `timeout_ms`, and the ack timeout `ack_timeout_ms`.

ghost:
All feedback is resolved @brooksmtownsend @stevelr. Should be ready to merge

ghost:
Also - the tests didn't run on my last commit? Is there something weird going on there? It only checked the signoff

EDIT: nvm, looks like they ran now

stevelr:
LGTM

brooksmtownsend (approved):
Awesome work again @mattwilkinsonn !

PR #246, opened 2022-03-07
Issue 57 - Use indicatif instead of spinners library for command line waiting spinner

byblakeorriver:
Signed-off-by: bwilloughby <bwilloughby@demark.com>


byblakeorriver:
@brooksmtownsend I think this one is ready to go. Let me know if you have any concerns with the way I did it.

src/reg.rs
```diff
@@ -134,14 +134,12 @@ pub(crate) async fn handle_pull(
...
-            format!(" Downloading {} ...", image.whole()),
-        )),
-        _ => None,
-    };
-    info!("Downloading {}", image.whole());
+
+    let spinner = Spinner::new(&output_kind);
+    let msg = format!(" Downloading {} ...", image.whole());
+    spinner.update_spinner_message(msg.clone());
+    info!("{}", msg);
```
brooksmtownsend: I think this can be removed and might be causing the issue I can reproduce

src/reg.rs
```diff
@@ -337,14 +334,10 @@ pub(crate) async fn handle_push(
...
-            &Spinners::Dots12,
-            format!(" Pushing {} to {} ...", cmd.artifact, cmd.url),
-        )),
-        _ => None,
-    };
-    info!(" Pushing {} to {} ...", cmd.artifact, cmd.url);
+    let spinner = Spinner::new(&output_kind);
+    let msg = format!(" Pushing {} to {} ...", cmd.artifact, cmd.url);
+    spinner.update_spinner_message(msg.clone());
+    info!("{}", msg);
```
brooksmtownsend: I think this can be removed and might be causing the issue I can reproduce

src/appearance/spinner.rs
```diff
@@ -0,0 +1,51 @@
...
+    spinner: Option<ProgressBar>,
+}
+
+impl Spinner {
+    pub(crate) fn new(output_kind: &OutputKind) -> Self {
+        match output_kind {
+            OutputKind::Text => {
+                let style = ProgressStyle::default_spinner()
+                    .tick_strings(DOTS_12)
+                    .template("{spinner:.blue} {msg}");
```
brooksmtownsend: The blue looks great in my terminal, but on Windows Powershell the default background color is blue so it's mostly invisible. I think a mid-gray might be a better option so that it shows on all colors including light themes

brooksmtownsend (changes requested):
Overall this looks awesome! Cleans up the code and our dependencies 😄 

I did notice one quirk about using the `reg` functions, it looks like there is an extra line with one dot that gets printed, see picture.
image

I don't see the same thing with the `ctl` commands, perhaps it's due the the info log that I reference in my review?

I also tested this on Windows in Powershell and it worked well, just see my comment around the blue spinner.

src/appearance/spinner.rs
byblakeorriver: Gotchya!

src/reg.rs
byblakeorriver: I thought these might cause some trouble. I didn't know if the logging was there for any particular reason so I kept them. I will get it cleaned up.

byblakeorriver:
@brooksmtownsend can you give this last commit a try when you get a chance? I tried to recreate your issue, but was not able to get it to occur. I tried on two terminals (alacritty and the default pop_os terminal). I also tried both of those using the fish shell to see if that could maybe reproduce it. 

I did however find an issue if the terminal window is really small. It would print the message over and over again. I fixed that using the wide_msg template setting in the indicatif lib. Maybe that could have something to do with it.

I did get the color changed. I couldn't find a specific option for the gray, but I did find a way to make it bold and dim. Let me know what you think. I also added a prefix of (>>>) not sure if that is ok with you. It could be anything. An emoji or just a couple spaces.

byblakeorriver:
I was able to find a way to choose any of these colors: https://www.ditig.com/256-colors-cheat-sheet

If you don't like the bold dim version, I change the colors, just need some artistic advice on which one to choose.

autodidaddict (approved):

brooksmtownsend (approved):
This is great, and I didn't even know we could fix the printing over and over again issue 😄

Issue #249, opened 2022-03-19
[BUG] wash ctl link put base64 padding character removal


emattiza:
**Describe the bug**
When parsing key value options provided to a link configuration, and a value contains `=`, the str.split function removes the `=` inside of the value token. This results in our base64 (which uses `=` as a padding character) config option failing to decode when supplied to the provider, and failing to start the link successfully. Applying the link and then echo'ing our base64 shows proper encode, but wash ctl link query shows a lack of '=' padding in the base64 value.

**Environment (please complete the following information) **
 - OS: Linux
 - bash
 - wash v0.10.0-alpha.1

**Additional context**
At issue is the split logic, which can be remedied through using splitn(2, '=') to parse pairs, and break on first `=` found, while leaving the value token remainder untouched. [See here](https://github.com/wasmCloud/wash/blob/52b61f91fc4fc49f62b039c956403a84eafb6ec9/src/util.rs#L133)

brooksmtownsend:
closed by #250

PR #250, opened 2022-03-19
[BUG] Fix/split options with base64 padding

emattiza:
Addresses #249. First commit adds a test with the reproduction, second commit uses splitn for extracting pairs without value token `=` character removal.

If there are other integration test points for link, happy to add the suitable command testing and integration if pointed in the right direction

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -316,3 +316,19 @@ pub(crate) fn set_permissions_keys(path: &Path) -> Result<()> {
 pub(crate) fn set_permissions_keys(path: &Path) -> Result<()> {
     Ok(())
 }
+
+mod test {
+    use super::labels_vec_to_hashmap;
+    use std::collections::HashMap;
+    #[test]
+    fn test_safe_base64_parse_option() {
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -130,7 +130,7 @@ pub(crate) fn convert_error(e: Box<dyn ::std::error::Error + Send + Sync>) -> an
 pub(crate) fn labels_vec_to_hashmap(constraints: Vec<String>) -> Result<HashMap<String, String>> {
     let mut hm: HashMap<String, String> = HashMap::new();
     for constraint in constraints {
-        let key_value = constraint.split('=').collect::<Vec<_>>();
+        let key_value = constraint.splitn(2, '=').collect::<Vec<_>>();
```
thomastaylor312: Nit: `split_once` might be a better choice here so you don't have to collect to an iterator

thomastaylor312 (dismissed):

src/util.rs
emattiza: that actually captures the bail case below it as well, so I'm happy to make that change. Didn't find that one in docs.rs!

thomastaylor312 (approved):
This is super awesome! Thank you

PR #255, opened 2022-04-05
Fixed breaking clap change, resolved clippy warnings

### LINKED ISSUE #252: [BUG] Enable `--help` flags on all ctl commands
A number of `ctl` commands have their `--help` and `-h` flags disabled with
```
#[clap(setting(AppSettings::DisableHelpFlag))]
```

We want to remove all of these to enable use of the help flag. There are some commands with flags that have a `-h` short option that will need to be removed to prevent conflicts, however.

_Originally posted by @mattwilkinsonn in https://github.com/wasmCloud/wash/issues/245#issuecomment-1076605553_

brooksmtownsend:
Fixes #252 as well

### DIFF
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -259,7 +259,7 @@ pub(crate) struct ActorMetadata {
     #[clap(short = 'q', long = "http_server")]
     http_server: bool,
     /// Enable the HTTP client standard capability
-    #[clap(short = 'b', long = "http_client")]
+    #[clap(short = 'j', long = "http_client")]
     http_client: bool,
     /// Enable access to the blob store capability
     #[clap(short = 'f', long = "blob_store")]
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -17,7 +17,7 @@ use crate::{
     util::{CommandOutput, OutputKind},
 };
 use anyhow::{bail, Context, Result};
-use clap::{AppSettings, Args, Parser, Subcommand};
+use clap::{Args, Parser, Subcommand};
 use nkeys::{KeyPair, KeyPairType};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_json::json;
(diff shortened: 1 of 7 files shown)


src/claims.rs
```diff
@@ -261,7 +259,7 @@ pub(crate) struct ActorMetadata {
     #[clap(short = 'q', long = "http_server")]
     http_server: bool,
     /// Enable the HTTP client standard capability
-    #[clap(short = 'h', long = "http_client")]
+    #[clap(short = 'b', long = "http_client")]
```
protochron: Did you mean to change this?

brooksmtownsend: Yeah I did, if we have `-h` and the help command enabled (re: #252) it'll break. Happy to take suggestions for another letter but all of the other letters are taken for other subcommands 😅

protochron: Seems fine, I just wasn't sure if it was related

stevelr: For consistency, the long args could use hyphens instead of underscore

thomastaylor312:
Looks like the tests are failing due to a name collision:

```
Command sign: Short option names must be unique for each argument, but '-b' is in use by both 'http-client' and 'not-before-days'', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.1.8/src/build/debug_asserts.rs:100:17
```

src/claims.rs
```diff
@@ -259,7 +259,7 @@ pub(crate) struct ActorMetadata {
     #[clap(short = 'q', long = "http_server")]
     http_server: bool,
     /// Enable the HTTP client standard capability
-    #[clap(short = 'b', long = "http_client")]
+    #[clap(short = 'j', long = "http_client")]
```
thomastaylor312: Not to sound too nit-picky, but why `j`? I personally would never remember that as a dev

brooksmtownsend: I honestly wouldn't mind killing the short options. There are too many reserved short options to pick any of the letters that are in `http_client` 😞

autodidaddict (approved):

PR #254, opened 2022-03-25
[FEATURE] overriding usage to add link_name to wash ctl link put

emattiza:
Addresses #247 

This is the current way I've seen to override the usage string for a specific command. I was unable to find a way to make certain [OPTIONS] come out for visibility (like a must_show_usage kind of attribute to add to an option).

Probably need to wrap the link-name option in [] too, but we can review that.

PR #238, opened 2022-02-17
feat: Add scaling actor command

ghost:
Fixes #228.

This PR adds a scaling actor command to `ctl scale actor` using the `scale_actor` method exposed on the `CtlClient` from the `wasmcloud-control-interface` crate. I tried to keep everything as similar to the other commands as possible, and added a test for the various parameters available on the command.

PR #256, opened 2022-04-14
0.10.0 Release

brooksmtownsend:
Release notes, looking for feedback on these as well:
### Breaking Changes
- ⚠️ The default behavior for `ctl` commands is now to wait for a successful event to occur for the corresponding command. For example, `wash ctl start actor [wasmcloud.azurecr.io/echo:0.3.4](http://wasmcloud.azurecr.io/echo:0.3.4)` will now wait for an `ActorStarted` event to be published by a host. This can be ignored in favor of finishing on the early ack by passing the `--skip-wait` flag. This is marked with a caution sign as it may affect existing scripts where timeouts are set to a short value expecting an early ack. Big thanks to @mattwilkinsonn for this massive PR #245
- Removed short `-h` flags from all commands so that `--help` can always be used for context in #255

### New Features
- `wash ctl scale actor` in #238 from @mattwilkinsonn to declaratively scale an actor to a number of replicas
- `--no-git-init` flag to `wash gen` to allow user to initialize their own git repository in #240 from @mattwilkinsonn
- Use `indicatif` instead of `spinners` for long-running operations, this also fixed a harmless panic that would occur in non-interactive TTY sessions (Windows Powershell, Github Actions) in #246 by @byblakeorriver

### Fixes and Misc
- Custom help text for `wash ctl link put` for command clarity in #254 by @emattiza
- Docker deployments to use an alpine image for better support in #239 by @jordan-rash
- Fixed a bug where base64 values with padding in link definitions would be incorrectly parsed in #250 by @emattiza

I'll note it in the release notes as well, but huge thanks to our community members for the features that went into this version!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3777,7 +3777,7 @@ dependencies = [
 
 [[package]]
 name = "wash-cli"
-version = "0.10.0-alpha.3"
+version = "0.10.0"
 dependencies = [
  "anyhow",
  "assert-json-diff",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.10.0-alpha.3"
+version = "0.10.0"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "command-line-utilities"]
 description = "wasmcloud Shell (wash) CLI tool"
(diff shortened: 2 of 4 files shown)


thomastaylor312:
@brooksmtownsend I generally like to put the breaking changes in their own "Breaking" section so that it is super clear to those who are updating what the breaking changes are

thomastaylor312 (approved):
Everything here looks good to me. Just a reminder we should open up a PR on the chocolatey and the brew repos once this is cut

brooksmtownsend:
> @brooksmtownsend I generally like to put the breaking changes in their own "Breaking" section so that it is super clear to those who are updating what the breaking changes are

Good call, I can reorganize it a bit

PR #257, opened 2022-04-22
Added support for oci manifest annotations

brooksmtownsend:
Certain registries, like [GHCR](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) depend on OCI manifest annotations in order to perform special actions. In Github's case, adding an annotation of `org.opencontainers.image.source=<your_repo>` will automatically associate a package (actor/provider) with a github repository.

I think this is _technically_ a minor bump but I left it as a patch since it's so small. Feel free to request that I change that, and I can change to `0.11.0-alpha.1` instead to indicate new features

### DIFF
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -1,10 +1,11 @@
 extern crate oci_distribution;
 
 use crate::appearance::spinner::Spinner;
-use crate::util::{cached_file, CommandOutput, OutputKind};
+use crate::util::{cached_file, labels_vec_to_hashmap, CommandOutput, OutputKind};
 use anyhow::{anyhow, bail, Result};
 use clap::{Parser, Subcommand};
 use log::{debug, warn};
+use oci_distribution::manifest::{OciDescriptor, OciManifest};
 use oci_distribution::{client::*, secrets::RegistryAuth, Reference};
 use provider_archive::ProviderArchive;
 use serde_json::json;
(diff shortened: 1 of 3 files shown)


src/reg.rs
```diff
@@ -423,19 +430,74 @@ pub(crate) async fn push_artifact(
...
             config_media_type,
             &auth,
-            None,
+            Some(manifest),
         )
         .await?;
     Ok(())
 }
 
+/// Modified version of oci_distribution::generate_manifest to support additional annotations
```
thomastaylor312, line 453: Might just want to upstream this change to oci-distribution as we can't be the only one wanting to add multiple annotations

thomastaylor312 (changes requested):
LGTM.

However, I think we should make sure we follow strict semver here (even though it is annoying) as it is a feature add. Although it probably doesn't _actually_ matter here, it does give future users confidence that they know what they are getting with each version

src/reg.rs
brooksmtownsend: I'm down to contribute upstream and then remove this function once it is released 👍🏻

brooksmtownsend: re-reading upstream, I actually don't think that this belongs in `oci_distribution`. I'm really just reusing some of the logic since I need to generate my own manifest, `oci_distribution` will call `generate_manifest` if you haven't supplied one.

thomastaylor312 (approved):

Issue #263, opened 2022-05-11
[FEATURE] `wash build`


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
The easiest way to compile and sign actors today is to use the included Makefile and run `make`. This is simple, but constraining to Windows developers and adds an additional tool that's required for the developer to use.

**Describe the solution you'd like**
I would love to have a subcommand, `wash build`, that auto-detected language (just Rust / TinyGo now, for our examples) and invoked that compiler toolchain on the local machine. In tandem with #262, this could replace the make build/sign steps with a single `wash build` command that builds and signs the actor.

Auto-detecting language to start can be naive, something like looking for a `Cargo.toml` or a `go.mod` file.

**Describe alternatives you've considered**
We could approach this with something like [just](https://github.com/casey/just), though again this just requires an additional install that's more Rust biased.

ghost:
Probably need to have our own configuration file along with this if we're getting rid of the makefiles, an equivalent to `package.json` or `Cargo.toml`. Perhaps just `wash.toml`? or `wasmcloud.toml`?

brooksmtownsend:
> Probably need to have our own configuration file along with this if we're getting rid of the makefiles, an equivalent to package.json or Cargo.toml.

I was wondering if we could get away with storing additional metadata in each language's config file just for ease-of-use, but if that can't be done in `go.mod` / `Cargo.toml` / `package.json` / whatever Zig uses / whatever C# uses then disregard

> Perhaps just wash.toml? or wasmcloud.toml?

`wash.toml` would be great for just wash uses, but I think that there's an opportunity here with the wasmcloud-otp notion of a host configuration file since wash and wasmcloud would share NATS connection information... I think `wasmcloud.toml` is the safer bet even if it starts with just `wash` configuration values

ghost:
> I was wondering if we could get away with storing additional metadata in each language's config file just for ease-of-use, but if that can't be done in `go.mod` / `Cargo.toml` / `package.json` / whatever Zig uses / whatever C# uses then disregard

I think that's a bit risky to rely on the language's config files, they might not have the features/UX we want, with our own file we can control that much more. Also would lead to us having to parse every other language's config files in `wash` just to read that data 😅

[...]

PR #271, opened 2022-05-22
add tinygo actor template

stevelr:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3777,7 +3777,7 @@ dependencies = [
 
 [[package]]
 name = "wash-cli"
-version = "0.10.1"
+version = "0.11.0-alpha.1"
 dependencies = [
  "anyhow",
  "assert-json-diff",
--- a/src/generate/favorites.toml
+++ b/src/generate/favorites.toml
@@ -14,9 +14,15 @@
 
 [[actor]]
 name = "hello"
-description = "a hello-world actor that responds over an http connection"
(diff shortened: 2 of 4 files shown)


stevelr:
also fixes some directory manipulation to make it easier to perform local testing

autodidaddict (approved):

brooksmtownsend (commented):
Looks great, though I worry now that our github actions to build/release actors now have to be language specific 🤔 We may just need to prioritize https://github.com/wasmCloud/wash/issues/263 appropriately in order to build actors in a language-agnostic way

stevelr:
@brooksmtownsend I don't think this is a problem:
- we don't have a `wash build` command , so that's not an issue (yet).
- Codegen _is_ language specific - if you run `wash gen` it reads codegen.toml, which has language-specific sections in it.
- As far as building, for now we just use `make` to build the actor, and they show up in `build/actor_s.wasm`. The makefile should do the right thing.

- we can't do `rustfmt` automatically - for that we'd have to check for the presence of Cargo.toml - (which itself should be inside a common action script).  Or, alternately, use `make test` to run tests or `make check` to check formatting.

stevelr:
This doesn't strictly depend on code generation, so I'll merge now ...

brooksmtownsend:
> As far as building, for now we just use make to build the actor, and they show up in build/actor_s.wasm. The makefile should do the right thing.

Duh! Good point. I may test this once we get everything merged and can make any necessary adjustments to the actions

PR #272, opened 2022-05-23
add codegen for TinyGo actors

stevelr:

Updates dependency on weld-codegen to 0.4.4 which is in PR https://github.com/wasmCloud/weld/pull/100

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,2 +1,1 @@
-#imports_granularity = "Crate"
 edition = "2021"
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,2 +1,2 @@
-imports_granularity = "Crate"
+#imports_granularity = "Crate"
 edition = "2021"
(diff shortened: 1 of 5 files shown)


autodidaddict:
Will review this when the relative path dependency is gone

stevelr:
Build checks won't pass until  https://github.com/wasmCloud/weld/pull/100 is merged. Then, we can remove the relative path and re-trigger the build checks.

autodidaddict (dismissed):

brooksmtownsend (dismissed):
Good to go once weld-codegen is released

.rustfmt.toml
```diff
@@ -1,2 +1,2 @@
-imports_granularity = "Crate"
+#imports_granularity = "Crate"
```
brooksmtownsend: Let's just remove this if we don't want it

brooksmtownsend (changes requested):
only a little nit

.rustfmt.toml
stevelr: Deleted. 
(I like it, but on rust stable it isn't supported and unrecognized options cause a hard fail)

brooksmtownsend (approved):

PR #276, opened 2022-05-25
update dependencies, fix clippy warnings, bump to 0.11.0

stevelr:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -330,7 +330,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b04815249b103828116ca940ae26a198a79400e6f1805c19fe250078d0472fd4"
 dependencies = [
  "base64 0.10.1",
- "uuid",
+ "uuid 0.8.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


autodidaddict (dismissed):

brooksmtownsend (dismissed):
LGTM just needs a little Cargo.toml merging

thomastaylor312:
Unless we are in a rush to finalize 0.11, I'd recommend we wait until the bugfix release of oci-distribution is released: https://github.com/krustlet/oci-distribution/pull/37. This would make it so we don't have to cut another release

brooksmtownsend:
> Unless we are in a rush to finalize 0.11, I'd recommend we wait until the bugfix release of oci-distribution is released: [krustlet/oci-distribution#37](https://github.com/krustlet/oci-distribution/pull/37). This would make it so we don't have to cut another release

@thomastaylor312 BTW, this PR https://github.com/wasmCloud/wash/pull/269 already resolved what that bugfix ends up fixing, it'll be a slight internal refactor on our side but won't require a semver bump (edit: except for configurable max chunk limits, which if we remove the flag then it'll just be a simple change)

brooksmtownsend (approved):

PR #285, opened 2022-06-10
Initial implementation of experimental wadm client

autodidaddict:
Signed-off-by: Kevin Hoffman <autodidaddict@users.noreply.github.com>

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1,8 +1,4 @@
-use std::{
-    collections::HashMap,
-    path::{Path, PathBuf},
-    time::Duration,
-};
+use std::{collections::HashMap, path::PathBuf, time::Duration};
 
 use crate::{
     appearance::spinner::Spinner,
@@ -171,11 +167,7 @@ pub(crate) async fn handle_command(
 async fn undeploy_model(cmd: UndeployCommand) -> Result<bool> {
     let res = json_request(cmd.opts, &["undeploy", &cmd.model_name], json!({})).await?;
 
-    if res.is_some() {
-        Ok(true)
-    } else {
-        Ok(false)
-    }
+    Ok(res.is_some())
 }
 
 async fn deploy_model(cmd: DeployCommand) -> Result<bool> {
@@ -267,7 +259,7 @@ fn list_models_output(results: Vec<ModelSummary>) -> CommandOutput {
     CommandOutput::new(output::list_models_table(results), map)
 }
 
-fn show_model_output(raw: String, vetted: String, md: ModelDetails) -> CommandOutput {
+fn show_model_output(raw: PathBuf, vetted: PathBuf, md: ModelDetails) -> CommandOutput {
     let mut map = HashMap::new();
     map.insert("model".to_string(), json!(md));
     CommandOutput::new(output::show_model_details(raw, vetted), map)
@@ -329,19 +321,18 @@ fn show_model_history(results: Vec<ModelRevision>) -> CommandOutput {
     CommandOutput::new(output::list_revisions_table(results), map)
 }
 
-fn write_model(model: ModelDetails) -> Result<(String, String)> {
+fn write_model(model: ModelDetails) -> Result<(PathBuf, PathBuf)> {
     let name = model.vetted["name"].as_str().unwrap_or("");
     let version = model.vetted["version"].as_str().unwrap_or("");
     let json_filename = format!("{}_v{}.json", name, version);
     let raw_filename = format!("{}_v{}.txt", name, version);
 
-    let json_path = Path::new(&json_filename);
-    let raw_path = Path::new(&raw_filename);
+    let json_buf = PathBuf::from(json_filename);
+    let raw_buf = PathBuf::from(raw_filename);
+    let _ = std::fs::write(&json_buf, serde_json::to_vec(&model.vetted).unwrap());
+    let _ = std::fs::write(&raw_buf, model.raw);
 
-    let _ = std::fs::write(json_path, serde_json::to_vec(&model.vetted).unwrap());
-    let _ = std::fs::write(raw_path, model.raw);
-
-    Ok((raw_filename, json_filename))
+    Ok((raw_buf, json_buf))
 }
 
 async fn nats_client_from_opts(opts: ConnectionOpts) -> Result<(Connection, Duration)> {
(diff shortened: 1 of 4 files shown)


autodidaddict:
Note: normally I'd put in unit tests that verify that the command line goes to the right place, but since this is so early and experimental, I figure the commands will be changing drastically enough that it made sense to hold off on that until the interface stabilizes a bit.

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+    version: String,
+
+    #[clap(flatten)]
+    opts: ConnectionOpts,
+}
+
+#[derive(Args, Debug, Clone)]
+pub(crate) struct PutCommand {
+    /// Input filename (JSON or YAML) containing app specification
+    source: PathBuf,
```
connorsmith256, line 88: Does this need a `clap` attribute?

connorsmith256 (changes requested):
Looks good, with a couple suggestions for `mod.rs`:
- group helper functions in order, preferably following the order in the `handle_command` match block. I had to jump around a bit
- there's a lot of repeated code in setting up for a call to `json_request`. Could `json_request` take a `ConnectionOpts` and extract that info itself, similar to how we construct a client in `call`/`ctl`?

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+
+pub(crate) async fn handle_command(
+    command: AppCliCommand,
+    output_kind: OutputKind,
+) -> Result<CommandOutput> {
+    use AppCliCommand::*;
+    let sp: Spinner = Spinner::new(&output_kind);
+    let out: CommandOutput = match command {
+        List(cmd) => {
+            sp.update_spinner_message(" Querying app spec list ...".to_string());
```
thomastaylor312: Nit: Unintentional extra space at the beginning?

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+        }
+    };
+    sp.finish_and_clear();
+
+    Ok(out)
+}
+
+async fn undeploy_model(cmd: UndeployCommand) -> Result<bool> {
+    let nc = nats_client_from_opts(cmd.opts.clone()).await?;
+    let timeout = Duration::from_millis(cmd.opts.ack_timeout_ms);
```
thomastaylor312: Nit: I see this and the prefix unwrap in almost every function. Perhaps you could have the nats client function return a tuple with the client and timeout

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+    let prefix = cmd.opts.lattice_prefix.unwrap_or("default".to_string());
+    let topic = format!("wadm.api.{}.model.undeploy.{}", prefix, cmd.model_name);
```
thomastaylor312: Nit: Consider creating a `generate_topic` function that just takes the prefix, noun, and model name to avoid similar copy/pastes

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+
+    let res = json_request(&nc, &topic, json!({}), timeout).await?;
+
+    if res.is_some() {
```
thomastaylor312: Can be shortened to `Ok(res.is_some())`.

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+    CommandOutput::new(output::list_revisions_table(results), map)
+}
+
+fn write_model(model: ModelDetails) -> Result<(String, String)> {
+    let name = model.vetted["name"].as_str().unwrap_or("");
+    let version = model.vetted["version"].as_str().unwrap_or("");
+    let json_filename = format!("{}_v{}.json", name, version);
+    let raw_filename = format!("{}_v{}.txt", name, version);
+
+    let json_path = Path::new(&json_filename);
```
thomastaylor312: These are unnecessary as `String` implements `AsRef<Path>`

thomastaylor312: Super nit: You could create these as `PathBuf`s and then return them as you are representing a filepath (rather than just a string). Basically it would more clearly express intent

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+            .unwrap_or_else(|| DEFAULT_NATS_HOST.to_string())
+    });
+
+    let ctl_port = opts.ctl_port.unwrap_or_else(|| {
+        ctx.as_ref()
+            .map(|c| c.ctl_port.to_string())
+            .unwrap_or_else(|| DEFAULT_NATS_PORT.to_string())
+    });
+
+    let ctl_jwt = if opts.ctl_jwt.is_some() {
```
thomastaylor312, line 360: All of these blocks could be written more cleanly like this:

```rust
let ctl_jwt = opts.ctl_jwt.unwrap_or_else(|| ctx.as_ref().map(|c| c.ctl_jwt.clone()).unwrap_or_default())
```

There might be a way to use `and_then` as well, but you get the point

src/app/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+    topic: &str,
+    req: &[u8],
+    timeout: Duration,
+) -> Result<Option<serde_json::Value>> {
+    let res = nc.request_timeout(topic, req, timeout).await?;
+    let env: WadmEnvelope = serde_json::from_slice(&res.data)?;
+    if env.result == "success" {
+        Ok(env.data)
+    } else {
+        bail!("{}", env.message.unwrap_or("".to_string()))
```
thomastaylor312: Should be `unwrap_or_else`

thomastaylor312 (dismissed):
All nits since this is an initial implementation

src/app/mod.rs
autodidaddict: no. by leaving it off it defaults to a positional argument called `source`

src/app/mod.rs
autodidaddict: going to save this for another time as when I tried to implement this, the end result ended up looking more bloated than my actual code because functions later on require an `Option<String>` and the new ctl_jwt is a string.

brooksmtownsend (approved):
This looks good for experimental 👍🏻

autodidaddict:
@connorsmith256 the latest version of this PR contains changes per your suggestion. Review blocked on your approval so where should I send the bribes?

connorsmith256 (approved):

connorsmith256:
Damn, I should have read the comments before reviewing again. Missed opportunity for a jet ski.

Issue #286, opened 2022-06-23
[CHORE] Mark releases as not prerelease


brooksmtownsend:
If we do this then the latest release shows up in the sidebar which is better for visibility

brooksmtownsend:
The consensus from #290 is that we _shouldn't_ do this automatically without noting that this is not "production ready" software so closing for now

PR #288, opened 2022-07-15
Refactor to use async-nats

### LINKED ISSUE #283: [BUG] Can't connect to `connect.ngs.global` using wash ctl
**Describe the bug**
When attempting to use an NGS credsfile to connect to connect.ngs.global, I get the following warning and error:
```
[2022-06-08T11:41:26Z WARN  rustls::conn] Sending fatal alert BadCertificate

presented server name type wasn't supported
```
**To Reproduce**
1. Get a set of credentials for `connect.ngs.global`
1. Run `wash ctl get hosts -r connect.ngs.global --ctl-credsfile <path/to/credsfile>`
1. See error

**Expected behavior**
I would expect, if given valid credentials, that we could connect through NGS just fine

**Additional context**
I've also been able to reproduce this error with a combination of `--ctl-seed` and `--ctl-jwt`

### LINKED ISSUE #284: [BUG] Occasional hang on `ctl` commands
**Describe the bug**
On `v0.11.0`, I've found that occasionally my commands will go far beyond the timeout value for ctl commands. For example:
```
 on ☁️  dev (us-east-2) took 2s  ~/demo/leafnode ➜
wash ctl start actor wasmcloud.azurecr.io/clinicapi:0.2.3 --host-id NBWURVUSDQ7I235EKRAEYSHRYCITOQBOCA573SWJODAHFWRAXUGYL4WH
>>> ⢋⠀  Starting actor wasmcloud.azurecr.io/clinicapi:0.2.3 ...
^C

 on ☁️  dev (us-east-2) took 55s  ~/demo/leafnode ➜
wash ctl start actor wasmcloud.azurecr.io/clinicapi:0.2.3 --host-id NBWURVUSDQ7I235EKRAEYSHRYCITOQBOCA573SWJODAHFWRAXUGYL4WH

Actor wasmcloud.azurecr.io/clinicapi:0.2.3 started on host NBWURVUSDQ7I235EKRAEYSHRYCITOQBOCA573SWJODAHFWRAXUGYL4WH
```

As you can see in the example above, the start actor command ran for 55 seconds before I CTRL+C'ed it. Our timeouts for starting an actor, when a host ID is supplied, is around 3000ms so it shoudn't take this long to time out.

More information is needed in terms of identifying the place this is timing out, it could potentially be on the initial NATS connect which might be fixed by moving to the `async-nats` crate

brooksmtownsend:
This PR fixes #283 
I ran a `ctl get hosts` 1000 times and could not reproduce #284, so I'm saying this fixes #284 since I could reproduce it consistently (@connorsmith256 would be curious if you can give this a try too)

This is a breaking change because:
- It removes the ability to authenticate with a NATS server with only an nkey seed
- I removed the `--ack-timeout-ms` flag in favor of a ubiquitous `--timeout-ms` for the control interface.
  - In the case where you don't provide a different timeout for starting actors/providers, I automatically increase the timeout if you don't add `--skip-wait` to avoid automatically "failing" to start a provider if it's your first download
  
This PR also fixes a bug where the default auction timeout was used for `wash ctl get hosts`, ignoring the specified timeout.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,8 +61,7 @@ wascap = "0.8.0"
 weld-codegen = "0.4.6"
 wasmcloud-control-interface = "0.16.2"
 wasmbus-rpc = "0.9"
-# wasmcloud-test-util = "0.4.1"
-wasmcloud-test-util = { git = "https://github.com/wasmcloud/wasmcloud-test", branch = "fix/async-nats-bug" }
+wasmcloud-test-util = "0.4.1"
 
 [dev-dependencies]
 reqwest = {version = "0.11", default-features = false, features = ["json", "rustls-tls"]}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,6 @@ tokio = {version = "1", features = ["full"]}
 toml = "0.5"
 walkdir = "2.3"
 which = "4.2.2"
-crossbeam-channel = "0.5.2"
 cloudevents-sdk = "0.4.0"
 
 wascap = "0.8.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.11.0"
+version = "0.12.0-alpha.1"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "command-line-utilities"]
 description = "wasmcloud Shell (wash) CLI tool"
@@ -15,6 +15,7 @@ maintenance = {status = "actively-developed"}
 
 [dependencies]
 anyhow = "1.0.51"
+async-nats = "0.17.0"
 atelier_core = "0.2"
 bytes = "1.0"
 cargo_atelier = "0.2"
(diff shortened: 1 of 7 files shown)


src/util.rs
```diff
@@ -227,35 +227,32 @@ pub(crate) async fn nats_client_from_opts(
...
-        anats::Options::with_jwt(
-            move || Ok(jwt_contents.clone()),
-            move |nonce| kp.sign(nonce).unwrap(),
-        )
-        .connect(&nats_url)
+        async_nats::ConnectOptions::with_jwt(jwt_contents, move |nonce| {
+            let key_pair = kp.clone();
+            async move { key_pair.sign(&nonce).map_err(async_nats::AuthError::new) }
+        })
+        .connect("localhost")
```
connorsmith256: Should this still be `&nats_url`?

connorsmith256 (commented):
Someone more familiar with this code should approve, but FWIW I can no longer reproduce https://github.com/wasmCloud/wash/issues/284 after compiling from this branch 👍

.github/workflows/release.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           release_name: Release ${{ github.ref }}
           token: ${{ secrets.GITHUB_TOKEN }}
           draft: false
-          prerelease: true
+          prerelease: false
```
ghost: Could this be in a separate PR? I know it fixes #286 but might look confusing when looking back through commit history

Cargo.toml
```diff
@@ -52,16 +53,15 @@ thiserror = "1.0"
...
-
-wasmcloud-test-util = "0.3.3"
-weld-codegen = "0.4.5"
 which = "4.2.2"
-crossbeam-channel = "0.5.2"
 cloudevents-sdk = "0.4.0"
 
+wascap = "0.8.0"
+weld-codegen = "0.4.6"
+wasmcloud-control-interface = "0.16.2"
```
ghost, line 61: While i'm here - could we set up Dependabot to handle creating version upgrade PRs for us?

ghost (dismissed):

src/util.rs
brooksmtownsend: Good catch! I think this was a documentation copy fail (cursed jwt nonce closure async blah)

Cargo.toml
brooksmtownsend: I wouldn't mind doing that, just in a separate PR 😄 I'll file an issue real quick to track that

ghost: Yep sounds good!

.github/workflows/release.yml
brooksmtownsend: Good call 👍🏻

brooksmtownsend: Dropped that commit

ghost (approved):

PR #290, opened 2022-07-18
Removed prerelease from github action

### LINKED ISSUE #286: [CHORE] Mark releases as not prerelease

brooksmtownsend:
Fixes #286 

This just enables our releases to show in the github sidebar

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,7 +58,7 @@ jobs:
           release_name: Release ${{ github.ref }}
           token: ${{ secrets.GITHUB_TOKEN }}
           draft: false
-          prerelease: true
+          prerelease: false
 
   crates:
     needs: release


ghost (approved):

thomastaylor312 (changes requested):
Sorry I missed the original issue, but this was done on purpose. I know it is annoying it doesn't show up automatically in the sidebar, but this isn't 1.0 software, so we should be marking it as pre-release given how often we may still make breaking changes. If we all agree to change this, I am ok with that, but generally I like making it clear that this is still in a pre-1.0 state

thomastaylor312:
To clarify, this is something we should bring up with GH (I think there is already a feature request). The other option is to find a way to explicitly call out that this is pre-1.0 software. If we do that, then I am ok with this moving forward

ghost:
> To clarify, this is something we should bring up with GH (I think there is already a feature request). The other option is to find a way to explicitly call out that this is pre-1.0 software. If we do that, then I am ok with this moving forward

Would something similar to this [Bevy warning](https://github.com/bevyengine/bevy#warning) in their readme suffice? Perhaps could just put it in each release notes instead of in the readme if it's a bit too scary

> **WARNING**

> Bevy is still in the very early stages of development. APIs can and will change (now is the time to make suggestions!). Important features are missing. Documentation is sparse. Please don't build any serious projects in Bevy unless you are prepared to be broken by API changes constantly.

thomastaylor312:
I think release notes is probably the best place for it. Wash is definitely stable enough to use for realsies, but we still reserve the right to break things as we evolve

Issue #264, opened 2022-05-11
[FEATURE] `wash install`


brooksmtownsend:
**Is your feature request related to a problem? Please describe.**
Installing the wasmCloud host isn't necessarily difficult, but could be much easier.

**Describe the solution you'd like**
In tandem with https://github.com/wasmCloud/wasmcloud-otp/issues/395, it would be great for our initial user install to be able to download `wash`, run `wash install`, and automatically have the NATS and wasmCloud binaries downloaded to something like `~/.wash/bin`. This is a familiar install process to other tools, and makes the initial setup of a wasmCloud developer environment easy and simple.

Installing NATS could certainly be an optional flag, as that's not always the desired action for the user. One thing I despise about `brew` for example is that `brew install wash` can also end up installing another Rust toolchain which can be confusing, and this process should not mess with a developer's environment by default.

ghost:
`brew install wash` can install another Rust toolchain? That's really odd, does brew build from source? I always thought it just served the binary.

Anyway, instead of `wash install`, should we have something like `rustup`? Which just installs everything you need for wasmCloud development (wash, nats, wasmcloud) all at once? `washup`? lol

Then we can do something similar to `rustup`'s website: https://rustup.rs/

brooksmtownsend:
> does brew build from source? I always thought it just served the binary.

Only if it can't find a binary, which we may not actually upload 🤔 But it installs the dependent packages anyway I believe

I'm fine with a rustup-y type script or like get.crit.sh, I'd only want to ensure that it's easy to maintain, my biggest concern is simply that we'd want some kind of `wash upgrade` or `wash install --upgrade` to update dependencies

PR #292, opened 2022-07-19
[FEATURE] Adding `wash-lib`, implementing `start` functionality

### LINKED ISSUE #266: [FEATURE] `wash up` / `wash start host`
Bring back `wash up`! A previous supported command in `0.18.0` that would launch a host in the integrated REPL. Though we've found that the REPL isn't the best experience for running a wasmCloud host, it would be great to have this command back again for a simple experience to start a host.

In tandem with #264, the first experience a developer would have with wasmcloud is
1. Look at documentation
2. Download `wash` for their respective OS/ARCH
3. Run `wash install`
4. Run `wash up`
5. Proceed with the example documentation

There are some edge cases to look out for, namely that we are going to either require that NATS is installed or accessible by the host.

brooksmtownsend:
Implements the functionality behind #266 and #264 

This PR implements the functionality in the new `wash-lib` crate to download and start NATS and the wasmCloud host. This PR will only contribute the library functionality, tests, doctests, etc and will not actually add the functionality in the wash CLI. The eventual goal of the `wash-lib` crate is to contain all of the library functionality that `wash` uses so that it can be re-used in other Rust projects in the wasmCloud ecosystem. We'll likely gate modules behind feature flags so that consumers of the crate can pull in only the functionality they need.

The ideal usage, from the CLI point of view, of the `wash start` command is basically that a developer using `wash` can simply run `wash start` and not worry about downloading NATS or wasmCloud, and once the command completes a local wasmCloud host will be running on `http://localhost:4000` alongside a NATS server. Flags will be provided to allow users to connect to existing NATS clusters, specify parameters to a wasmCloud host, etc.

Starting multiple wasmCloud hosts on the same machine is complicated, so for now `wash start` will only support launching a single host. This can be added later, it just requires manually modifying the erlang VM args. I've included a unit test to verify this behavior.


brooksmtownsend:
Had to deal with a few last-second failures on windows (`.\bin\wasmcloud_host ping` always returns a zero exit code so I had to check for the `pong` existence) and pinging without running the script first always times out, but those should be fixed now

autodidaddict (dismissed):
:clap:

.github/workflows/washlib.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  push:
+    tags:
+      - "wash-lib-v*" # Push events to matching v*, i.e. v1.0, v20.15.10
+
+env:
+  working-directory: ./crates/wash-lib
+jobs:
+  cargo_check:
+    runs-on: ubuntu-latest
```
thomastaylor312: We should use the rust setup thing from actions-rs to make sure we have the latest version and all necessary components

Makefile
```diff
@@ -6,15 +6,16 @@
...
 
 test-integration: ##Run integration test suite
 	docker compose -f ./tools/docker-compose.yml up --detach
 	cargo test --no-fail-fast --test "integration*" -- --nocapture
 	docker compose -f ./tools/docker-compose.yml down
 
 rust-check:
-	rustfmt --edition 2021 --check src/*.rs src/*/*.rs 
-	cargo clippy --all-features --all-targets
+	rustfmt --edition 2021 --check src/*.rs src/*/*.rs crates/*/**/*.rs
```
thomastaylor312: Nit: I think we should prefer cargo fmt rather than rustfmt. It is the same thing, but I rarely see other projects use it directly

crates/wash-lib/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+maintenance = {status = "actively-developed"}
+
+[features]
+default = ["start"]
+start = []
+
+[dependencies]
+anyhow = "1.0.51"
+futures = "0.3"
+tokio = {version = "1", features = ["full"]}
```
thomastaylor312: So we probably shouldn’t do full here. Because this is a library, We’re not actually starting the tokio  run time. So we should only pull in the absolutely necessary and anyone consuming the crate can enable the full tokio feature in their main cargo file

crates/wash-lib/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+//! A crate that implements the functionality behind the wasmCloud shell
+//!
+//! The `wash` command line interface <https://github.com/wasmcloud/wash> is a great place
+//! to find examples on how to fully utilize this library.
+
+#[cfg(feature = "start")]
```
thomastaylor312, line 6: Any reason to put this behind a feature flag?

crates/wash-lib/src/start/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+//!     let install_dir = PathBuf::from("/tmp");
+//!     let os = std::env::consts::OS;
+//!     let arch = std::env::consts::ARCH;
+//!
+//!     // Download NATS if not already installed
+//!     let nats_binary = ensure_nats_server(os, arch, "v2.8.4", &install_dir).await?;
+//!
+//!     // Start NATS server, redirecting output to a log file
+//!     let nats_log_path = install_dir.join("nats.log");
+//!     let nats_log_file = std::fs::File::create(&nats_log_path)?;
```
thomastaylor312: This should be using the tokio file type since it is in an async function

crates/wash-lib/src/start/nats.rs
```diff
@@ -0,0 +1,265 @@
...
+    ))
+}
+
+/// Helper function to execute a NATS server binary with wasmCloud arguments
+/// # Arguments
+///
+/// * `bin_path` - Path to the nats-server binary to execute
+/// * `stderr` - Specify where NATS stderr logs should be written to. If logs aren't important, use std::process::Stdio::null()
+/// * `port` - Port for NATS to listen on
+pub fn start_nats_server<P, T>(bin_path: P, stderr: T, port: u16) -> Result<Child>
```
thomastaylor312: I know that we don’t technically need this to be async, but it will almost certainly be used in an async context. The tcp connection is blocking IO and could deadlock the async stuff so this function and the stuff in it should probably use the async stuff

crates/wash-lib/src/start/nats.rs
```diff
@@ -0,0 +1,265 @@
...
+        .map_err(|e| anyhow!(e))
+}
+
+/// Helper function to ensure the NATS server binary is successfully
+/// installed in a directory
+pub async fn is_nats_installed<P>(dir: P) -> bool
+where
+    P: AsRef<Path>,
+{
+    metadata(dir.as_ref().join(NATS_SERVER_BINARY))
```
thomastaylor312, line 166: Nit: you may want to check that it is actually a file as well

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -0,0 +1,363 @@
...
+/// # async fn main() {
+/// use wash_lib::start::ensure_wasmcloud;
+/// let os = std::env::consts::OS;
+/// let arch = std::env::consts::ARCH;
+/// let res = ensure_wasmcloud(os, arch, "v0.55.1", "/tmp/wasmcloud/").await;
+/// assert!(res.is_ok());
+/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/bin/wasmcloud_host".to_string());
+/// # }
+/// ```
+pub async fn ensure_wasmcloud<P>(os: &str, arch: &str, version: &str, dir: P) -> Result<PathBuf>
```
thomastaylor312: Super Nit: the most common usage here will likely be the os and arch it is running on. What if for both here and in nats we made these parameters optional and defaulted them to the os/arch combo it is running on? Actually, an even better idea would be to have this be the “with specific os/arch” function and have another one without the os/arch that defaults to the current os/arch of the system

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -0,0 +1,363 @@
...
+        .map_err(|e| anyhow!(e))
+}
+
+/// Helper function to ensure the wasmCloud host tarball is successfully
+/// installed in a directory
+pub async fn is_wasmcloud_installed<P>(dir: P) -> bool
+where
+    P: AsRef<Path>,
+{
+    use futures::future::join_all;
```
thomastaylor312: Please put imports in the top section and not in functions pretty please

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -0,0 +1,363 @@
...
+        (true, Some(path)) => Ok(path),
+        (true, None) => Err(anyhow!(
+            "wasmCloud was installed but the binary could not be located"
+        )),
+        (false, _) => Err(anyhow!(
+            "wasmCloud was not installed successfully, please see logs"
+        )),
+    }
+}
+
```
thomastaylor312, line 144: Please move up top

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -0,0 +1,363 @@
...
+use std::collections::HashMap;
+/// Helper function to start a wasmCloud host given the path to the elixir release script
+/// /// # Arguments
+///
+/// * `bin_path` - Path to the wasmcloud_host script to execute
+/// * `stdout` - Specify where wasmCloud stdout logs should be written to. Logs can be written to stdout by the erlang process
+/// * `stderr` - Specify where wasmCloud stderr logs should be written to. Logs are written to stderr that are generated by wasmCloud
+/// * `env_vars` - Environment variables to pass to the host, see <https://wasmcloud.dev/reference/host-runtime/host_configure/#supported-configuration-variables> for details
+pub fn start_wasmcloud_host<P, T, S>(
```
thomastaylor312: Same comment here as the function in nats. Should probably be async

thomastaylor312 (changes requested):
This is excellent! Very flexible and ergonomic. Just a few minor changes with some blocking things on async functions

crates/wash-lib/src/lib.rs
brooksmtownsend: The idea was that each "module" of the library could be behind a feature flag, all of them are enabled by default but this would let consumers of the library choose which parts to pull in. If this is excessive optimization then I can remove that part 😄

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: I like that second idea, it would be an uncommon use case to download for a different OS and ARCH for sure and I'd be happy to have this function use the OS/ARCH of the host, then have another helper that uses this "with specific os/arch" logic

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: Only because you asked nicely

crates/wash-lib/src/lib.rs
autodidaddict: I'd assumed that this let ad hoc consumers choose which pieces of wash-lib they wanted, whereas the official `wash` binary would grab all the thingz

thomastaylor312:
Oh also, I don’t think is is breaking current wash functionality is it? Just wanted to confirm we actually needed the breaking label

crates/wash-lib/src/lib.rs
autodidaddict: I actually like this plan of categorizing the lib by feature

thomastaylor312: Ok sweet. That is what I was thinking but wanted to confirm

brooksmtownsend:
> Oh also, I don’t think is is breaking current wash functionality is it? Just wanted to confirm we actually needed the breaking label

Oh very true! I'll remove that since this is in the lib

Makefile
brooksmtownsend: Oh yeah, good call

crates/wash-lib/src/start/nats.rs
brooksmtownsend: Since downloading the NATS server / wasmCloud is an async context, I think that makes sense

crates/wash-lib/src/start/mod.rs
brooksmtownsend: I _can_ use the tokio file type, but then I have to pass in something that implements `Into<Stdio>` for the output redirecting. So this would have to look like
```rust
let nats_log_file = tokio::fs::File::create(&nats_log_path).await?.into_std();
```

thomastaylor312: Yeah it isn’t ideal but I do know that we should try to make sure it is non-blocking. We can look if there is an async equivalent that we could use instead and do the std conversion in the function.

crates/wash-lib/src/start/nats.rs
brooksmtownsend: `map_or` worked great for this but I wanted to use https://doc.rust-lang.org/std/result/enum.Result.html#method.is_ok_and :(

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -0,0 +1,416 @@
...
+{
+    // wasmCloud host logs are sent to stderr as of https://github.com/wasmCloud/wasmcloud-otp/pull/418
+    let mut cmd = &mut Command::new(bin_path.as_ref());
+
+    // If we can connect to the local port, a wasmCloud host won't be able to listen on that port
+    let port = env_vars
+        .get("PORT")
+        .cloned()
+        .unwrap_or_else(|| "4000".to_string());
+    if std::net::TcpStream::connect(format!("127.0.0.1:{}", port)).is_ok() {
```
thomastaylor312: This should use the TCP stream from tokio

crates/wash-lib/src/start/nats.rs
```diff
@@ -0,0 +1,278 @@
...
+/// * `bin_path` - Path to the nats-server binary to execute
+/// * `stderr` - Specify where NATS stderr logs should be written to. If logs aren't important, use std::process::Stdio::null()
+/// * `port` - Port for NATS to listen on
+pub async fn start_nats_server<P, T>(bin_path: P, stderr: T, port: u16) -> Result<Child>
+where
+    P: AsRef<Path>,
+    T: Into<Stdio>,
+{
+    // If we can connect to the local port, NATS won't be able to listen on that port
+    if std::net::TcpStream::connect(format!("127.0.0.1:{}", port)).is_ok() {
```
thomastaylor312: You'll need to use the tokio one here as you are async

crates/wash-lib/src/start/wasmcloud.rs~
```diff
@@ -0,0 +1,399 @@
+use anyhow::{anyhow, Result};
```
thomastaylor312: Looks like this file is a possible `cp` failure/artifact?

thomastaylor312 (changes requested):
Just a few more teensy changes

crates/wash-lib/src/start/wasmcloud.rs~
brooksmtownsend: yep, editing in windows :(

thomastaylor312 (approved):

.github/workflows/rust.yml
```diff
@@ -14,27 +14,19 @@ jobs:
...
-      with:
-        test-options: ${{ matrix.test-options }}
+      - uses: actions/checkout@v2
+      - name: Run Rust check + clippy
+        run: make rust-check
+      - name: Run unit tests
+        run: make test
+      - name: Launch integration test services
+        if: ${{ startswith(matrix.os, 'ubuntu') }}
+        uses: sudo-bot/action-docker-compose@latest
```
williamdes, line 27: Thank you for using my action !
Feel free to ping me if there is an incident with it

PR #294, opened 2022-07-27
`wash up` implementation

### LINKED ISSUE #264: [FEATURE] `wash install`

### LINKED ISSUE #266: [FEATURE] `wash up` / `wash start host`

brooksmtownsend:
This PR implements the `wash start` functionality which fixes #264 and fixes #266

Developers will now be able to install wash and run `wash up` to get the following output (and of course a running host):
```
wash up --detached

🛁 wash up completed successfully
🕸 NATS is running in the background at http://127.0.0.1:4222
🌐 The wasmCloud dashboard is running at http://localhost:4000

🛑 To stop the wasmCloud host and the NATS server, run:
/Users/brooks/.wash/downloads/bin/wasmcloud_host stop; kill 98697
```

or in interactive mode:
```
wash up
🏃 Running in interactive mode, your host is running at http://localhost:4000
🚪 Press `CTRL+c` at any time to exit
17:46:47.251 [info] Wrote "./host_config.json"
17:46:47.251 [info] Wrote "/Users/brooks/.wash/host_config.json"
17:46:47.252 [info] Connecting to control interface NATS without authentication
17:46:47.252 [info] Connecting to lattice rpc NATS without authentication
17:46:47.253 [info] Host NDTX2AZYORL6HARPSXK3CMCA54ZYW2P7XECPMS5SDXD2BQU7SEDA4YC2 (cool-firefly-8971) started.
17:46:47.253 [info] Valid cluster signers: CDV43ERT2JMQSO2AFC3K2AQBLZ6M5353WYYK4LXPICN6QC6GBAJTWRN2
17:46:47.257 [info] Started wasmCloud OTP Host Runtime
17:46:47.259 [info] Running WasmcloudHostWeb.Endpoint with cowboy 2.9.0 at :::4000 (http)
17:46:47.259 [info] Access WasmcloudHostWeb.Endpoint at http://localhost:4000
17:46:47.358 [info] Lattice cache stream created or verified as existing (0 consumers).
17:46:47.358 [info] Attempting to create ephemeral consumer (cache loader)
17:46:47.359 [info] Created ephemeral consumer for lattice cache loader
```
Some design choices in this PR:
- ~"interactive" mode is an opt-in flag, I felt this was more ergonomic than `wash up --daemon`~ changed to `wash up` in the foreground by default, `wash up --detached` in the background to mirror docker-compose
- `--connect-only` is an option to only connect to a NATS server, rather than launching one, so wash can quickly exit instead of failing silently when NATS infrastructure already exists
- All defaults mirror https://wasmcloud.dev/reference/host-runtime/host_configure/
- If `--ctl-host`, `--ctl-port` and their `rpc` and `prov-rpc` commands are omitted, then by default I use the `nats-host` and `nats-port` arguments instead. This ends up simplifying a command to connect to NATS at `127.0.0.1:4223` from
```
wash up --nats-host 127.0.0.2 --ctl-host 127.0.0.2 --rpc-host 127.0.0.2 --prov_rpc-host 127.0.0.2 --nats-port 4223 --ctl-port 4223  --rpc-port 4223 --prov-rpc-port 4223
```
to
```
wash up --nats-host 127.0.0.2 --nats-port 4223
```
- In `wash-lib` we always write a nats conf and use that for configuration to support jetstream domains. I considered accepting a file to write this to, but felt the API got a little less ergonomic if I did so. The downside is that we'd fail if we can't write the file, but wash should be writing to `~/.wash/downloads` anyways.
- Running wasmCloud in the foreground allows me to stream logs and write them to `~/.wash/downloads/wasmcloud.log` when not in interactive mode rather than dealing with `var/log/erlang.log.*`
  - Worth noting here too that to run `./bin/wasmcloud_host start` on Windows, you first have to install it as a Service and then start it. It's not a big deal to do this, and it still makes capturing logs hard, but the "correct" way to run `start` is more difficult than running with `foreground`
- This has some unix and windows specific code because I wanted to make sure it worked for all platforms :)
- wasmCloud and NATS versions can be specified, but default to a pinned version rather than the latest for compatibility
- Instead of a `wash up --down` or `wash up --cleanup` or `wash drain --hosts` command, I decided to output the command to stop the host and NATS as a part of the output to let users stop the resources as necessary. In interactive mode, pressing CTRL+c cleans up running resources.


crates/wash-lib/src/start/nats.rs
```diff
@@ -125,36 +126,156 @@ where
...
+            r#"
+jetstream {{
+    domain={}
+}}
+{}
+"#,
+            self.js_domain.unwrap_or_else(|| "core".to_string()),
+            leafnode_section
+        );
+        write(path, config).await.map_err(|e| anyhow!(e))
```
thomastaylor312: Nit: Rather than invoking the macro, you might be able to get away with `map_err(anyhow::Error::from)`

thomastaylor312:
I'm halfway through a review, but since this is a large PR, it should have a second review from a maintainer

src/up/credsfile.rs
```diff
@@ -0,0 +1,49 @@
+//! A temporary module to parse NATS credsfiles and translate
+//! their contents into a JWT and Seed value
+use anyhow::{anyhow, Result};
+use regex::Regex;
+use std::path::Path;
```
thomastaylor312: Super Nit: Separate `std` imports so they are up top above the rest and a line break separating

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
+use crate::appearance::spinner::Spinner;
+use crate::cfg::cfg_dir;
+use crate::util::{CommandOutput, OutputKind};
+use anyhow::{anyhow, Result};
+use clap::Parser;
+use serde_json::json;
+use std::collections::HashMap;
```
thomastaylor312, line 1: Same nit with `std` imports here. Crate imports should also be separated

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+    // Capture listen address to keep the value after the nats_opts are moved
+    let nats_listen_address = format!("{}:{}", cmd.nats_opts.nats_host, cmd.nats_opts.nats_port);
+
+    // Avoid downloading + starting NATS if the user already runs their own server. Ignore connect_only
+    // if this server has a remote and credsfile as we have to start a leafnode in that scenario
+    let nats_opts = cmd.nats_opts.clone();
+    let mut nats_process = if !cmd.nats_opts.connect_only
+        || cmd.nats_opts.nats_remote_url.is_some() && cmd.nats_opts.nats_credsfile.is_some()
+    {
+        if cmd.nats_opts.connect_only {
```
thomastaylor312: Nit: `clap` has the ability to enforce mutual exclusivity. Definitely not required for this PR

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+        };
+        let spinner = Spinner::new(&output_kind);
+        spinner.update_spinner_message(
+            "CTRL+c received, gracefully stopping wasmCloud and NATS...".to_string(),
+        );
+
+        // Terminate wasmCloud and NATS processes
+        stop_wasmcloud(wasmcloud_executable.clone()).await?;
+
+        if let Some(mut process) = nats_process {
```
thomastaylor312: Nit: I _think_ you could get away with an `as_ref_mut` here rather than unwrapping and reassigning

thomastaylor312 (dismissed):
LGTM. Nothing but nits here

src/up/credsfile.rs
autodidaddict: Is there a configuration file somewhere we can write to make `cargo fmt` do this work for us?

ghost:
Starting off with design decision stuff, I'll read through the code more closely in a bit.

> If --ctl-host, --ctl-port and their rpc and prov-rpc commands are omitted, then by default I use the nats-host and nats-port arguments instead. This ends up simplifying a command to connect to NATS at 127.0.0.1:4223 from
> wash up --nats-host 127.0.0.2 --ctl-host 127.0.0.2 --rpc-host 127.0.0.2 --prov_rpc-host 127.0.0.2 --nats-port 4223 --ctl-port 4223  --rpc-port 4223 --prov-rpc-port 4223
>to
> wash up --nats-host 127.0.0.2 --nats-port 4223

Very nice! Like this a lot

> --connect-only is an option to only connect to a NATS server, rather than launching one, so wash can quickly exit instead of failing silently when NATS infrastructure already exists

Could this be `--nats_connect_only` (or similar) instead? The current version seems ambiguous to me.

> Instead of a wash up --down or wash up --cleanup or wash drain --hosts command, I decided to output the command to stop the host and NATS as a part of the output to let users stop the resources as necessary. In interactive mode, pressing CTRL+c cleans up running resources.

I think that a `wash down` command would be a better UX, but this is fine for now.

src/up/credsfile.rs
thomastaylor312: It will do it so long as you have a newline in between imports. The auto import function in rust-analyzer/vscode does this for you automatically

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+            "500",
+            "--rpc-tls",
+            "--structured-log-level",
+            "warn",
+            "--wasmcloud-js-domain",
+            "domain",
+            "--wasmcloud-version",
+            "v0.55.1",
+            "--lattice-prefix",
+            "anotherprefix",
```
ghost, line 585: Whew that's a lot of flags! 😅

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+                    Some(creds.to_owned()),
+                )
+                .await
+                {
+                    return Err(anyhow!("Could not connect to leafnode remote: {}", e));
+                } else {
+                    cmd.nats_opts
+                }
+            }
+            (_, _) => cmd.nats_opts,
```
ghost: nit: consider putting this check in it's own function for reusability & reducing the level of nesting here

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+                    "Could not connect to NATS at {}, exiting since --connect-only was set",
+                    nats_listen_address
+                )
+            })?
+    };
+
+    // Download wasmCloud if not already installed
+    spinner.update_spinner_message(" Downloading wasmCloud ...".to_string());
+    let wasmcloud_executable =
+        ensure_wasmcloud(&cmd.wasmcloud_opts.wasmcloud_version, &install_dir).await?;
```
ghost: `ensure_wasmcloud` doesn't really explain what this function does IMO. But I see the problem is this function is doing two things, checking if the executable exists and then downloading it if not.

I think we might be better off using the `is_wasmcloud_installed()` function from [wasmcloud.rs](https://github.com/wasmCloud/wash/blob/main/crates/wash-lib/src/start/wasmcloud.rs#L238-L254) directly, and then calling a `download_wasmcloud()` function if it returns false. Some users might also prefer to fail out if the binary doesn't exist rather than downloading one right away (or be prompted before the download begins), which is a use case we can't support with a `ensure` function.

ghost (commented):
Looks very good! Left a few comments. The main thing missing IMO is an integration test, although I can see how that might be hard to set up reliably. What do you think?

src/up/mod.rs
brooksmtownsend: I think this logic makes sense, and I could refactor the logic from `ensure_wasmcloud` to a `download_wasmcloud` function so that crate consumers can use this in the way they choose. I can add a `--wasmcloud-start-only` flag as well so that we can quit out, similar to how we quit out early if we can't connect to NATS, but I'd prefer to leave the interactive prompt out as it just means a longer process to get to the running host

thomastaylor312: I think it is ok to have a top level `ensure_wasmcloud`. The `ensure_*` type of functions are common across various projects I have worked on. However, we should have it be a helper function that calls the publicly exported `is_wasmcloud_installed` and `download_wasmcloud`. It makes the shorthand default real easy, but it also gives flexibility for those who wish to use it differently

thomastaylor312 (dismissed):
Good to go once tests pass

src/up/mod.rs
brooksmtownsend: edit: After trying this for a bit, I actually don't like the way install + download / ensure feels. I think the most common use case is just to immediately download if not installed which is what ensure does. Currently the API would allow any consumers of the crate to check if wasmCloud is installed and appropriately handle that if it's not installed.

I added the flag to avoid downloading since I could see it being useful, and implemented the behavior you mentioned

Edit edit: missed Taylor's comment, will revisit after community call

ghost: Ok, sounds good!

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+
+    #[clap(flatten)]
+    pub(crate) wasmcloud_opts: WasmcloudOpts,
+}
+
+#[derive(Parser, Debug, Clone)]
+pub(crate) struct NatsOpts {
+    /// If a connection can't be established, exit and don't start a NATS server. Will be ignored if a remote_url and credsfile are specified
+    #[clap(long = "connect-only", env = "NATS_CONNECT_ONLY")]
+    pub(crate) connect_only: bool,
```
ghost, line 49: nit: rename to `nats_connect_only`? To make the meaning clearer when using the flag, `--connect-only` seems like it could be for wasmcloud or nats

ghost (dismissed):
LGTM, just one nit. Also the nit from the previous review on making the 'testing connection to nats' part it's own function.

src/up/mod.rs
ghost: Looking at the newer commits it looks like you did do this? But then in the files changed it still has the older version. Not sure what's happening with that, feel free to close this if I'm looking at the wrong thing

src/up/mod.rs
brooksmtownsend: I ended up adding this, so both NATS and wasmCloud have `is_installed`, `download`, and `ensure` functions!

src/up/mod.rs
brooksmtownsend: Weirdly I see what you mean, but it is done so 🤞🏻  and I'll resolve this

thomastaylor312 (dismissed):

src/up/mod.rs
```diff
@@ -0,0 +1,651 @@
...
+                )
+                .await
+                {
+                    return Err(anyhow!("Could not connect to leafnode remote: {}", e));
+                } else {
+                    cmd.nats_opts
+                }
+            }
+            (_, _) => cmd.nats_opts,
+        };
```
ghost: Did you want to put this in it's own function? It's not necessary but could help with the nesting here. Feel free to dismiss this if you don't want to though

ghost (dismissed):

src/up/mod.rs
brooksmtownsend: I was thinking that abstracting it into the `start_nats` function was good to keep it separate, and that it wasn't really necessary to abstract further.

ghost: I think this is another case of the PR not updating the file? Weird.

ghost (approved):

Issue #282, opened 2022-06-01
[RFC] wasmCloud.toml


ghost:
## Is your feature request related to a problem? Please describe.
Currently the wasmCloud development process uses Makefiles to compile and sign actors. This is split into a "compiler" makefile ([actor.mk](https://github.com/wasmCloud/examples/blob/main/build/makefiles/actor.mk))  the project's `Makefile` ([example](https://github.com/wasmCloud/examples/blob/main/actor/echo/Makefile)), which is basically a config file in a trenchcoat.

## Describe the solution you'd like
Create a wasmCloud specific configuration file,`wasmcloud.toml` to replace the current project Makefiles. This should be usable for actors, interfaces, and providers. This file will be consumed by `wash` to compile and sign the actor using `wash build` (see #263).

### Remaining with current Makefile setup

Makefiles are tricky for Windows developers, and are not really intended as configuration files. The developer experience is not ideal with the current setup.

### Using the language's configuration file (`Cargo.toml`, `go.mod`, etc)

This carries some advantages in terms of DX, as we're reducing the number of configuration files a developer needs to manage. However this could become a constraint as we add more language support to wasmCloud, if those languages don't support the features we need in their configuration file.

However, we could allow parameters to be optional in the `wasmcloud.toml` file, and attempt to parse them out of the language's file if they are missing. This would reduce duplication. For example, the actor's name and version number could be stored only in the language file and parsed out, as opposed to being stored in both files and the developer needing to update them in two places on each change.

## Additional context
Issue #263 (`wash build`) has some more context on this, and these two features are pretty intertwined.

ghost:
Here's my initial stab at the fields for this file, using the Makefiles in the [examples repo](https://github.com/wasmCloud/examples) as a reference. This WIP, I'm just throwing everything into a comment and will edit it later.

For all configurations:
- language: enum "rust" | "tinygo", more to be added as we expand.
- type: enum "actor" | "provider" | "interface"
- name: optional string, if not provided will attempt to infer from language's config file (`Cargo.toml` etc)
- version: optional string, should be semver. Same as name, will attempt to infer from language's config file if not provided.

For actors:
- claims: optional array of strings.
- registry: optional string url to the registry, if not provided will ask for it on each command that it is needed.
- push_insecure: optional boolean, default false.
- revision: required int. However I'd like to explore getting rid of this if possible, could someone shine light on why we need it instead of just using the version number?
- actor_name: optional string, but can't we just use the name field?
[...]

PR #297, opened 2022-08-03
Create `wash build` command and add configuration parsing

ghost:
This PR adds a configuration parsing functionality to `wash-lib` which can parse `wasmcloud.toml` files according to the spec laid out in https://github.com/wasmCloud/wash/issues/282. It also introduces a `wash build` command which can build an actor, provider or interface using the `wasmcloud.toml` file, https://github.com/wasmCloud/wash/issues/263.

This only contains the implementation of `wash build` for Rust and TinyGo actors. Providers and Interfaces will be implemented in a subsequent PR to keep the amount of changes here (somewhat) manageable.


thomastaylor312:
Oh man I’m excited for this. Will take a look soon

ghost:
Gonna focus on getting a version of this that works for building & signing Rust and TinyGo actors and then ready to submit. Pushing to the registry as well as building other types (providers, interfaces) can be implemented later

ghost:
Ready for review! @brooksmtownsend @thomastaylor312 @autodidaddict

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    /// This is renamed to "type" but is named project_type here to avoid clashing with the type keyword in Rust.
+    #[serde(rename = "type")]
+    pub project_type: TypeConfig,
+    /// Configuration common amoung all project types & languages.
+    pub common: CommonConfig,
+}
+
+#[derive(serde::Deserialize, Debug, PartialEq, Clone, Default)]
+pub struct ActorConfig {
+    /// The list of provider claims that this actor requires. eg. ["wasmcloud:httpserver"]
```
brooksmtownsend: nit: would be good to show 2 contracts here just to show it's a comma separated list and not a string

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub claims: Vec<String>,
+    /// The registry to push to. eg. "localhost:8080"
+    pub registry: Option<String>,
+    /// Whether to push to the registry insecurely. Defaults to false.
+    pub push_insecure: bool,
+    /// The directory to store the private keys in.
```
brooksmtownsend: nit: `private signing keys`

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub key_directory: PathBuf,
+    /// The filename of the signed wasm actor.
+    pub filename: Option<String>,
+    /// The target wasm target to build for.
+    pub wasm_target: String,
```
brooksmtownsend: This would be worth noting that it defaults to wasm32-unknown-unknown

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+            codegen_config: raw_config
+                .codegen_config
+                .unwrap_or_else(|| PathBuf::from("./codegen.toml")),
+        })
+    }
+}
+
+#[derive(serde::Deserialize, Debug, PartialEq, Clone, Default)]
+pub struct RustConfig {
+    /// The path to the cargo binary. Optional, will default to the default `cargo` if not specified.
```
brooksmtownsend: nit: this might read better as "will default to search the user's PATH for `cargo` if not specified"

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    pub cargo_path: Option<PathBuf>,
+    /// Path to cargo/rust's `target` directory. Optional, defaults to `./target`.
+    pub target_path: Option<PathBuf>,
+}
+#[derive(serde::Deserialize, Debug, PartialEq, Default, Clone)]
+
+struct RawRustConfig {
+    /// The path to the cargo binary. Optional, will default to the default `cargo` if not specified.
+    pub cargo_path: Option<PathBuf>,
```
brooksmtownsend: Same comment re: `cargo` as above

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    raw_project_config
+        .try_into()
+        .map_err(|e: anyhow::Error| anyhow!("{} in {}", e, path.display()))
+}
+
+impl TryFrom<RawProjectConfig> for ProjectConfig {
+    type Error = anyhow::Error;
+
+    fn try_from(raw_project_config: RawProjectConfig) -> Result<Self> {
+        let project_type_config = match raw_project_config.project_type.as_str() {
```
brooksmtownsend: It may be worth converting the project type to lowercase just in case the type is specified as `Actor`. 

I'm fine with pushback here I just wouldn't want the project to fail to parse if the above case happened

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+
+            _ => {
+                return Err(anyhow!(
+                    "Unknown project type: {}",
+                    raw_project_config.project_type
+                ));
+            }
+        };
+
+        let language_config = match raw_project_config.language.as_str() {
```
brooksmtownsend: same comment about lowercasing

src/build.rs
```diff
@@ -0,0 +1,254 @@
...
+fn build_provider(
+    _command: BuildCommand,
+    _output_kind: OutputKind,
+    _provider_config: ProviderConfig,
+    _language_config: LanguageConfig,
+    _common_config: CommonConfig,
+) -> Result<CommandOutput> {
+    Ok(CommandOutput::from_key_and_text(
+        "result",
+        "wash build has not be implemented for providers yet. Please use the Makefiles for now!"
```
brooksmtownsend: I would suggest that the user just run `make` to build instead, since that's a simple action item

src/build.rs
```diff
@@ -0,0 +1,254 @@
...
+fn build_interface(
+    _command: BuildCommand,
+    _output_kind: OutputKind,
+    _interface_config: InterfaceConfig,
+    _language_config: LanguageConfig,
+    _common_config: CommonConfig,
+) -> Result<CommandOutput> {
+    Ok(CommandOutput::from_key_and_text(
+        "result",
+        "wash build has not be implemented for interfaces yet. Please use the Makefiles for now!"
```
brooksmtownsend: Same comment re: user running `make`

src/build.rs
```diff
@@ -0,0 +1,254 @@
...
+    };
+    let sign_output = sign_file(sign_options, output_kind)?;
+
+    if !command.push {
+        return Ok(sign_output);
+    }
+
+    println!("Signed actor: {}", sign_output.text);
+
+    // push it
```
brooksmtownsend, line 114: 🧂 🌶️

brooksmtownsend: I love the bop-it style comments here (but I think we don't need to push it at the end of `build`)

crates/wasmcloud-config-parser/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
+Cargo.lock
+
+# These are backup files generated by rustfmt
+**/*.rs.bk
+
+
+# Added by cargo
+
+/target
```
brooksmtownsend: Should this be removed?

crates/wasmcloud-config-parser/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "wasmcloud-config-parser"
+version = "0.1.0"
+edition = "2021"
+
```
brooksmtownsend: Should this be removed?

crates/wasmcloud-config-parser/.gitignore
brooksmtownsend: Looks like the config-parser is still leftover from copying the crate over

src/build.rs
```diff
@@ -0,0 +1,254 @@
...
+
+fn build_actor(
+    command: BuildCommand,
+    output_kind: OutputKind,
+    actor_config: ActorConfig,
+    language_config: LanguageConfig,
+    common_config: CommonConfig,
+) -> Result<CommandOutput> {
+    // build it
+    println!("Building actor...");
```
brooksmtownsend, line 67: What's the generally best way to support logging things where consumers of the crate can have control over their output? For example, if using this module with `wash build -o json` (once we have the JSON output), how can we suppress these printlns so that it can be piped?

cc @thomastaylor312 since I feel like you've probably dealt with this

brooksmtownsend (commented):
A couple nit requests and comments, but overall this looks great!👌🏻

crates/wasmcloud-config-parser/.gitignore
ghost: Weird. Deleted.

src/build.rs
ghost: Push isn't default, will only run if they pass a `--push` flag. Open to removing entirely though.

crates/wash-lib/src/parser/mod.rs
ghost: Added `trim` and `to_lowercase` for both project and language parsing.

crates/wash-lib/src/parser/mod.rs
ghost: Done.

crates/wash-lib/src/parser/mod.rs
ghost: Done.

crates/wash-lib/src/parser/mod.rs
ghost: Done.

crates/wash-lib/src/parser/mod.rs
ghost: Done.

src/build.rs
ghost: Probably need to set up a logging framework of some kind?

src/build.rs
brooksmtownsend: I do like `wash build --push` pushing to the registry, and it not being the default 👍🏻

src/build.rs
thomastaylor312: I've done this a couple ways and I am honestly not sure which one is best. I've done the super simple option of putting "log messages" like this to stderr and anything from the `-o` flag to stdout. I've also done something behind a trait/interface/class that implements this functionality for each type of output (might be the best option for Rust). I think this kind of functionality can wait though and we can add it as we iterate/improve `wash build`

crates/wash-lib/Cargo.toml
```diff
@@ -14,19 +14,26 @@ repository = "https://github.com/wasmcloud/wash"
...
+config = {version = "0.13.1", features = ["toml"]}
+semver = { version = "1.0.12", features = ["serde"]}
+serde = { version = "1", features = ["derive"] }
+serde_json = "1.0.82"
 
 [dev-dependencies]
-tokio = {version = "1", features = ["full"]}
\ No newline at end of file
+tokio = {version = "1", features = ["full"]}
+claim = { git = "https://github.com/Turbo87/rust-claim.git", rev = "23892a3" }
```
thomastaylor312: There a reason we aren't pulling in a normal version from crates? I am not sure we can publish the crate with a git dependency, even if it is in dev-dependencies

crates/wash-lib/Cargo.toml
```diff
@@ -14,19 +14,26 @@ repository = "https://github.com/wasmcloud/wash"
...
 command-group = { version = "1.0.8", features = ["with-tokio"] }
+anyhow = "1.0.58"
 futures = "0.3"
 tokio = {version = "1", default-features = false, features = ["process"]}
 reqwest = {version = "0.11", default-features = false, features = ["json", "rustls-tls", "stream"]}
 async-compression = { version = "0.3", default-features = false, features = ["tokio", "gzip"] }
 tokio-tar = "0.3"
 tokio-stream = "0.1"
 log = "0.4"
+config = {version = "0.13.1", features = ["toml"]}
```
thomastaylor312: I think a lot of these new deps can be `optional = true`? Then they can be added into the `parser` function and only pulled in when the user needs them in the library

crates/wash-lib/src/parser/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+#[derive(serde::Deserialize, Debug, PartialEq)]
+struct RawProviderConfig {
+    /// The capability ID of the provider.
+    pub capability_id: String,
+    /// The vendor name of the provider. Optional, defaults to 'NoVendor'.
+    pub vendor: Option<String>,
+}
+
+impl TryFrom<RawProviderConfig> for ProviderConfig {
```
thomastaylor312, line 102: I am not sure if I follow along with the need for a set of `Raw*` configs that convert to the normal configs. Most of the things I am seeing are defaulting things (doable with a custom impl of `Default` instead) and validation. It might be cleaner (and less structs) to have a private `Validate` trait that each of the structs implements to do any necessary validation rather than a bunch of `TryFrom` implementations. For example, if you defined the trait as

```rust
trait Validate {
    fn validate(&self) -> Result<()> {
        Ok(())
    }
}
```

That way you could just do `impl Validate for Blah {}` if you don't have anything to validate, and then otherwise put the validation logic in.

You could also look at the [`config` crate](https://docs.rs/config/latest/config/) as well to be able to merge different config options with defaults (I think there are others too)

crates/wash-lib/tests/parser/files/noconfig/test.txt
```diff
@@ -0,0 +1 @@
+This is here so the folder gets saved to git
```
thomastaylor312: Nit: The defacto way of doing this is an empty file called `.gitkeep` in the directory

src/build.rs
```diff
@@ -0,0 +1,254 @@
...
+    if !wasm_file.exists() {
+        bail!(
+            "Could not find compiled wasm file to sign: {}",
+            wasm_file.display()
+        );
+    }
+
+    // move the file out into the build/ folder for parity with tinygo and convienience for users.
+    let copied_wasm_file = PathBuf::from(format!("build/{}.wasm", common_config.name));
+    fs::create_dir_all(copied_wasm_file.parent().unwrap())?;
```
thomastaylor312: Basically the only instance where this returns `None` is if we are at `/`. Might be better to do an `if let Some(p) = copied_wasm_file.parent()` around the `create_dir_all` call rather than panicking. Same goes for the similar logic in the tinygo actor function below

thomastaylor312 (changes requested):
Mostly minor changes and discussion points. This is looking real good!

crates/wash-lib/Cargo.toml
ghost: Done.

crates/wash-lib/Cargo.toml
ghost: Ended up not using this, removed.

crates/wash-lib/src/parser/mod.rs
ghost: Looked into using `config`'s defaults for this, but it's a key/value based system ([link](https://docs.rs/config/latest/config/builder/struct.ConfigBuilder.html#examples)) which I think would get very messy to maintain.

I think a `Validate` trait works for the second-level configs, but i still need a `RawProjectConfig` because i'm changing the structure of the type when converting to `ProjectConfig`. I'll work on removing the other `Raw` ones though.

crates/wash-lib/tests/parser/files/noconfig/test.txt
ghost: Done thanks!

src/build.rs
ghost: Done!

crates/wash-lib/src/parser/mod.rs
ghost: Was looking into this some more. Tried using `config` defaults, it doesn't have enough flexibility to do what I want for this.

The problem i'm running into is that I want to narrow the types with the defaults, so we don't have tons of `Option`s to unwrap everywhere, leading to some potential '2nd defaults' where we unwrap the values. But to do that I need the `Raw` types I think, unless there is another way to do that more effectively. Any suggestions @thomastaylor312?

thomastaylor312: I think it's fine for now. We can come back and revisit as needed. The raw types aren't public so we can change that as desired

thomastaylor312 (dismissed):

thomastaylor312:
@brooksmtownsend You should probably give your final ✅ before @mattwilkinsonn merges since this is a bigger feature/PR

brooksmtownsend (changes requested):
Only requesting changes to temporarily hold the merge back against accidents, this looks great and I just want to release `0.12.0` before we merge

brooksmtownsend (dismissed):
This LGTM once tests pass! 🚀

autodidaddict:
Looks like the build is failing because there are 2 different versions of async nats being pulled into the dependency tree. We see this one routinely @brooksmtownsend

ghost:
> Looks like the build is failing because there are 2 different versions of async nats being pulled into the dependency tree. We see this one routinely @brooksmtownsend

Yep will fix today! Was busy the past few days.

brooksmtownsend (approved):
Time to build!

PR #300, opened 2022-08-29
Use `Into<String>` generic to make `CommandOutput` slightly easier to construct.

ghost:
A small change that reduces a bit of boilerplate. You don't have to convert `&str` and `String` everywhere.


brooksmtownsend (approved):
A great little ergonomic improvement 😄

PR #303, opened 2022-08-30
Update wash-lib with minimum version requirement and mix releases

brooksmtownsend:
This PR updates the `wash-lib` crate to use the wasmCloud host version 0.57.0, which is using mix releases instead of distillery releases. As a part of this the commands changed to run and evaluate wasmCloud hosts, so the wash-lib needed an upgrade. After some consensus, we've decided to restrict the minimum version to 0.57 and drop support for distillery releases with wash-lib going forward as this will bring an unnecessary maintenance burden for a small use case (manually specifying old versions of the host)

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -21,7 +21,7 @@ pub(crate) const WASMCLOUD_HOST_BIN: &str = "bin/wasmcloud_host";
 #[cfg(target_family = "windows")]
 pub(crate) const WASMCLOUD_HOST_BIN: &str = "bin\\wasmcloud_host.bat";
 
-// Any version of wasmCloud under 0.57.0 uses distillery releases and are incompatible
+// Any version of wasmCloud under 0.57.0 uses distillery releases and is incompatible
 const MINIMUM_WASMCLOUD_VERSION: &str = "v0.57.0";
 
 /// A wrapper around the [ensure_wasmcloud_for_os_arch_pair] function that uses the
@@ -243,38 +243,17 @@ where
         ));
     }
 
-    // Windows powershell will ping forever if it's the first command,
-    // this is essentially an initialization
-    //TODO: Mix release version of this
-    // #[cfg(target_family = "windows")]
-    // {
-    //     let tmp_log = std::env::temp_dir().join("init.log");
-    //     let init_run = tokio::fs::File::create(&tmp_log).await?.into_std().await;
-    //     let mut child = tokio::process::Command::new(bin_path.as_ref())
-    //         .stdout(init_run)
-    //         .arg("ping")
-    //         .spawn()?;
-    //     // Give the wasmcloud initialization a few seconds to unpack
-    //     for _ in 0..5 {
-    //         let log_contents = tokio::fs::read_to_string(&tmp_log).await?;
-    //         if log_contents.is_empty() {
-    //             // Give just a little bit of time for the startup logs to flow in
-    //             tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
-    //         } else {
-    //             tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
-    //             break;
-    //         }
-    //     }
-    //     let _ = child.kill();
-    // }
-
     match Command::new(bin_path.as_ref()).arg("pid").output().await {
         // If ping was successful, returning "pong", another host is already running
-        Ok(output) if output.status.success() => {
-            return Err(anyhow!(
-                "Another wasmCloud host is already running on this machine with PID {}",
-                String::from_utf8_lossy(&output.stdout)
-            ));
+        Ok(output) => {
+            // Stderr will include :nodedown if no other host is running, otherwise
+            // stdout will contain the PID
+            if !String::from_utf8_lossy(&output.stderr).contains(":nodedown") {
+                return Err(anyhow!(
+                    "Another wasmCloud host is already running on this machine with PID {}",
+                    String::from_utf8_lossy(&output.stdout)
+                ));
+            }
         }
         _ => (),
     }
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -472,19 +472,23 @@ mod test {
         assert!(child_res.is_err());
 
         // Should fail because another erlang wasmcloud_host node is running
-        let mut host_env = HashMap::new();
-        host_env.insert("PORT".to_string(), "4002".to_string());
-        host_env.insert("WASMCLOUD_RPC_PORT".to_string(), nats_port.to_string());
-        host_env.insert("WASMCLOUD_CTL_PORT".to_string(), nats_port.to_string());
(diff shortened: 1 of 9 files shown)


crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -239,36 +245,36 @@ where
...
+    //         } else {
+    //             tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+    //             break;
+    //         }
+    //     }
+    //     let _ = child.kill();
+    // }
 
-    match Command::new(bin_path.as_ref()).arg("ping").output().await {
+    match Command::new(bin_path.as_ref()).arg("pid").output().await {
```
connorsmith256, line 247: This probably doesn't matter for the majority of use cases, but I do wonder about users who want to `wash up` when a host is already runnning

brooksmtownsend: Are you suggesting to add the feature to start multiple hosts, or to add better log verbosity telling the user what host is running?

src/up/mod.rs
```diff
@@ -477,24 +477,21 @@ async fn stop_wasmcloud<P>(bin_path: P) -> Result<()>
...
-                    return;
-                }
-            }
-        }
-    })
-    .await?;
+        .output()
+        .await?;
+
+    //TODO: evaluate necessity
```
ghost: Is this needed still?

ghost (commented):
Could you change the title of the PR slightly, I thought using mix releases was the primary change, but really it is setting a version requirement at 0.57 which then runs mix releases

src/up/mod.rs
brooksmtownsend: I don't think so! 😄

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -319,9 +303,23 @@ fn wasmcloud_url(os: &str, arch: &str, version: &str) -> String {
         WASMCLOUD_GITHUB_RELEASE_URL, version, arch, os
     )
 }
+
+/// Helper function to ensure the version of wasmCloud is above the minimum
+/// supported version (v0.57.0) that runs mix releases
+fn check_version(version: &str) -> Result<()> {
+    if version < MINIMUM_WASMCLOUD_VERSION {
```
thomastaylor312: This will fail if we end up at a `0.100` version. Highly recommend we use the `semver` crate instead. 

Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=06bdf7e362d503a1393f259d2c6689af

thomastaylor312 (dismissed):

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: Interesting! Good call

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -227,46 +233,25 @@ where
...
-    }
-
-    match Command::new(bin_path.as_ref()).arg("ping").output().await {
-        // If ping was successful, returning "pong", another host is already running
-        Ok(output) if output.status.success() => {
-            if String::from_utf8_lossy(&output.stdout).contains("pong") {
+    #[cfg(target_family = "unix")]
+    match Command::new(bin_path.as_ref()).arg("pid").output().await {
+        Ok(output) => {
+            // Stderr will include :nodedown if no other host is running, otherwise
```
ghost, line 249: I might be misreading this but isn't this code reversed? It reads to me like:
if `:nodedown` is returned, then a wasmCloud host is already running

src/up/mod.rs
```diff
@@ -477,24 +477,11 @@ async fn stop_wasmcloud<P>(bin_path: P) -> Result<()>
...
-
-            while let Ok(Some(line)) = lines.next_line().await {
-                if line == "ok" {
-                    return;
-                }
-            }
-        }
-    })
-    .await?;
+        .output()
```
ghost, line 483: Do we want to check the output here to make sure it stopped successfully?

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: It's a little weird, but the scenario is
- If a wasmCloud host is running with that node name (defaults to wasmcloud_host@localhost or similar), return a PID
- If a wasmCloud host is not running with that node name, return an RPC error containing `:nodedown`

What I want to check here is that the response does not contain `:nodedown`, which means that a PID came back instead. I did it this way instead of trying to regex for a pid.

To be fair, this block should go away soon entirely, I want to add multiple host support so that `wash up` can handle more than one mix release running at a time

src/up/mod.rs
brooksmtownsend: Let me check on what all this output contains and if I can simulate a failure, I've never seen the `stop` command fail or give back information other than `ok`

brooksmtownsend: Good call, I can at least check for a non-zero exit status

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -319,18 +304,38 @@ fn wasmcloud_url(os: &str, arch: &str, version: &str) -> String {
         WASMCLOUD_GITHUB_RELEASE_URL, version, arch, os
     )
 }
+
+/// Helper function to ensure the version of wasmCloud is above the minimum
+/// supported version (v0.57.0) that runs mix releases
+fn check_version(version: &str) -> Result<()> {
+    let version_req = semver::VersionReq::parse(&format!(">={}", MINIMUM_WASMCLOUD_VERSION))?;
+    match semver::Version::parse(version.trim_start_matches('v')) {
+        Ok(parsed_version) if !version_req.matches(&parsed_version) => Err(anyhow!(
```
connorsmith256, line 313: TIL about this crate. Cool!
https://docs.rs/semver/1.0.13/semver/struct.VersionReq.html

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -456,23 +468,56 @@ mod test {
...
+        assert!(check_version("v0.100.0").is_ok());
+        assert!(check_version("v0.203.0").is_ok());
+
+        // Ensure we deny versions < 0.57.0
+        assert!(check_version("v0.48.0").is_err());
+        assert!(check_version("v0.56.0").is_err());
+        assert!(check_version("v0.12.0").is_err());
+        assert!(check_version("v0.56.999").is_err());
+        if let Err(e) = check_version("v0.56.0") {
+            assert_eq!(e.to_string(), "wasmCloud version v0.56.0 is earlier than the minimum supported version of v0.57.0");
```
connorsmith256, line 512: Nit: I wouldn't couple tests passing to matching on this specific error message. If we tweak the error message, this test would need to be updated, which shouldn't be necessary

connorsmith256 (approved):

crates/wash-lib/src/start/wasmcloud.rs
ghost: Sg

ghost (approved):

Issue #273, opened 2022-05-24
[BUG] `wash claims inspect` reports incorrect error message for missing file


brooksmtownsend:
**Describe the bug**
I was attempting to inspect an actor file when I had the incorrect filename, and I got the following error:
```
wash claims inspect build/echotg_s.wasm

Pulling artifacts with tag 'latest' is prohibited. This can be overriden with a flag
```

Or with JSON output:
```
wash claims inspect build/echotg_s.wasm -o json

{
  "success": false,
  "error": "Pulling artifacts with tag 'latest' is prohibited. This can be overriden with a flag"
}
```

This is because the original failure to find a file results in a fallback to pull an OCI artifact. We should be able to parse the argument for `claims inspect` and `par inspect` to determine if it's truly a path or a URL and report errors accordingly.

**To Reproduce**
1. Install `wash`
2. Run `wash claims inspect hooha/notareal/path.wasm`
3. View error

**Expected behavior**
In this case where I had the filename wrong, I should get an error stating that no file was found. We can still default to trying to pull an OCI artifact _just in case_ there's some edge case with a URL that looks like a filepath (e.g. `build.com/myartifact/wasm` is both a valid URL and path) but we should at least have an indication that `wash` couldn't find a file.

brooksmtownsend:
Pretty sure this was fixed with #379

PR #305, opened 2022-09-01
add artifact not found error

Mo-Fatah:
relates to #273 

I think this can be a temporary fix to indicate that the artifact couldn't be found locally or from a registry.

### DIFF
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -202,10 +202,13 @@ pub(crate) async fn pull_artifact(
 ) -> Result<Vec<u8>> {
     let image: Reference = url.parse()?;
 
-    if image.tag().unwrap_or("latest") == "latest" && !allow_latest {
-        bail!(
-            "Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag --allow-latest"
-        );
+    if image.tag().unwrap_or("latest") == "latest" {
+        if !allow_latest {
+            bail!(
+                "Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag --allow-latest"
+            );
+        }
+        bail!("Artifact not found");
     };
 
     let mut client = Client::new(ClientConfig {


ghost:
Hi @Mo-Fatah, thanks for contributing! From reading the code I think your changes would fix #273, although I haven't tested it yet. Not sure why you closed the PR, but if you need any help with this feel free to reach out. We often discuss changes and ask for advice on the wasmCloud Slack, here is the [link](https://slack.wasmcloud.com/).

Mo-Fatah:
Thanks @mattwilkinsonn !! I thought that my fix solves the issue. But after I filed the PR , I ran the tests and I had a failed test case 
```
test integration_claims_inspect ... FAILED
```
This why I closed my PR thinking that my fix is causing the test failure.

But I noticed something now, this test case also fails on `main upstream`. Is this issue known for you or I am missing something here ? 
I am using `Ubuntu 20.04.4 LTS`

> I have reopened the PR

thomastaylor312:
@Mo-Fatah I didn't see any failing tests in main (we had a failed build of a docker image, but nothing from tests). Let's see how the PR tests do because perhaps it could be local.

Also, you'll need to sign off your commit! Details instructions can be found by clicking the "Details` link on the failed DCO check

Mo-Fatah:
> @Mo-Fatah I didn't see any failing tests in main (we had a failed build of a docker image, but nothing from tests). Let's see how the PR tests do because perhaps it could be local.
> 
> Also, you'll need to sign off your commit! Details instructions can be found by clicking the "Details` link on the failed DCO check

Yeah, all the checks here have passed, apparently this is a local failure. 

I have singed off my commit, Thank you!

src/reg.rs
```diff
@@ -202,10 +202,13 @@ pub(crate) async fn pull_artifact(
...
-        bail!(
-            "Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag --allow-latest"
-        );
+    if image.tag().unwrap_or("latest") == "latest" {
+        if !allow_latest {
+            bail!(
+                "Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag --allow-latest"
+            );
+        }
+        bail!("Artifact not found");
```
brooksmtownsend, line 211: Just to give the user a better sense for what went on, it might be better to phrase this as 
```
Artifact not found in registry, please ensure {} exists and is a valid OCI reference
```
Or something like that, so that users can see what reference they are trying to download. Thank you for submitting this!

brooksmtownsend (changes requested):

brooksmtownsend:
This also seems to still reproduce the same issue, likely because the image tag doesn't exist so it's unwrapping to `latest` as a fallback 
```
> wash claims inspect echo_s.wasm

Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag --allow-latest
```

Mo-Fatah:
Yes you are right.
I thought of it as: when the user enter a not-found artifact, they will be prompted to try the `--allow-latest` flag. When they try again with the flag, they will get a not-found error. A kind of: try everything before getting a final not-found error. 

Not very accurate. Do you have suggestions ? besides a more detailed message.

brooksmtownsend:
> Do you have suggestions ? besides a more detailed message.

So what we're really trying to do here is provide a useful error message for people when they try to inspect a local file but they type in the name wrong. Here's what I think would be a good solution:

- If the user types in an OCI reference we require a tag (e.g. you can't just do `wasmcloud.azurecr.io/echo`, you have to type out `wasmcloud.azurecr.io/echo:latest` to use the `latest` tag)
    - I know this isn't what Docker does, but we have explicit `latest` checks so I think it would be better to explicitly require that tag.
- If the user doesn't include a tag, then assume it's a file, and if it's not found then we return an error message about the file not being found
- If the user includes the `latest` tag, then we can do any checks for the `--allow-latest` flag

Thoughts @mattwilkinsonn / @thomastaylor312 ?

thomastaylor312:
> If the user doesn't include a tag, then assume it's a file, and if it's not found then we return an error message about the file not being found

This error should probably say something about file not found and then 

```
(Were you specifying an image? Image tags are required for wasmCloud)
```

brooksmtownsend:
@Mo-Fatah I'm going to close this as it seems to have gotten a little stale, please feel free to re-open once you've added a few changes or want to re-approach this PR

PR #307, opened 2022-09-02
Add Docker build job in pull requests

ghost:
Adds the Docker build job to all pull requests, so we don't run into a postsubmit problem with Docker like we have ATM.

### DIFF
--- a/.github/workflows/docker_check.yml
+++ b/.github/workflows/docker_check.yml
@@ -1,4 +1,4 @@
-name: Build Docker Image
+name: Docker
 
 on:
   push:
--- a/.github/workflows/docker_check.yml
+++ b/.github/workflows/docker_check.yml
@@ -1,0 +1,52 @@
+name: Build Docker Image
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  build_image:
+    name: Build Docker Image
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/docker_check.yml
```diff
@@ -0,0 +1,52 @@
...
+          tags: |
+            wasmcloud/wash:alpine
+          labels: ${{ steps.meta.outputs.labels }}
+      - name: Build and push - scratch
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          file: "Dockerfile"
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
brooksmtownsend: This looks like an extra step

.github/workflows/docker_check.yml
```diff
@@ -0,0 +1,52 @@
...
+            org.opencontainers.image.title=wash
+            org.opencontainers.image.description=WAsmcloud SHell
+            org.opencontainers.image.vendor=wasmCloud
+            org.opencontainers.image.source=https://github.com/${{ github.repository }}
+      - name: Build
+        id: docker_build_builder
+        uses: docker/build-push-action@v2
+        with:
+          file: "Dockerfile"
+          target: release-alpine
```
brooksmtownsend, line 42: We should probably add an explicit `push: false` to this action just to make sure we aren't pushing on PR

I know it's the default, but I don't want to get burned by a GH action accidentally updating under us 😄

brooksmtownsend (changes requested):
Small request

.github/workflows/docker_check.yml
ghost: Removed.

.github/workflows/docker_check.yml
ghost: Done.

brooksmtownsend (approved):

Issue #281, opened 2022-05-31
[CHORE] `wash new` should initialize a git repository with the `main` branch


brooksmtownsend:
Mostly title, most new initializations of git repositories use `main` instead of `master` for their primary branch

brooksmtownsend:
Closed by #308

PR #308, opened 2022-09-02
`wash new` : set the default branch to main instead of master for new git repos

Mo-Fatah:
this should solve #281


brooksmtownsend (approved):
Worked great!

PR #309, opened 2022-09-02
Parallelize the Rust CI jobs for faster feedback

ghost:
Splits our one CI job into three: clippy, unit tests, integration tests. Only runs unit tests on multiple platforms (clippy will always be the same and the integration tests were already only running on Linux).

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -12,7 +12,7 @@ env:
 jobs:
   fmt_clippy:
     name: Fmt & Clippy Checks
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
       - name: Run Rust check + clippy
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_ci.yml
```diff
@@ -10,23 +10,31 @@ env:
...
+      - uses: actions/checkout@v2
+      - name: Run Rust check + clippy
+        run: make rust-check
+  unit_tests:
+    name: Unit Tests
     strategy:
-      fail-fast: false
       matrix:
-        os: [ubuntu-20.04, windows-2022, macos-11]
+        os: [ubuntu-latest, windows-2022, macos-11]
```
brooksmtownsend: I'm not hard-set in my ways but I do prefer `ubuntu-22.04`, not a blocker since they are the same thing

brooksmtownsend (dismissed):

.github/workflows/rust_ci.yml
ghost: Changed

brooksmtownsend (approved):

PR #310, opened 2022-09-05
Add more context to `ctl` error messages

ghost:
When getting errors from `ctl` commands, it is hard to determine where/why the error happened because there isn't much context. In this PR i'll be going through and adding lots of `with_context()` calls to errors. WIP still.


stevelr:
@mattwilkinsonn this will be a great improvement in error messages! Seems like we should merge the client debug change first and then review this pr after you've changed the messages.

brooksmtownsend (dismissed):
Love this improvement and how we can take error messages from elsewhere but still add on additional context (apt naming :D)

Approved for now and can re-review once the wasmcloud-control-interface version 0.20.0 is used

ghost:
No point leaving this as a draft forever! Marking it for review so it can be submitted.

I'll do another pass on adding these Soon™

autodidaddict (approved):

PR #314, opened 2022-09-21
Only run unit tests in release job

ghost:
Fixes release CI by only running the unit tests. Prevents us from having to maintain integration test dependencies in 4 different jobs. There's probably a fancy way to set this up so they all call out to the same job, but this is good enough IMO. We can just not cut a release when integration tests fail.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Build
         run: cargo build --verbose
       - name: Run tests
-        run: make test-all
+        run: make test
       - name: Check fmt
         run: cargo fmt -- --check
 
--- a/.github/workflows/release_amd64.yml
+++ b/.github/workflows/release_amd64.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Build
         run: cargo build --verbose
       - name: Run tests
-        run: make test-all
+        run: make test
       - name: Check fmt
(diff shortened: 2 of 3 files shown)


brooksmtownsend (commented):
We should be running the same tests in PRs that we are when we release, right? I just want to make sure we're not just bandaiding a release issue when it might end up releasing code that fails integration tests due to GitHub trying to smart-merge (which has happened before)

ghost:
> We should be running the same tests in PRs that we are when we release, right? I just want to make sure we're not just bandaiding a release issue when it might end up releasing code that fails integration tests due to GitHub trying to smart-merge (which has happened before)

PRs have to pass integration tests before they can merge anyway no? And then we run the tests again after merge into main. IMO that's enough layers of protection, and prevents us from maintaining multiple parallel integration test setups.

The main edge case here would be if we merged a PR which was still pending the integration tests, didn't wait for the `main` tests to finish either, and cut a new release from it. Even in that case (which shouldn't happen unless we're really overly rushing something, the test takes around ~15mins) we can just release a +0.0.1 patch version afterwards.

brooksmtownsend (approved):
Sounds good to me, I'm happy to try this and, like you said, this should only affect us if we rush releases out before waiting for the tests to run. Probably best to keep the release time quick

PR #317, opened 2022-09-30
clap v4

ricochet:
A series of chores with the end goal of updating to clap v4. The following was done in stages and is self-contained by commit:

- Fixed clippy warnings
- Bumped dependencies
- Then updated clap to v4

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -40,11 +40,27 @@ pub(crate) struct UpCommand {
 
 #[derive(Parser, Debug, Clone)]
 pub(crate) struct NatsOpts {
+    /// Optional path to a NATS credentials file to authenticate and extend existing NATS infrastructure.
+    #[clap(
+        long = "nats-credsfile",
+        env = "NATS_CREDSFILE",
+        requires = "nats_remote_url"
+    )]
+    pub(crate) nats_credsfile: Option<PathBuf>,
+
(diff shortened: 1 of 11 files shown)


src/up/mod.rs
```diff
@@ -149,15 +149,15 @@ pub(crate) struct WasmcloudOpts {
     pub(crate) prov_rpc_port: Option<u16>,
 
     /// A seed nkey to use to authenticate to NATS for Provider RPC messages
-    #[clap(long = "prov-rpc-seed", env = WASMCLOUD_PROV_RPC_SEED, requires = "prov-rpc-jwt")]
+    #[clap(long = "prov-rpc-seed", env = WASMCLOUD_PROV_RPC_SEED, requires = "prov_rpc_jwt")]
```
brooksmtownsend, line 152: was this actually incorrect before? Curious if the behavior changed when you specify only one

brooksmtownsend (commented):
Looks good! just one question that we may not test automatically

src/up/mod.rs
ricochet: Clap V4 changed this. It was a compilation error. There were changes with how they check for "Verbatim" args. I bet there's a clap setting I could have also set but this seemed to work with their defaults.

brooksmtownsend (approved):

thomastaylor312 (approved):

PR #318, opened 2022-10-12
Set stdin to null when starting a wasmcloud host with wash-lib

protochron:
By default `tokio::process::Command` inherits stdin from the running process, which means that if you're trying to start a wasmcloud host with wash-lib it will capture stdin and make that particular terminal unusable until you kill the wasmcloud process.

This fixes the issue by setting stdin to null when starting the process.


brooksmtownsend (commented):
Should we do this for the NATS server start as well? LGTM

protochron:
@brooksmtownsend probably, yeah. Let me add that too

brooksmtownsend (approved):

PR #324, opened 2022-10-30
remove git2 and openssh dependencies

stevelr:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ bytes = "1.0"
 cargo_atelier = "0.2"
 clap = { version="4", features=["derive", "env"] }
 cloudevents-sdk = "0.5.0"
-cmd_lib = "1.3.0"
 console = "0.15"
 ctrlc = "3.2.2"
 derive_more = {version = "0.99.16", default-features = false, features = ["display"]}
@@ -60,11 +59,12 @@ which = "4.2.2"
 wash-lib = { version = "0.3.1", path = "./crates/wash-lib" }
 wascap = "0.8.0"
 weld-codegen = "0.5.0"
-wasmcloud-control-interface = "0.20.1"
+wasmcloud-control-interface = "0.21.0"
 wasmbus-rpc = "0.10.0"
 wasmcloud-test-util = "0.5.0"
 
 [dev-dependencies]
+cmd_lib = "1.3.0"
 reqwest = {version = "0.11", default-features = false, features = ["json", "rustls-tls"]}
 tempfile = "3"
 test-case = "2.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.13.0"
+version = "0.14.0-alpha.1"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "command-line-utilities"]
 description = "wasmcloud Shell (wash) CLI tool"
@@ -22,6 +22,7 @@ bytes = "1.0"
 cargo_atelier = "0.2"
 clap = { version="4", features=["derive", "env"] }
 cloudevents-sdk = "0.5.0"
+cmd_lib = "1.3.0"
 console = "0.15"
 ctrlc = "3.2.2"
 derive_more = {version = "0.99.16", default-features = false, features = ["display"]}
@@ -29,7 +30,6 @@ dialoguer = "0.10.2"
 dirs = "4.0"
 env_logger = "0.9"
 envmnt = "0.10.2"
-git2 = "0.15.0"
 heck = "0.4"
 ignore = "0.4"
 indicatif = "0.17.0"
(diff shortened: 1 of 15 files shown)


autodidaddict (commented):
Does this work on Linux Windows and Mac?

stevelr:
> Does this work on Linux Windows and Mac?

Yes Yes and Yes

(Windows tested with WSL. Should be tested on Windows without WSL but I ran into unrelated complications building the 'ring' dependency and need to create a new build/test environment).

stevelr:
This resolves one of the most common sources of cross build failures and improves portability.   (The other platform dependency is 'ring' which is harder to remove but has better support for cross compilation).

My first plan to remove the OpenSSH dependency was to replace the git2 crate with pure rust crates that use git's rest api over reqwest and rustls, but that would have had much higher complexity than the approach taken here, which adds a dependency on git cli.

The minimum version of git cli required is 2.26.0 (3/22/2020) when 'sparse-checkout set' was added. Current git cli version is 2.38.1 as of 10/07/22.

Cargo.toml
```diff
@@ -22,14 +22,14 @@ bytes = "1.0"
 cargo_atelier = "0.2"
 clap = { version="4", features=["derive", "env"] }
 cloudevents-sdk = "0.5.0"
+cmd_lib = "1.3.0"
```
brooksmtownsend: Is there a reason why we use this instead of what the other things in `wash` use, tokio Commands? I like the convenience of this macro but I'm concerned with having multiple ways of "shelling out"

brooksmtownsend (commented):
This works well, just a single question 😄

Cargo.toml
stevelr: I used it because it was simpler. I think there are three ways to shell out now: `build` uses std::process and `up` uses tokio::process. If we wanted consistency I'd vote to file an issue to convert them to macro. I don't think there's any need for them to be async if they are always used in a cli, but if we needed async we could call the macro inside a spawn.

autodidaddict (dismissed):

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-cli"
-version = "0.13.0"
+version = "0.14.0-alpha.1"
```
connorsmith256, line 3: Do we want to merge alpha versions to Cargo.toml? @thomastaylor312 thoughts?

brooksmtownsend: My motivation for starting this a little while ago came from the "Rust for Rustaceans" book, it's a good way to update the version without people actually auto-upgrading to this version and indicate that there have been changes in main. I don't mind bumping in separate PRs but I like this a little bit better because as soon as this merges `main` is not `0.13.0` anymore. Definitely curious of Taylor's thoughts

Cargo.toml
thomastaylor312: I think we should stick with tokio. Nothing we are doing is fancy enough to warrant the use of the macro and there is no support for async (meaning we'd have to do thread blocking). If we were heavily shelling, I'd consider it, but I think I am against adding this dependency

Cargo.toml
thomastaylor312: We talked about this in meat space, but alpha version changes to Cargo.toml are fine and do exactly what @brooksmtownsend is intending for them to do

tests/integration_up.rs
```diff
@@ -1,27 +1,20 @@
...
-        .args(&["up", "--nats-port", "5893", "-o", "json", "--detached"])
-        .stdout(stdout)
-        .spawn()
-        .expect("Could not spawn wash up process");
-
-    let status = up_cmd.wait().expect("up command failed to complete");
-
-    assert!(status.success());
-    let out = read_to_string(&path).expect("could not read output of wash up");
+    //let path = dir.join("washup.log");
```
thomastaylor312: Should these still be commented out?

thomastaylor312 (dismissed):
Nothing much new here. Just putting as a request changes until we can discuss more around the cmd macros

thomastaylor312:
I would like to note that I'm not opposed to having the `run_cmd` macros be a `dev-dependency`. I can see its usefulness there, but I'd also consider that as out of scope for this PR

Cargo.toml
stevelr: > there is no support for async (meaning we'd have to do thread blocking)

actually cmd_lib does have support for async - it's just a different macro.
but many (most?) of the uses of tokio Command in wash don't need tokio because it just awaits the result.

Is your objection because of the fact that it's synchronous, or because we've added another dependency to wash?

tests/integration_up.rs
stevelr: fixed

Cargo.toml
thomastaylor312: Added another dependency (this got picked up in a later comment, but putting response here for history's sake)

thomastaylor312:
To clarify my last comment: I am not in favor of adding the new macro dep. I don’t see the need to add yet another dependency to something that already has a ton of deps. Plus it abstracts away too much IMO. Being able to access the status of things is important, plus I still think the normal rust stuff reads better (in addition to being out of scope for a PR like this)

stevelr:
Followed Taylor's suggestion to remove from dependencies. Left the existing uses in tests (run_cmd moved to dev-dependencies).

brooksmtownsend (approved):
LGTM!

PR #327, opened 2022-11-04
Feat/wash down

### LINKED ISSUE #322: [FEATURE] Add a `wash down` command
### Context

Starting `wash` in detached mode results in several processes being started:

* A NATS server
* An Erlang port mapper daemon
* An Erlang OTP server (`wasmcloud_host`)
* A couple of Erlang runtime servers

### Problem

Once `wash up -d` command has started these processes, control is then passed back to the command issuer (typically a shell).  

A side-effect of the return of control is that there is now no clean way to shut down the above processes.  Instead, the relevant process ids must first be identified with a command such as `ps -ef|grep wash` and then killed manually.

### Proposed Solution

Implement a `wash down` command to stop all processes previously started by the `wash up -d` command

brooksmtownsend:
Fixes #322 

This PR
1. Adds the command `wash down` which cleans up running wasmCloud hosts and NATS servers launched by `wash up`
2. Adds an additional output file for NATS which includes the PID of the `nats-server` process
3. Adds another option to `wash drain` which lets users easily clean up their wash downloads directory
4. Bumps the host version to `0.58.2`
5. Creates an alias for `detached` called `detach` because I kept typing it wrong and it made me angry
6. Simplifies some of the `up` code now that we don't need to output a complicated stop command

### DIFF
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.3.1"
+version = "0.3.2"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "wasmcloud"]
 description = "wasmcloud Shell (wash) libraries"
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.3.2"
+version = "0.4.0-alpha.1"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "wasmcloud"]
 description = "wasmcloud Shell (wash) libraries"
--- a/src/down/mod.rs
+++ b/src/down/mod.rs
@@ -1,10 +1,11 @@
-use anyhow::{anyhow, Result};
-use clap::Parser;
-use serde_json::json;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::process::{Output, Stdio};
 use std::time::Duration;
+
+use anyhow::Result;
+use clap::Parser;
+use serde_json::json;
 use tokio::process::Command;
 
 use crate::appearance::spinner::Spinner;
@@ -83,7 +84,7 @@ where
         .arg("stop")
         .output()
         .await
-        .map_err(|e| anyhow!(e))
+        .map_err(anyhow::Error::from)
 }
 
 /// Helper function to send the nats-server the stop command
@@ -104,7 +105,7 @@ where
         .stdin(Stdio::null())
         .output()
         .await
-        .map_err(|e| anyhow!(e));
+        .map_err(anyhow::Error::from);
 
     // remove PID file
     if pid_file.is_file() {
--- a/src/down/mod.rs
+++ b/src/down/mod.rs
@@ -1,0 +1,94 @@
+use anyhow::Result;
+use clap::Parser;
+use serde_json::json;
+use std::collections::HashMap;
+use std::process::Stdio;
+use std::time::Duration;
+use tokio::process::Command;
+
+use crate::appearance::spinner::Spinner;
+use crate::cfg::cfg_dir;
+use crate::up::DOWNLOADS_DIR;
+use crate::util::{CommandOutput, OutputKind};
+use wash_lib::start::*;
+
+#[derive(Parser, Debug, Clone)]
+pub(crate) struct DownCommand {}
+
+pub(crate) async fn handle_command(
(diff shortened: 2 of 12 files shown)


crates/wash-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.3.1"
+version = "0.3.2"
```
thomastaylor312: You are adding new functionality (and constants) to wash-lib, so this should be a 0.4. Also, might want this to be an alpha until we cut a new wash release

src/down/mod.rs
```diff
@@ -0,0 +1,122 @@
+use anyhow::{anyhow, Result};
+use clap::Parser;
+use serde_json::json;
+use std::collections::HashMap;
```
thomastaylor312, line 1: Nit: please put `std` imports up top, separated from imported crates

src/down/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+            if output.stderr.is_empty() && output.stdout.is_empty() {
+                // if there was a host running, 'stop' has no output.
+                // Give it time to stop before stopping nats
+                tokio::time::sleep(Duration::from_secs(6)).await;
+                out_json.insert("host_stopped".to_string(), json!(true));
+                out_text.push_str("✅ wasmCloud host stopped successfully\n");
+            } else {
+                out_json.insert("host_stopped".to_string(), json!(true));
+                out_text.push_str(
+                    "🤔 Host did not appear to be running, assuming it's already stopped\n",
```
thomastaylor312, line 49: A+ emoji usage

src/down/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+                "❌ NATS server did not stop successfully: {:?}\n",
+                e
+            ));
+        } else {
+            out_json.insert("nats_stopped".to_string(), json!(true));
+            out_text.push_str("✅ NATS server stopped successfully\n");
+        }
+    }
+
+    out_json.insert("success".to_string(), json!(true));
```
thomastaylor312, line 70: Out of scope comment for future discussion: I think it might be better to just have a `serde_json::Value` that we use instead of a `HashMap` just to make all of this read easier

src/down/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+pub(crate) async fn stop_wasmcloud<P>(bin_path: P) -> Result<Output>
+where
+    P: AsRef<Path>,
+{
+    Command::new(bin_path.as_ref())
+        .stdout(Stdio::piped())
+        .arg("stop")
+        .output()
+        .await
+        .map_err(|e| anyhow!(e))
```
thomastaylor312: Nit: Might be cleaner to just do `.map_err(anyhow::Error::from)`

src/down/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+        .arg("--signal")
+        .arg(signal)
+        .stdin(Stdio::null())
+        .output()
+        .await
+        .map_err(|e| anyhow!(e));
+
+    // remove PID file
+    if pid_file.is_file() {
+        let _ = tokio::fs::remove_file(&pid_file).await;
```
thomastaylor312, line 112: Any need for a warning if remove file fails? I don't think so but wanted to check

thomastaylor312 (changes requested):
Only one blocking semver violation and a few other nits/ramblings

src/down/mod.rs
brooksmtownsend: I don't think so, if we fail to remove it for some reason the next time we run `wash up` we'll just overwrite it. Only thing I could think of is, in the future, if we supported multiple `wash up`'s in a row to "smart" check for a running NATS server 🤔 

For now I think I'll leave it off and we can approach that when it comes time

src/down/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+    output
+}
+
+/// Helper function to get the path to the NATS server pid file
+pub(crate) fn nats_pid<P>(install_dir: P) -> PathBuf
```
autodidaddict: super ultra nit: I assume a function called `nats_pid` returns a number/string with the actual pid in it, not a pathbuf.

autodidaddict (dismissed):

src/down/mod.rs
brooksmtownsend: It's a super ultra nit sir, but it checks out

brooksmtownsend:
Waiting on #329 to merge only because that modifies the `up` test to not infinite hang

thomastaylor312 (approved):

PR #328, opened 2022-11-08
Initial implementation of support for lattice artifacts in wash-lib (no CLI)

autodidaddict:
This adds a `lattice_artifacts` module to wash lib, allowing CLI tools like `wash` to be able to read/write artifacts to the lattice artifact cache. See [this PR](https://github.com/wasmCloud/wasmcloud-otp/pull/495) for the corresponding wasmCloud host implementation.

⚠️ The wash binary will not compile here because wash lib uses a version of async nats that is significantly newer than what wash currently has, so there will be a dependency tree failure during compilation. Leaving this PR as a draft until the wash main binary can be upgraded to use a version of async nats _greater than_ **0.21.0** because we're also awaiting a fix from NATS in the Rust client.

Recap:
* Once fix is available, upgrade wash-lib to newest async nats
* Once fix is available downstream from wasmbus-rpc, upgrade `wash` binary to use that
* Build will compile, can un-draft this PR

### DIFF
--- a/crates/wash-lib/src/lattice_artifacts/mod.rs
+++ b/crates/wash-lib/src/lattice_artifacts/mod.rs
@@ -1,0 +1,95 @@
+//! The `lattice_artifacts` module contains functionality relating to reading and writing
+//! actor module bytes to and from a NATS object store corresponding to a lattice.
+//!
+use anyhow::{anyhow, Result};
+use async_nats::jetstream::object_store::{Config, ObjectStore};
+use tokio::io::AsyncReadExt;
+
+fn bucket_name(lattice_id: &str) -> String {
+    format!("ARTIFACT_{}", lattice_id)
+}
+
+/// Creates a new artifact store named ARTIFACT_{lattice_id} using the default
+/// options (file-backed with no limitations) or, if one already exists, returns
+/// a reference to that object store.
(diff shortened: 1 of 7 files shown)


crates/wash-lib/src/lattice_artifacts/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+
+/// Removes an item from the artifact store
+pub async fn remove_item_from_store(store: ObjectStore, public_key: &str) -> Result<()> {
+    store
+        .delete(public_key)
+        .await
+        .map_err(|e| anyhow!("Failed to delete item from artifact store: {}", e))
+}
+
+pub async fn get_item_from_store(store: ObjectStore, public_key: &str) -> Result<Vec<u8>> {
```
stevelr, line 52: It might be more generally useful and safer to return the stream, e.g.,  `Result<Box<dyn AsyncRead>>`

autodidaddict: If we want a stream then we can always add a method for that. For now this covers the minimal expected usage

autodidaddict:
After some hands on time with this experiment we've decided not to implement this

PR #329, opened 2022-11-09
Fix credentials path format for Windows

brooksmtownsend:
This PR fixes a small issue with NATS configs that require backslashes in Windows paths to be properly escaped with a backslash. 

For example, `C:\Users\tmp\nats.creds` is invalid, NATS is expecting `C:\\Users\\tmp\\nats.creds`.

So you may be wondering, how does this PR fix the issue?
```
-            credentials: "{}"
+           credentials: {:?}
```

Essentially here, instead of quoting the string path that comes out of `path.to_string_lossy()`, we instead use the debug output. This, on unix systems, doesn't change anything other than adding quotes around the edges (hence removing the quotes) and on Windows this escapes the backslashes. I wrote a test to verify this functionality and to ensure we're writing out credsfiles in the proper format. I don't love that it contains newlines and very specific space counts, and would be happy to consider an alternative if desired.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4020,6 +4020,7 @@ dependencies = [
  "claims",
  "command-group",
  "config",
+ "dirs",
  "futures 0.3.25",
  "log",
  "reqwest",
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -34,5 +34,6 @@ serde = { version = "1", features = ["derive"], optional = true}
 serde_json = {version = "1.0.82", optional = true}
 
 [dev-dependencies]
+dirs = "4.0"
 tokio = {version = "1", features = ["full"]}
 claims = "0.7.1"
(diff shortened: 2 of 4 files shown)


connorsmith256:
I'm curious: if you use `.display()` instead of `.to_string_lossy()`, does that also fix this? I will be disappointed if you explicitly have to use `Debug` to correctly format a Windows path 😬

brooksmtownsend:
> I'm curious: if you use `.display()` instead of `.to_string_lossy()`, does that also fix this? I will be disappointed if you explicitly have to use `Debug` to correctly format a Windows path 😬

Nope, I changed it to `.display()` but the escaped slashes didn't change. I was going down the road of creating a custom function to replace `\`'s on Windows with `\\` but landed on this instead

connorsmith256:
Consider me disappointed

autodidaddict (dismissed):

brooksmtownsend:
re-requesting review @autodidaddict / @connorsmith256 , @stevelr I backed out some changes that refactored a test to use the `cmd_lib` library since it was infinite hanging.

autodidaddict (approved):

PR #330, opened 2022-11-09
pass env vars when checking for running host

connorsmith256:
Before this change, attempting to start a wasmcloud host with wash-lib on a system where one or more wasmcloud hosts are already running would fail. Now wash-lib can be successful, so long as env vars like `WASMCLOUD_DASHBOARD_PORT` and `RELEASE_NODE` are set to unique values

Signed-off-by: Connor Smith <connor@cosmonic.com>

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -244,7 +244,12 @@ where
     }
 
     #[cfg(target_family = "unix")]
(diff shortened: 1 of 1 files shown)


autodidaddict (dismissed):

brooksmtownsend (dismissed):
This seems like a great way to get into testing multiple hosts without explicitly providing it as a first-class feature

brooksmtownsend:
@connorsmith256 after #329 is merged the test shouldn't time out anymore 👍🏻

autodidaddict (approved):

PR #333, opened 2022-11-15
Parse version and name from `Cargo.toml` when not provided in `wasmcloud.toml`.

### LINKED ISSUE #332: `version` parameter in wasmcloud.toml should not be required
Currently the `version` parameter in `wasmcloud.toml` is [required](https://github.com/wasmCloud/wash/blob/main/crates/wash-lib/src/parser/mod.rs#L177), which goes against the [mini-spec](https://github.com/wasmCloud/wash/issues/282#issuecomment-1145522739) i wrote in the wasmcloud.toml RFC. This should be changed to optional, and attempt to parse the language's config file version if not present. If the language does not offer a package version parameter (unlikely) and there's none provided in wasmcloud.toml then `wash build` should simply fail.

ghost:
Fixes #332.

This makes both `version` and `name` optional in `wasmcloud.toml`, where if not provided `wash` will attempt to parse them from a language config file. This was easy to do in Rust, but Go doesn't appear to have any standard configuration file with simply a package name and version number. In Go's case we simply fail if the user hasn't provided these two params.

### DIFF
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -1,4 +1,5 @@
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
+use cargo_toml::Manifest;
 use config::Config;
 use semver::Version;
 use std::{fs, path::PathBuf};
@@ -188,7 +189,7 @@ struct RawProjectConfig {
     /// Name of the project.
     pub name: String,
     /// Semantic version of the project.
-    pub version: Version,
+    pub version: Option<Version>,
     pub actor: Option<RawActorConfig>,
     pub provider: Option<RawProviderConfig>,
     pub rust: Option<RawRustConfig>,
@@ -332,12 +333,36 @@ impl TryFrom<RawProjectConfig> for ProjectConfig {
             }
         };
 
+        let version_result: Result<Version> = match raw_project_config.version {
+            Some(version) => Ok(version),
+            None => match language_config {
+                LanguageConfig::Rust(_) => {
+                    let cargo_toml_path = PathBuf::from("Cargo.toml");
+                    if !cargo_toml_path.is_file() {
+                        bail!("No Cargo.toml file found in the current directory");
+                    }
+
+                    Ok(Version::parse(
+                        Manifest::from_path(cargo_toml_path)?
+                            .package
+                            .ok_or_else(|| anyhow!("Missing package in Cargo.toml"))?
+                            .version
+                            .get()?
+                            .as_str(),
+                    )?)
+                }
+                LanguageConfig::TinyGo(_) => {
+                    bail!("Version must be specified in wasmcloud.toml for TinyGo projects.");
+                }
+            },
+        };
+
         Ok(Self {
             language: language_config,
             project_type: project_type_config,
             common: CommonConfig {
                 name: raw_project_config.name,
-                version: raw_project_config.version,
+                version: version_result?,
             },
         })
     }
--- a/crates/wash-lib/src/parser/mod.rs
+++ b/crates/wash-lib/src/parser/mod.rs
@@ -187,7 +187,7 @@ struct RawProjectConfig {
     #[serde(rename = "type")]
     pub project_type: String,
     /// Name of the project.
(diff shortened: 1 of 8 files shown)


crates/wash-lib/src/parser/mod.rs
```diff
@@ -272,73 +273,100 @@ pub fn get_config(opt_path: Option<PathBuf>, use_env: Option<bool>) -> Result<Pr
...
+
+                        let version = match self.version {
+                            Some(version) => version,
+                            None => Version::parse(cargo_toml.version.get()?.as_str())?,
+                        };
+
+                        let name = match self.name {
+                            Some(name) => name,
+                            None => cargo_toml.name,
+                        };
```
brooksmtownsend: This could be consolidated to `self.name.unwrap_or_else(|| cargo_toml.name)`

crates/wash-lib/src/parser/mod.rs
```diff
@@ -272,73 +273,100 @@ pub fn get_config(opt_path: Option<PathBuf>, use_env: Option<bool>) -> Result<Pr
...
+                        }
+
+                        let cargo_toml = Manifest::from_path(cargo_toml_path)?
+                            .package
+                            .ok_or_else(|| anyhow!("Missing package information in Cargo.toml"))?;
+
+                        let version = match self.version {
+                            Some(version) => version,
+                            None => Version::parse(cargo_toml.version.get()?.as_str())?,
+                        };
```
brooksmtownsend, line 346: Same comment re: `unwrap_or_else` but it's a more complex `None` case so I'm fine with leaving it

crates/wash-lib/tests/parser/files/withcargotoml/minimal_rust_actor_with_cargo.toml
```diff
@@ -0,0 +1,5 @@
+language = "rust"
+type = "actor"
+
+[actor]
+claims = ["wasmcloud:httpserver"]
```
brooksmtownsend, line 5: LOOK AT HOW CLEAN IT IS

brooksmtownsend (dismissed):
Left a few nit comments, and feel free to bump to `0.4.0-alpha.2` so we can release this if you'd like 😄 Nothing required

crates/wash-lib/tests/parser/files/withcargotoml/minimal_rust_actor_with_cargo.toml
ghost: 😄

crates/wash-lib/src/parser/mod.rs
ghost: This doesn't actually work with `unwrap_or_else` because `Version::parse` returns a Result and the closure needs to return the same type as the unwrapped value

crates/wash-lib/src/parser/mod.rs
ghost: Done.

crates/wash-lib/tests/parser/files/withcargotoml/minimal_rust_actor_with_cargo.toml
ghost: This makes me think though - do we even need the type field? 🤔

brooksmtownsend: Honestly, if there will always be a necessary field for `[actor/provider/interface]` we may not!

brooksmtownsend (dismissed):

ghost:
Fixed cargo toml conflicts, hopefully good to go now

thomastaylor312:
@mattwilkinsonn I think the test failure was a flake. I re-ran it and hopefully you should be good to merge with a re-approval from @brooksmtownsend

brooksmtownsend (dismissed):

thomastaylor312:
@mattwilkinsonn Feel free to just merge after rebase

ghost:
rebased, probably needs another approval @brooksmtownsend

brooksmtownsend (approved):

PR #337, opened 2022-11-16
feat(lib): Adds drain command to wash lib

thomastaylor312:
This also starts the process of creating a `config` module that I'll continue to update as I push forward the other PRs. Please note that this is the first of many PRs. I plan on doing each command as a separate PR instead of a mega PR

### DIFF
--- a/crates/wash-lib/src/config.rs
+++ b/crates/wash-lib/src/config.rs
@@ -7,7 +7,7 @@ use std::{
 
 const WASH_DIR: &str = ".wash";
 
-pub const DOWNLOADS_DIR: &str = "downloads";
+const DOWNLOADS_DIR: &str = "downloads";
 pub const DEFAULT_NATS_HOST: &str = "127.0.0.1";
 pub const DEFAULT_NATS_PORT: &str = "4222";
 pub const DEFAULT_LATTICE_PREFIX: &str = "default";
@@ -33,10 +33,12 @@ pub fn cfg_dir() -> IoResult<PathBuf> {
     Ok(wash)
 }
 
+/// Returns the path to the caching directory for smithy files
 pub fn model_cache_dir() -> IoResult<PathBuf> {
     weld_codegen::weld_cache_dir().map_err(|e| Error::new(ErrorKind::Other, e.to_string()))
 }
 
+/// The path to the downloads directory for wash
 pub fn downloads_dir() -> IoResult<PathBuf> {
     cfg_dir().map(|p| p.join(DOWNLOADS_DIR))
 }
--- a/crates/wash-lib/src/config.rs
+++ b/crates/wash-lib/src/config.rs
@@ -1,0 +1,42 @@
+//! Common config constants and functions for loading, finding, and consuming configuration data
+use std::{
+    fs,
+    io::{Error, ErrorKind, Result as IoResult},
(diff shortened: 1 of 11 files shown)


crates/wash-lib/src/config.rs
```diff
@@ -0,0 +1,42 @@
...
+    Ok(wash)
+}
+
+pub fn model_cache_dir() -> IoResult<PathBuf> {
+    weld_codegen::weld_cache_dir().map_err(|e| Error::new(ErrorKind::Other, e.to_string()))
+}
+
+pub fn downloads_dir() -> IoResult<PathBuf> {
+    cfg_dir().map(|p| p.join(DOWNLOADS_DIR))
+}
```
brooksmtownsend: These should have a quick blurb about what these directories are used for

crates/wash-lib/src/drain.rs
```diff
@@ -0,0 +1,97 @@
+use std::{env, fs, io::Result, path::PathBuf};
+
+use crate::config::{downloads_dir, model_cache_dir};
+
+/// A type that allows you to clean up (i.e. drain) a set of caches and folders used by wasmcloud
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "cli", derive(clap::Subcommand))]
```
brooksmtownsend, line 7: niiiiice

crates/wash-lib/src/drain.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Remove cached binaries extracted from provider archives
+    Lib,
+    /// Remove cached smithy files downloaded from model urls
+    Smithy,
+    /// Remove downloaded and generated files from launching wasmCloud hosts
+    Downloads,
+}
+
+// Drain::All.drain()
+// drain(DrainType::All)
```
brooksmtownsend: I think these are leftover from the stream

crates/wash-lib/src/lib.rs
```diff
@@ -8,3 +8,6 @@ pub mod start;
 
 #[cfg(feature = "parser")]
 pub mod parser;
+
+pub mod config;
+pub mod drain;
```
brooksmtownsend, line 13: Thoughts on continuing the feature flag gating individual modules? I know that one per-module is probably overkill, but we could gate `drain` behind a flag. Config seems lightweight and applicable to most installation things

brooksmtownsend (changes requested):

crates/wash-lib/src/config.rs
thomastaylor312: Oh whoops. Forgot the doc comments

crates/wash-lib/src/drain.rs
thomastaylor312: Yes, yes they are

crates/wash-lib/src/lib.rs
thomastaylor312: My plan was to add those gates at the _veeeerrrry_ end. It is a bit hard to tell where those gates should be until we get all done

brooksmtownsend: Okay that sounds good to me, good call

brooksmtownsend (approved):

PR #339, opened 2022-11-17
ref(context): Creates new context library

thomastaylor312:
This creates a new context library with some extendable traits for managing contexts as well as a fully featured module for handling context on disk.


crates/wash-lib/src/context/fs.rs
```diff
@@ -0,0 +1,334 @@
...
+    fn save_context(&self, ctx: &WashContext) -> Result<()> {
+        let filepath = context_path_from_name(&self.0, &ctx.name);
+        let file = std::fs::File::create(filepath)?;
+        serde_json::to_writer(file, ctx).map_err(anyhow::Error::from)
+    }
+
+    fn delete_context(&self, name: &str) -> Result<()> {
+        let path = context_path_from_name(&self.0, name);
+        std::fs::remove_file(path).map_err(anyhow::Error::from)
+    }
```
brooksmtownsend: Kind of a weird edge case, but if we delete the default context we should probably unset that default. If the behavior here is such that the default just resets to the `host_config` context, then that's fine

crates/wash-lib/src/context/fs.rs
```diff
@@ -0,0 +1,334 @@
...
+fn context_path_from_name(dir: impl AsRef<Path>, name: &str) -> PathBuf {
+    dir.as_ref().join(format!("{}.json", name))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn round_trip_happy_path() {
```
brooksmtownsend, line 178: tests <3

brooksmtownsend (dismissed):
Clean 🚀

crates/wash-lib/src/context/fs.rs
thomastaylor312: I thought about this but was trying to model current behavior. Let me try doing it this way instead

thomastaylor312: Ok added this

brooksmtownsend (approved):

PR #344, opened 2022-11-21
feat(lib): Adds new keys module to wash-lib

thomastaylor312:
Please note that this introduces one small breaking change to output that removes the `.nk` suffix from the list of keys. However, there is backward compatibility for providing <key_name>.nk to `wash keys get` so it will still function as it did previously. This change was specifically made because the key name is more important than the suffix. If desired, I can back out that change, but it seemed to make more sense to make it less like a wash-specific `ls` of a directory

### DIFF
--- a/crates/wash-lib/src/keys/fs.rs
+++ b/crates/wash-lib/src/keys/fs.rs
@@ -1,0 +1,251 @@
+//! A filesystem directory based implementation of a `KeyManager`
+
+use std::{
+    ops::Deref,
+    path::{Path, PathBuf},
+};
+
+use anyhow::Result;
+use nkeys::KeyPair;
+
+use super::KeyManager;
+
+pub const KEY_FILE_EXTENSION: &str = "nk";
+
+pub struct KeyDir(PathBuf);
+
+impl AsRef<Path> for KeyDir {
+    fn as_ref(&self) -> &Path {
+        &self.0
+    }
+}
+
+impl Deref for KeyDir {
+    type Target = Path;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl KeyDir {
+    /// Creates a new KeyDir, erroring if it is unable to access or create the given directory.
+    pub fn new(path: impl AsRef<Path>) -> Result<KeyDir> {
+        let p = path.as_ref();
+        let exists = p.exists();
+        if exists && !p.is_dir() {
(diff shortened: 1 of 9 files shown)


crates/wash-lib/src/keys/fs.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    let metadata = path.as_ref().metadata()?;
+    match metadata.file_type().is_dir() {
+        true => std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o700))?,
+        false => std::fs::set_permissions(path, std::fs::Permissions::from_mode(0o600))?,
+    };
+    Ok(())
+}
+
+#[cfg(target_os = "windows")]
```
brooksmtownsend, line 140: In the `start` code I used `target_family = "unix"` and `target_family = "windows"`, does this change the target at all or does it encapsulate the same differences?

crates/wash-lib/src/keys/fs.rs
```diff
@@ -0,0 +1,251 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod test {
+    use nkeys::KeyPairType;
+
+    use super::*;
+
+    const TEST_KEY: &str = "SMAAGJ4DY4FNV4VJWA6QU7UQIL7DKJR4Z3UH7NBMNTH22V6VEIJGJUBQN4";
```
brooksmtownsend, line 151: Hope you legitimately generated this one

tests/integration_par.rs
```diff
@@ -62,6 +62,7 @@ fn integration_par_create(issuer: &str, subject: &str, archive: &str) {
         ])
         .output()
         .expect("failed to create provider archive file");
+    println!("Output: {:?}", create);
```
brooksmtownsend: not a big deal because test but this is probs left over ;)

crates/wash-lib/src/keys/fs.rs
thomastaylor312: This was copy pasted over almost verbatim. I can change if you'd like

tests/integration_par.rs
thomastaylor312: Oh yep, will remove

crates/wash-lib/src/keys/fs.rs
brooksmtownsend: If it ain't broke, don't fix it

brooksmtownsend (approved):

PR #345, opened 2022-11-22
feat(lib): Moves claims and registry code into wash lib

thomastaylor312:
Sorry for the big PR here, but due to a bunch of codependent code,
I had to move a bunch of stuff at once. There are two main threads
to this PR. First, I noticed that the claims code is all CLI specific,
but it is likely that anyone building a CLI will not want to rewrite that
again. If you are doing this purely in code, you can just use the
wascap library. To make this work, I started added the CLI specific stuff
to the `cli` module of wash lib. There will probably be other things we
add to it as we finish this refactor

Second, this moves the reusable registry bits into its own module, which
is super handy even for those not doing a CLI as it avoids direct
interaction with the lower level OCI crates

### DIFF
--- a/crates/wash-lib/src/registry.rs
+++ b/crates/wash-lib/src/registry.rs
@@ -87,7 +87,7 @@ pub async fn get_oci_artifact(
 
 /// Pull down the artifact from the given url and additional options
 pub async fn pull_oci_artifact(url: String, options: OciPullOptions) -> Result<Vec<u8>> {
-    let image: Reference = url.parse()?;
+    let image: Reference = url.to_lowercase().parse()?;
 
     if image.tag().unwrap_or("latest") == "latest" && !options.allow_latest {
         bail!("Pulling artifacts with tag 'latest' is prohibited.");
@@ -143,7 +143,7 @@ pub async fn push_oci_artifact(
     artifact: impl AsRef<Path>,
     options: OciPushOptions,
 ) -> Result<()> {
-    let image: Reference = url.parse()?;
+    let image: Reference = url.to_lowercase().parse()?;
 
     if image.tag().unwrap() == "latest" && !options.allow_latest {
         bail!("Pushing artifacts with tag 'latest' is prohibited");
--- a/crates/wash-lib/src/registry.rs
+++ b/crates/wash-lib/src/registry.rs
@@ -1,0 +1,238 @@
+//! Utilities for pulling and pushing artifacts to various registries
+
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
+
+use anyhow::{anyhow, bail, Result};
+use oci_distribution::manifest::OciImageManifest;
+use oci_distribution::{client::*, secrets::RegistryAuth, Reference};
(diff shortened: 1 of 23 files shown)


thomastaylor312:
This is dependent on #344 being merged. If you want to start reviewing, you can filter down to the relevant commits

crates/wash-lib/src/registry.rs
```diff
@@ -0,0 +1,238 @@
...
+            cached_artifact.read_to_end(&mut buf).await?;
+            return Ok(buf);
+        }
+    }
+    pull_oci_artifact(url_or_file, options).await
+}
+
+/// Pull down the artifact from the given url and additional options
+pub async fn pull_oci_artifact(url: String, options: OciPullOptions) -> Result<Vec<u8>> {
+    let image: Reference = url.parse()?;
```
brooksmtownsend: This may not have made it in from the other PR, but this URL needs to be `.to_lowercase` or `.to_ascii_lowercase`'ed so it doesn't fail the parse

crates/wash-lib/src/registry.rs
```diff
@@ -0,0 +1,238 @@
...
+        .flat_map(|l| l.data.clone())
+        .collect::<Vec<_>>())
+}
+
+pub async fn push_oci_artifact(
+    url: String,
+    artifact: impl AsRef<Path>,
+    options: OciPushOptions,
+) -> Result<()> {
+    let image: Reference = url.parse()?;
```
brooksmtownsend: Same comment re: lowercasing

src/keys.rs
```diff
@@ -77,40 +73,27 @@ pub(crate) fn generate(kt: &KeyPairType) -> Result<CommandOutput> {
...
+        .ok_or_else(|| anyhow::anyhow!("Key {} doesn't exist", keyname))?;
 
-    Ok(CommandOutput::from_key_and_text("seed", seed.trim()))
+    Ok(CommandOutput::from_key_and_text("seed", key.seed()?))
 }
 
 /// Lists all keypairs (file extension .nk) in a specified directory or $WASH_KEYS($HOME/.wash/keys) if directory is not specified
 pub(crate) fn list(directory: Option<PathBuf>) -> Result<CommandOutput> {
-    let dir = determine_directory(directory)?;
+    let key_dir = KeyDir::new(determine_directory(directory)?)?;
```
brooksmtownsend: Will probably want to do a rebase after #344 to make sure nothing weird happens

tests/integration_par.rs
```diff
@@ -62,6 +62,7 @@ fn integration_par_create(issuer: &str, subject: &str, archive: &str) {
         ])
         .output()
         .expect("failed to create provider archive file");
+    println!("Output: {:?}", create);
```
brooksmtownsend: leftover print ;)

brooksmtownsend (changes requested):
Just a request to carry over the changes from #340 and rebase after the keys PR is merged

brooksmtownsend (approved):

PR #353, opened 2022-12-06
Moved project build functionality to wash-lib

brooksmtownsend:
This PR moves majority of the `build` module's functionality into `wash-lib`, exposing two functions:
- `build_project` that simply takes a parsed wasmcloud.toml and builds based on the project
- `build_actor` that takes actor, language, and common config to build an actor (along with another option `no_sign`). I removed `push` as an option since we may not implement it at all, and it would be a breaking change to add either way.

I also chose to gate this behind the `cli` field for now to safe effort on refactoring `cli/claims.rs` to a different module. The only reason why this is behind the `cli` flag anyways is to reuse the `sign_file` function.

This PR also adds in a few more flags into the `wash build` command which allow you to sign an actor/provider with provided issuer or subject keys, just like `wash claims sign` allows

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -1,0 +1,205 @@
+//! Build (and sign) a wasmCloud actor, provider, or interface. Depends on the "cli" feature
+//!
+
+use std::{fs, path::PathBuf, process, str::FromStr};
+
+use anyhow::{anyhow, bail, Result};
+
+use crate::cli::{
+    claims::{sign_file, ActorMetadata, SignCommand},
+    OutputKind,
+};
+use crate::parser::{
+    ActorConfig, CommonConfig, InterfaceConfig, LanguageConfig, ProjectConfig, ProviderConfig,
+    RustConfig, TinyGoConfig, TypeConfig,
+};
+
+/// Using a [ProjectConfig], usually parsed from a `wasmcloud.toml` file, build the project
+/// with the installed language toolchain. This will delegate to [build_actor] when the project is an actor,
+/// or return an error when trying to build providers or interfaces. This functionality is planned in a future release.
+///
+/// This function returns the path to the compiled artifact, a signed Wasm module, signed provider archive, or compiled
+/// interface library file.
+///
+/// # Usage
+/// ```no_run
+/// use wash_lib::{build::build_project, parser::get_config};
+/// let config = get_config(None, Some(true))?;
+/// let artifact_path = build_project(config)?;
+/// println!("Here is the signed artifact: {}", artifact_path.to_string_lossy());
+/// ```
+pub fn build_project(config: ProjectConfig) -> Result<PathBuf> {
+    match config.project_type {
+        TypeConfig::Actor(actor_config) => {
+            build_actor(actor_config, config.language, config.common, false)
+        }
+        TypeConfig::Provider(_provider_config) => Err(anyhow!(
+            "wash build has not be implemented for providers yet. Please use `make` for now!"
+        )),
+        TypeConfig::Interface(_interface_config) => Err(anyhow!(
+            "wash build has not be implemented for interfaces yet. Please use `make` for now!"
+        )),
+    }
+}
+
+/// Builds a wasmCloud actor using the installed language toolchain, then signs the actor with
+/// keys, capability claims, and additional friendly information like name, version, revision, etc.
+///
+/// # Arguments
+/// * `actor_config`: [ActorConfig] for required information to find, build, and sign an actor
+/// * `language_config`: [LanguageConfig] specifying which language the actor is written in
+/// * `common_config`: [CommonConfig] specifying common parameters like [CommonConfig::name] and [CommonConfig::version]
+/// * `no_sign`: If `true`, build the actor but don't sign. Useful for just checking build status without needing signing keys or fully referenced capabilities
+pub fn build_actor(
+    actor_config: ActorConfig,
+    language_config: LanguageConfig,
+    common_config: CommonConfig,
+    no_sign: bool,
+) -> Result<PathBuf> {
+    // Build actor based on language toolchain
+    let file_path = match language_config {
+        LanguageConfig::Rust(rust_config) => {
+            build_rust_actor(common_config.clone(), rust_config, actor_config.clone())
+        }
+        LanguageConfig::TinyGo(tinygo_config) => {
+            build_tinygo_actor(common_config.clone(), tinygo_config)
+        }
+    }?;
+
+    // Exit early if signing isn't desired
+    if no_sign {
+        Ok(file_path)
+    } else {
+        let source = file_path
+            .to_str()
+            .ok_or_else(|| anyhow!("Could not convert file path to string"))?
+            .to_string();
+
+        let destination = format!("build/{}_s.wasm", common_config.name);
+        let destination_file = PathBuf::from_str(&destination);
+
+        let sign_options = SignCommand {
+            source,
+            destination: Some(destination),
+            metadata: ActorMetadata {
+                name: common_config.name,
+                ver: Some(common_config.version.to_string()),
+                custom_caps: actor_config.claims,
+                call_alias: actor_config.call_alias,
+                ..Default::default()
+            },
+        };
+        sign_file(sign_options, OutputKind::Json)?;
+
+        Ok(destination_file?)
+    }
+}
+
+/// Builds a rust actor and returns the path to the file.
+fn build_rust_actor(
+    common_config: CommonConfig,
+    rust_config: RustConfig,
+    actor_config: ActorConfig,
+) -> Result<PathBuf> {
+    let mut command = match rust_config.cargo_path {
+        Some(path) => process::Command::new(path),
+        None => process::Command::new("cargo"),
+    };
+
+    let result = command.args(["build", "--release"]).status()?;
+
+    if !result.success() {
+        bail!("Compiling actor failed: {}", result.to_string())
+    }
+
+    let wasm_file = PathBuf::from(format!(
+        "{}/{}/release/{}.wasm",
+        rust_config
+            .target_path
+            .unwrap_or_else(|| PathBuf::from("target"))
+            .to_string_lossy(),
+        actor_config.wasm_target,
+        common_config.name,
+    ));
+
+    if !wasm_file.exists() {
+        bail!(
+            "Could not find compiled wasm file to sign: {}",
+            wasm_file.display()
+        );
+    }
+
+    // move the file out into the build/ folder for parity with tinygo and convienience for users.
+    let copied_wasm_file = PathBuf::from(format!("build/{}.wasm", common_config.name));
+    if let Some(p) = copied_wasm_file.parent() {
+        fs::create_dir_all(p)?;
+    }
+    fs::copy(&wasm_file, &copied_wasm_file)?;
+    fs::remove_file(&wasm_file)?;
+
+    Ok(copied_wasm_file)
+}
+
+/// Builds a tinygo actor and returns the path to the file.
+fn build_tinygo_actor(common_config: CommonConfig, tinygo_config: TinyGoConfig) -> Result<PathBuf> {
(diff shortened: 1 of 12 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,205 @@
...
+/// # Arguments
+/// * `actor_config`: [ActorConfig] for required information to find, build, and sign an actor
+/// * `language_config`: [LanguageConfig] specifying which language the actor is written in
+/// * `common_config`: [CommonConfig] specifying common parameters like [CommonConfig::name] and [CommonConfig::version]
+/// * `no_sign`: If `true`, build the actor but don't sign. Useful for just checking build status without needing signing keys or fully referenced capabilities
+pub fn build_actor(
+    actor_config: ActorConfig,
+    language_config: LanguageConfig,
+    common_config: CommonConfig,
+    no_sign: bool,
```
connorsmith256: Nit: `skip_signing`? `build_only`?

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,205 @@
...
+                name: common_config.name,
+                ver: Some(common_config.version.to_string()),
+                custom_caps: actor_config.claims,
+                call_alias: actor_config.call_alias,
+                ..Default::default()
+            },
+        };
+        sign_file(sign_options, OutputKind::Json)?;
+
+        Ok(destination_file?)
```
connorsmith256: Should we check that `destination_file` is `Ok` before trying to sign, rather than down here?

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,205 @@
...
+) -> Result<PathBuf> {
+    let mut command = match rust_config.cargo_path {
+        Some(path) => process::Command::new(path),
+        None => process::Command::new("cargo"),
+    };
+
+    let result = command.args(["build", "--release"]).status()?;
+
+    if !result.success() {
+        bail!("Compiling actor failed: {}", result.to_string())
```
connorsmith256, line 134: Should a library `bail!`? Feels a bit extreme. Why not return an error and let the caller handle it?

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,205 @@
...
+/// interface library file.
+///
+/// # Usage
+/// ```no_run
+/// use wash_lib::{build::build_project, parser::get_config};
+/// let config = get_config(None, Some(true))?;
+/// let artifact_path = build_project(config)?;
+/// println!("Here is the signed artifact: {}", artifact_path.to_string_lossy());
+/// ```
+pub fn build_project(config: ProjectConfig) -> Result<PathBuf> {
```
connorsmith256: Could this take a ref instead of moving the struct?

crates/wash-lib/src/build.rs
brooksmtownsend: That's what this does 😄 https://docs.rs/anyhow/1.0.66/anyhow/macro.bail.html super useful from `anyhow`

connorsmith256: 🤦 I totally thought this was essentially a prettier panic. TIL

crates/wash-lib/src/build.rs
brooksmtownsend: Definitely, I'll throw some references around

crates/wash-lib/src/build.rs
brooksmtownsend: Yeah I like affirmative instead of the double negative. Will reword to something

crates/wash-lib/src/build.rs
brooksmtownsend: Yeah that's pretty fair, I was thinking that if I failed to create the pathbuf no need to prevent me from signing the file, but I'll just move the `?` upwards 😄

crates/wash-lib/src/build.rs
```diff
@@ -14,6 +15,37 @@ use crate::parser::{
...
 
+// This struct requires the `cli` feature, which this whole module is gated by. If that changes,
+// this struct needs to be adjusted with the derive macros
+#[derive(Parser, Debug, Clone)]
+pub struct SignConfig {
+    /// Location of key files for signing. Defaults to $WASH_KEYS ($HOME/.wash/keys)
+    #[clap(long = "keys-directory", env = "WASH_KEYS", hide_env_values = true)]
+    pub keys_directory: Option<PathBuf>,
+
+    /// Path to issuer seed key (account). If this flag is not provided, the will be sourced from $WASH_KEYS ($HOME/.wash/keys) or generated for you if it cannot be found.
```
connorsmith256: Nit: typo in "the will be"? Same with the comment on `subject`

crates/wash-lib/src/build.rs
```diff
@@ -51,61 +86,65 @@ pub fn build_project(config: ProjectConfig) -> Result<PathBuf> {
 /// * `common_config`: [CommonConfig] specifying common parameters like [CommonConfig::name] and [CommonConfig::version]
 /// * `no_sign`: If `true`, build the actor but don't sign. Useful for just checking build status without needing signing keys or fully referenced capabilities
```
connorsmith256: Nit: this comment needs to be updated

crates/wash-lib/src/parser/mod.rs
```diff
@@ -237,23 +239,30 @@ pub fn get_config(opt_path: Option<PathBuf>, use_env: Option<bool>) -> Result<Pr
...
-        path = wasmcloud_path;
-    }
-
-    if !path.is_file() {
-        return Err(anyhow!("No config file found at {}", path.display()));
-    }
-
-    let mut config = Config::builder().add_source(config::File::from(path.clone()));
+        (path, wasmcloud_path)
+    } else if path.is_file() {
```
connorsmith256, line 251: Do we need to double-check that this file is a `wasmcloud.toml` ?

src/build.rs
```diff
@@ -1,40 +1,36 @@
...
 use wash_lib::cli::CommandOutput;
 use wash_lib::parser::{get_config, TypeConfig};
 
 /// Build (and sign) a wasmCloud actor, provider, or interface
 #[derive(Debug, Parser, Clone)]
 #[clap(name = "build")]
 pub(crate) struct BuildCommand {
-    /// If set, pushes the signed actor to the registry.
-    #[clap(short = 'p', long = "push")]
-    pub(crate) push: bool,
```
connorsmith256, line 22: Is this a breaking change, or was this functionality not implemented yet?

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,252 @@
+//! Build (and sign) a wasmCloud actor, provider, or interface. Depends on the "cli" feature
+//!
```
thomastaylor312: Super nit: Trailing empty comment

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,252 @@
...
+use crate::cli::{
+    claims::{sign_file, ActorMetadata, SignCommand},
+    OutputKind,
+};
+use crate::parser::{
+    ActorConfig, CommonConfig, InterfaceConfig, LanguageConfig, ProjectConfig, ProviderConfig,
+    RustConfig, TinyGoConfig, TypeConfig,
+};
+
+// This struct requires the `cli` feature, which this whole module is gated by. If that changes,
```
thomastaylor312: Did we want to move over all of these `clap` things before we test out whether we can configure `env` variable names?

src/build.rs
```diff
@@ -1,40 +1,36 @@
...
-                config.language,
-                config.common,
-                command.no_sign,
-            )?;
+        TypeConfig::Actor(ref _actor_config) => {
+            let actor_path = build_project(&config, Some(command.signing_config))?;
             let json_output = HashMap::from([
                 ("actor_path".to_string(), json!(actor_path)),
-                ("signed".to_string(), json!(!command.no_sign)),
+                ("signed".to_string(), json!(true)),
```
connorsmith256: Just double-checking I understand this: now `wash build` will always sign actors, though wash-lib supports not signing (by passing None for the signing config to `build_project`)?

crates/wash-lib/tests/parser/main.rs
```diff
@@ -42,6 +42,9 @@ fn rust_actor() {
         CommonConfig {
             name: "testactor".to_string(),
             version: Version::parse("0.1.0").unwrap(),
+            path: PathBuf::from("./tests/parser/files/")
+                .canonicalize()
```
thomastaylor312, line 46: Nit: Do you need canonicalize since this is a test?

src/build.rs
connorsmith256: Looks like the integration test failed due to something around `no-sign`, but idk if that was on outdated code, since I see you did update the test...

connorsmith256 (changes requested):
Just a couple nits and questions

thomastaylor312 (dismissed):
All good pending passing tests

crates/wash-lib/tests/parser/main.rs
brooksmtownsend: Actually, yeah, but the real reason why I do it is to turn this into the absolute path instead of `./tests/...`

crates/wash-lib/src/build.rs
brooksmtownsend: I don't think so, I'm going to move them into the CLI section and transform this struct into something that doesn't have the derive macros

src/build.rs
brooksmtownsend: Yeah I did unintentionally remove that ability, adding back in a `--build-only` flag

src/build.rs
brooksmtownsend: This functionality wasn't implemented yet, so I figured that implementing it was going to be a breaking change anyways and it was better to be left off. There are plenty of other breaking changes here, though!

crates/wash-lib/src/parser/mod.rs
brooksmtownsend: I don't think so, we wouldn't want to enforce it needing to be called `wasmcloud.toml` (though I think that's fine.) IMO failing here at the parse step is more appropriate

crates/wash-lib/src/build.rs
brooksmtownsend: Moved in latest commit

thomastaylor312 (dismissed):

src/build.rs
```diff
@@ -1,241 +1,96 @@
...
-) -> Result<CommandOutput> {
-    // build it
-    println!("Building actor...");
-    let file_path = match language_config {
-        LanguageConfig::Rust(rust_config) => {
-            build_rust_actor(common_config.clone(), rust_config, actor_config.clone())
+        TypeConfig::Actor(ref _actor_config) => {
+            let actor_path = build_project(
+                &config,
+                if command.build_only {
```
ghost: Could you set this second param to a variable first? Gets a bit nested here inside the function call

crates/wash-lib/src/build.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    /// Path to issuer seed key (account). If this flag is not provided, the seed will be sourced from ($HOME/.wash/keys) or generated for you if it cannot be found.
+    pub issuer: Option<String>,
+
+    /// Path to subject seed key (module or service). If this flag is not provided, the seed will be sourced from ($HOME/.wash/keys) or generated for you if it cannot be found.
+    pub subject: Option<String>,
+
+    /// Disables autogeneration of keys if seed(s) are not provided
+    pub disable_keygen: bool,
+}
```
ghost, line 30: Should this be a part of `wasmcloud.toml`?

ghost (dismissed):
LGTM with one nit and one question

crates/wash-lib/src/build.rs
brooksmtownsend: I don't think so, I think this is just part of runtime command line flags for when you are building the actor in CI and you don't want autogenerated keys, but I could be convinced otherwise if you have a use case in mind

connorsmith256 (approved):

PR #354, opened 2022-12-14
fixed 352, added js_domain to context

### LINKED ISSUE #352: [BUG] `wash` does not properly parse the host_config context lattice prefix
**Describe the bug**
When you launch a wasmCloud host locally, the `host_config` used to create that host is also placed in `~/.wash/host_config.json`. This is used to parse the `host_config` context, which essentially serves as a way to immediately be able to query a local wasmCloud host without specifying new flags. The connection ports and host parse well, but `ctl_lattice_prefix` and `rpc_lattice_prefix` stay as "default" even if you start a host with a different prefix.

**To Reproduce**
1. Launch a host with `wash up -x notdefault -d`
1. After a few seconds, run `wash ctl get hosts`
1. See no hosts
1. Run `wash ctl get hosts -x notdefault`
1. See your one host

**Expected behavior**
After launching a host with a lattice prefix that isn't `default` locally, I expect the host_config context to include that lattice prefix.

brooksmtownsend:
Fixes #352 

This PR also adds parsing the `js_domain` to the Context, which will be useful after https://github.com/wasmCloud/wasmcloud-otp/pull/513 and the control interface client has Jetstream domain support

### DIFF
--- a/src/ctx/mod.rs
+++ b/src/ctx/mod.rs
@@ -360,7 +360,7 @@ fn prompt_for_context() -> Result<WashContext> {
     )?;
 
     let js_domain = match user_question(
-        "What JetStream domain will th host be running, if any?",
+        "What JetStream domain will the host be running, if any?",
         &Some("".to_string()),
     ) {
         Ok(s) if s.is_empty() => None,
--- a/src/ctx/mod.rs
+++ b/src/ctx/mod.rs
@@ -311,7 +311,7 @@ fn prompt_for_context() -> Result<WashContext> {
 
     let cluster_seed = match user_question(
         "What cluster seed do you want to use to sign invocations?",
-        &Some("".to_string()),
+        &Some(String::new()),
     ) {
         Ok(s) if s.is_empty() => None,
         Ok(s) => Some(s.parse::<ClusterSeed>()?),
@@ -327,7 +327,7 @@ fn prompt_for_context() -> Result<WashContext> {
     )?;
     let ctl_jwt = match user_question(
         "Enter your JWT that you use to authenticate to the control interface connection, if applicable",
-        &Some("".to_string()),
+        &Some(String::new()),
     ) {
         Ok(s) if s.is_empty() => None,
         Ok(s) => Some(s),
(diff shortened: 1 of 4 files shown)


src/ctx/mod.rs
```diff
@@ -353,10 +353,21 @@ fn prompt_for_context() -> Result<WashContext> {
...
-    let ctl_lattice_prefix = user_question(
-        "What is the control interface connection lattice prefix?",
+
+    let lattice_prefix = user_question(
+        "What is the lattice prefix that the host will communicate on?",
         &Some(DEFAULT_LATTICE_PREFIX.to_string()),
     )?;
+
+    let js_domain = match user_question(
+        "What JetStream domain will th host be running, if any?",
```
thomastaylor312: Spelling: s/th/the

src/ctx/mod.rs
```diff
@@ -353,10 +353,21 @@ fn prompt_for_context() -> Result<WashContext> {
...
+        &Some("".to_string()),
```
thomastaylor312: Nit: I think it reads better (and is more idiomatic) to do `String::new()` instead of what you have here

thomastaylor312 (changes requested):
One required spelling change (since it is output). Also, isn't there a reason we have both an RPC and a ctl prefix or was that just a mistake?

src/ctx/mod.rs
connorsmith256: Is there a difference between `Some(String::new())` and `None` here?

thomastaylor312: Actually, good point. Probably better to just use `None` if possible

brooksmtownsend: Yeah None works 👍🏻

brooksmtownsend: Actually, and don't shoot me on this one, supplying `None` as the default value for some reason requires the user to input something and they can't accept the default of nothing. Changed them to `String::new()` for idiomatism

brooksmtownsend:
> Also, isn't there a reason we have both an RPC and a ctl prefix or was that just a mistake?

@thomastaylor312 
Funny enough, that was just a mistake when originally writing the CTX module. We don't support using different lattice prefixes for different NATS connections in the host

thomastaylor312 (approved):

PR #355, opened 2022-12-14
Moved generate module to wash-lib

brooksmtownsend:
This PR moves the `generate` module to `wash-lib` exposing one main function `generate_project`. I tried to make this very similar to the `build` module in that consumers of this module gather their **Project** information and can call this function to build the project. This module is gated behind the `cli` feature.


crates/wash-lib/Cargo.toml
```diff
@@ -25,29 +25,39 @@ async-compression = { version = "0.3", default-features = false, features = ["to
 clap = { version = "4", features = ["derive", "env"], optional = true }
 command-group = { version = "1.0.8", features = ["with-tokio"] }
 config = { version = "0.13.1", features = ["toml"], optional = true }
+console = "0.15"
```
thomastaylor312: All of these specific generate things should be `optional = true` and enabled when `cli` is specified (unless it is used elsewhere)

thomastaylor312 (changes requested):
Code looks all good, just some stuff around deps

connorsmith256 (dismissed):

connorsmith256 (approved):

thomastaylor312 (approved):

PR #361, opened 2023-01-04
chore: bump wascap

ricochet:


autodidaddict:
I'm okay with this if @brooksmtownsend and @mattwilkinsonn are.

autodidaddict:
@ricochet - the DCO requirement needs the commits to be signed

for the curious - this change is significant because anything signed by this version of wash/washlib will _not_ be backwards compatible with old versions of the host (the module hash verification check will fail)

autodidaddict:
This seems to have broken an integration test. I'm not sure if it's because one of the fixtures is different but there's a claims mismatch

ricochet:
Ah, this is a good test. 

wash claims inspect output is different. The hashes changed (expected), there is now a version on the claim (expected) but two other diffs look suspect:

```json
{
...
        "capabilities": ["HTTP Server", "Logging"],
        "call_alias": "wasmcloud/logger_onedotzero"
 }
```

```json
{
...
        "capabilities": ["HTTP Server", "wasmcloud:logging"],
         "call_alias": "(Not set)"
}
```

It seems to have lost the call_alias and the logging claim name is odd. I think I prefer the wasmcloud prefix, but why didn't the HTTP Server change? Digging in...

ricochet:
I think there must be history around naming as this claim is for "wasmcloud:logging" and not "wasmcloud:builtin:logging". "Logging" should now == "wasmcloud:builtin:logging" and not "wasmcloud:logging".

This could be avoided if wascap normalized friendly name capabilities or if friendly names were removed entirely.

autodidaddict:
Yep we're looking at some scary tissue here from early days when there were a handful of predetermined contract IDs. The upside is this is mostly showing the age in the test and not a big problem with the app itself. 

100% agree that we should be consistent in how we treat this

ricochet:
I think that mostly leaves call_alias as the key change that I can't explain. I added a test over in wascap and locally also tried with the same string as the one here with success.

I manually ran the integration test with a wash before/after wascap update... and yeah call_alias is gone Jim.

autodidaddict:
Doing some digging on this. I've verified that doing this manually with the previous/released version of wash works just fine. Looking into some more things

autodidaddict:
@ricochet if you bump wascap to 0.9.1 now the double-sign test should pass.

Cargo.toml
```diff
@@ -50,7 +50,7 @@ tokio = {version = "1", features = ["full"]}
 toml = "0.5"
 which = "4.2.2"
 wash-lib = { version = "0.4.0-alpha.2", path = "./crates/wash-lib", features = ["cli"] }
-wascap = "0.8.0"
+wascap = "0.9.0"
```
autodidaddict: this should be 0.9.1 as the washlib one is also using 0.9.1.

autodidaddict (changes requested):

Cargo.toml
ricochet: cargo update betraying me. TY

autodidaddict (dismissed):

autodidaddict (dismissed):

connorsmith256 (dismissed):
Woohoo, tests pass now! 🎉

brooksmtownsend (dismissed):

connorsmith256 (approved):

PR #362, opened 2023-01-05
fix: change test reg port

ricochet:
Macs are quirky and it's best to avoid ports 5000 and 7000
https://developer.apple.com/forums/thread/682332

### DIFF
--- a/tools/docker-compose.yml
+++ b/tools/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   registry:
     image: registry:2
     ports:
-     - "5000:5000"
+     - "5001:5000"
   nats:
     image: nats:2.1.9
     ports:
--- a/src/reg.rs
+++ b/src/reg.rs
@@ -255,7 +255,7 @@ mod tests {
     use clap::Parser;
 
     const ECHO_WASM: &str = "wasmcloud.azurecr.io/echo:0.2.0";
-    const LOCAL_REGISTRY: &str = "localhost:5000";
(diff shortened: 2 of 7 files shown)


thomastaylor312 (approved):

thomastaylor312:
@ricochet Code looks good but integration test is failing

tools/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
   registry:
     image: registry:2
     ports:
-     - "5000:5000"
+     - "5001:5000"
```
autodidaddict, line 6: If we're trying to avoid mac conflicts with 5000, should this be `5001:5001` ? Please disregard if I'm just not understanding the docker port mapping implications.

brooksmtownsend: I think it should be `5001:5000`, since the registry in the container is listening on 5000 but we want to avoid the port conflict on the host machine

brooksmtownsend: Unless I have that mapping backwards. I feel like I'm always close to having them mixed up

autodidaddict: Same. I always forget the order.

ricochet: Well I looked it up at the time bc same, can never remember. I think this is what I wanted. Did the test flake?

ricochet: https://docs.docker.com/registry/deploying/#customize-the-published-port

autodidaddict (approved):

PR #368, opened 2023-01-12
add echo-messaging to default templates

connorsmith256:
New project template added here: https://github.com/wasmCloud/project-templates/pull/63

Signed-off-by: Connor Smith <connor@cosmonic.com>

### DIFF
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "wasmcloud"]
 description = "wasmcloud Shell (wash) libraries"
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.4.1"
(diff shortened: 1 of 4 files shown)


ricochet (dismissed):

crates/wash-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.4.0"
+version = "0.4.1"
```
brooksmtownsend: I think this should _technically_ be a minor bump but I'm not too worried about it

brooksmtownsend (dismissed):

crates/wash-lib/Cargo.toml
connorsmith256: Done!

connorsmith256:
@brooksmtownsend I think tests are failing because wash has this dep:
```
wash-lib = { version = "0.4.0", path = "./crates/wash-lib", features = ["cli"] }
```

~Which apparently fails now that wash-lib in the path dep only has 0.5.0 available? Not sure if I'm reading the cargo errors correctly. Should we be using a path dep here, or depending on a published version of the crate?~

After talking to @brooksmtownsend offline, he pointed out that when publishing wash to crates.io, despite this dep using `path`, the published crate will actually be used. So my concern is moot :)

brooksmtownsend (approved):

PR #375, opened 2023-01-31
Allow prerelease tags with warning

### LINKED ISSUE #374: [BUG] WASMCLOUD_VERSION doesnt like rc tags
**Describe the bug**
When using `WASMCLOUD_VERSION` to have `wash up` use .60-rc8....it gets unhappy.
```
#9 7.489 wasmCloud version v0.60.0-rc.8 is earlier than the minimum supported version of v0.57.0
```

**To Reproduce**
1. `WASMCLOUD_VERSION=v0.60.0-rc.8 wash up`

brooksmtownsend:
Fixes #374 

This PR allows prerelease tags as defined by semver (the portion after the vX.Y.Z, separated by a dash, then the prerelease type, then the number) of wasmCloud as valid versions to specify. This will let users of wash test out prerelease versions if they so choose, with a warning log.

Just a general note, wash default wasmCloud versions should _not_ be prereleases, this functionality should only be exercised by manually specifying the prerelease version

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -3,6 +3,7 @@ use async_compression::tokio::bufread::GzipDecoder;
 #[cfg(target_family = "unix")]
 use command_group::AsyncCommandGroup;
 use futures::future::join_all;
+use log::warn;
 use std::collections::HashMap;
 use std::io::Cursor;
 #[cfg(target_family = "unix")]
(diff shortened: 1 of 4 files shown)


jordan-rash:
> wash default wasmCloud versions should not be prereleases

what if we live on the _wild side_?

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -314,6 +315,10 @@ fn wasmcloud_url(os: &str, arch: &str, version: &str) -> String {
 fn check_version(version: &str) -> Result<()> {
     let version_req = semver::VersionReq::parse(&format!(">={}", MINIMUM_WASMCLOUD_VERSION))?;
     match semver::Version::parse(version.trim_start_matches('v')) {
+        Ok(parsed_version) if !parsed_version.pre.is_empty() => {
+            warn!("Using prerelease version {} of wasmCloud", version);
+            Ok(())
+        }
```
ghost, line 321: Just checking - we allowed this before and we're just adding an extra warning? Or this enables using a prerelease version?

ghost (approved):

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: This actually enables using prerelease versions, per the cargo version docs all prerelease versions parse as being lesser than published versions. Previously it would fail and we wouldn't allow the prerelease versions

E.g. our rule is `>= 0.57.0`, which `0.60.0-rc.1` does not succeed because it's not published

PR #376, opened 2023-01-31
Create default context if host_config not found

### LINKED ISSUE #371: [BUG] Control interface and context commands fail without a host config
**Describe the bug**
On a fresh `wash` machine, with no `~/.wash/host_config.json` file, any control interface command fails with `No such file or directory`. I thought this was because I was failing to connect to NATS, but it turns out that `wash ctx` commands also fail. I assume this is because we're attempting to form a control client based on those values, but we should be able to gracefully handle a host_config that doesn't exist.

**To Reproduce**
1. `rm ~/.wash/host_config.json`
2. Run any control interface command

**To Workaround**
All you need is a valid `host_config.json`, so running the following commands will allow you to start with default values.
```
wash ctx new host_config; cp ~/.wash/contexts/host_config.json ~/.wash/host_config.json
```

brooksmtownsend:
Fixes #371 

Basically, instead of bubbling up when we can't find `~/.wash/host_config.json`, we will create default values and proceed as normal.

An alternative fix to this would be not creating the default context at all, which I considered but generally involved more steps to manage the state of no host_config

### DIFF
--- a/src/ctx/mod.rs
+++ b/src/ctx/mod.rs
@@ -259,7 +259,8 @@ pub(crate) fn ensure_host_config_context(context_dir: &ContextDir) -> Result<()>
 fn create_host_config_context(context_dir: &ContextDir) -> Result<()> {
     let host_config_ctx = WashContext {
         name: HOST_CONFIG_NAME.to_string(),
-        ..load_context(cfg_dir()?.join(format!("{}.json", HOST_CONFIG_NAME)))?
+        ..load_context(cfg_dir()?.join(format!("{}.json", HOST_CONFIG_NAME)))
+            .unwrap_or_else(|_| WashContext::default())
     };
     context_dir.save_context(&host_config_ctx)?;
     // Set the default context if it isn't set yet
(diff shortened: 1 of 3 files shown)


thomastaylor312 (dismissed):

src/ctx/mod.rs
```diff
@@ -259,7 +259,8 @@ pub(crate) fn ensure_host_config_context(context_dir: &ContextDir) -> Result<()>
 fn create_host_config_context(context_dir: &ContextDir) -> Result<()> {
     let host_config_ctx = WashContext {
         name: HOST_CONFIG_NAME.to_string(),
-        ..load_context(cfg_dir()?.join(format!("{}.json", HOST_CONFIG_NAME)))?
+        ..load_context(cfg_dir()?.join(format!("{}.json", HOST_CONFIG_NAME)))
```
ghost: optional nit: `format!("{HOST_CONFIG_NAME}.json")`

ghost (dismissed):

src/ctx/mod.rs
brooksmtownsend: Huh, didn't know they let you do this! cool

ghost (approved):

PR #379, opened 2023-02-04
[fix] add better error handling for empty tags when --allow-latest is false

ceejimus:
I originally went to address issue #273, when I fell down a small rabbit hole.

**tl;dr** 1) is my hacky way to parse the user's specified tag OK? 2) how should we be smart about looking for images on disk vs. a registry

We have this `--allow-latest` flag which defaults to false, that I'm assuming is for CI purposes. We'd like to be able to inspect claims for actors and providers on disk, in a registry or from a cache. We'd like to be "smart" about it and infer what the user wants with decent fallbacks. We'd like to give the user helpful error messages when they try wonky stuff.

Firstly, if `--allow-latest` is false, we shouldn't default the tag to "latest" when it's not specified and then give the user an error message that what we did behind the scenes isn't allowed. We should tell the user that they need to specify a tag. If the user specifies the "latest" tag explicitly, then we should show the error about not allowing it. This is what my initial commit addresses. Unfortunately, the logic that parses OCI reference strings and the logic that defaults the tag value are both behind the same API, and in order to parse out the input tag in the most consistent way possible, I had to steal some code [from the `oci_distribution` dep](https://github.com/krustlet/oci-distribution). So it's a little hacky, and maybe peeps won't dig it. I suppose we could work with the maintainers of the crate but I'll let future us worry about that.

Secondly, being "smart" about checking local or remote is a tad tricky because, to my knowledge, there is no hard-and-fast disambiguation b/w valid OCI registry URLs and valid file paths (at least in UNIX). A path that looks like a registry reference would be "weird" but valid, and a registry reference could easily look path-ish. I'm being real neurotic here, but I can make a local relative file "localhost:5001/my_actor:tag_s.wasm" and have an image in a registry with the same reference. In this funky case, the user can't inspect the claims in the registry, since the system defaults to looking on disk and finds the .wasm file.
 
There are a few possible solutions that come to mind:
 - be a little smart and if the user happens to make a local wasm/par file that exactly matches a registry URL and they want to check the registry, then out of luck they be
 - be a little smart but give an escape hatch flag (like `--no-local` or `--reg-only`) if the user wants to host an image at a reference like "myreg:5000/myimage:tag_s.wasm"
 - don't be smart and have an opt-in flag (like `--local` or `--on-disk`) for checking local files
 - create a convention that wasmCloud OCI reference URLs MUST NOT end with ".wasm" or ".par" or ".par.gz" and enforce this in wash and wasmCloud, then we can be real smrt

In all cases, a more informative output specifying exactly where we found the bits would be helpful.

Happy nitpicking! :D

P.S. Maybe we could be smarter on Windows since paths are more strict?
P.P.S Also relates to #305

### DIFF
--- a/crates/wash-lib/src/registry.rs
+++ b/crates/wash-lib/src/registry.rs
@@ -13,6 +13,7 @@ use oci_distribution::{
     Reference,
 };
 use provider_archive::ProviderArchive;
+use regex::RegexBuilder;
 use tokio::fs::File;
 use tokio::io::AsyncReadExt;
 
@@ -22,6 +23,9 @@ const PROVIDER_ARCHIVE_CONFIG_MEDIA_TYPE: &str =
 const WASM_MEDIA_TYPE: &str = "application/vnd.module.wasm.content.layer.v1+wasm";
 const WASM_CONFIG_MEDIA_TYPE: &str = "application/vnd.wasmcloud.actor.archive.config";
 const OCI_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar";
+
+// straight up stolen from oci_distribution::Reference
+pub const REFERENCE_REGEXP: &str = r"^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}))?$";
 
 /// Additional options for pulling an OCI artifact
 #[derive(Default)]
@@ -93,9 +97,28 @@ pub async fn get_oci_artifact(
 pub async fn pull_oci_artifact(url: String, options: OciPullOptions) -> Result<Vec<u8>> {
     let image: Reference = url.to_lowercase().parse()?;
 
-    if image.tag().unwrap_or("latest") == "latest" && !options.allow_latest {
-        bail!("Pulling artifacts with tag 'latest' is prohibited.");
-    };
+    // NOTE(ceejimus): the FromStr implementation for the oci_distribution "Reference"
+    // struct defaults the tag to "latest" if unspecified. Ideally, they would expose
+    // some method to check if our valid input string has a tag or not, but, alas ...
+    // earwax.
+    // They don't even make the regex string public, so I stole it. These lines shouldn't fail
+    // if the parsing for "image" doesn't. Unless of course the change it... what a pickle.
+    let re = RegexBuilder::new(REFERENCE_REGEXP)
+        .size_limit(10 * (1 << 21))
+        .build()?;
+    let input_tag = match re.captures(&url) {
+        Some(caps) => caps.get(2).map(|m| m.as_str().to_owned()),
+        None => bail!("Invalid OCI reference URL."),
+    }
+    .unwrap_or(String::from(""));
+
+    if !options.allow_latest {
+        if input_tag == "latest" {
+            bail!("Pulling artifacts with tag 'latest' is prohibited. This can be overriden with the flag '--allow-latest'.");
+        } else if input_tag.is_empty() {
+            bail!("Registry URLs must have explicit tag. To default missing tags to 'latest', use the flag '--allow-latest'.");
+        }
+    }
 
     let mut client = Client::new(ClientConfig {
         protocol: if options.insecure {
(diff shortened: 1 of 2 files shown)


crates/wash-lib/src/registry.rs
```diff
@@ -19,6 +20,9 @@ const WASM_MEDIA_TYPE: &str = "application/vnd.module.wasm.content.layer.v1+wasm
 const WASM_CONFIG_MEDIA_TYPE: &str = "application/vnd.wasmcloud.actor.archive.config";
 const OCI_MEDIA_TYPE: &str = "application/vnd.oci.image.layer.v1.tar";
 
+// straight up stolen from oci_distribution::Reference
+pub const REFERENCE_REGEXP: &str = r"^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)(?::([\w][\w.-]{0,127}))?(?:@([A-Za-z][A-Za-z0-9]*(?:[-_+.][A-Za-z][A-Za-z0-9]*)*[:][[:xdigit:]]{32,}))?$";
```
brooksmtownsend, line 28: I remember that this regex string can possibly result in a very large compiled use of memory, @thomastaylor312 can you check me on that assumption?

crates/wash-lib/src/registry.rs
```diff
@@ -89,9 +93,28 @@ pub async fn get_oci_artifact(
...
     let image: Reference = url.to_lowercase().parse()?;
 
-    if image.tag().unwrap_or("latest") == "latest" && !options.allow_latest {
-        bail!("Pulling artifacts with tag 'latest' is prohibited.");
-    };
+    // NOTE(ceejimus): the FromStr implementation for the oci_distribution "Reference"
+    // struct defaults the tag to "latest" if unspecified. Ideally, they would expose
+    // some method to check if our valid input string has a tag or not, but, alas ...
+    // earwax.
+    // They don't even make the regex string public, so I stole it. These lines shouldn't fail
```
brooksmtownsend, line 104: FWIW, Taylor is one of the maintainers of `oci-distribution` and I've contributed over there before, so we can certainly look to contribute something!

brooksmtownsend (commented):
Would love to consider adding an error enum to `oci-distribution` for the `parse()` method to return specific error kinds or something to indicate why a parse failed. Otherwise, it would also be great to expose the regex builder from the `oci-distribution` crate to not copy paste. I like the solution, though!

crates/wash-lib/src/registry.rs
ceejimus: I did have to use the size_limit() modifier on the RegexBuilder to increase the max compile size of the thing to 20MB (I also just stole this from oci_distribution). The oci_distribution crate uses lazy_static to build this at runtime. We could do the same thing, but we'd need to add a dependency.

ceejimus:
In order to get the tests to pass I changed the regex dependency to required. It was optional under the "cli" feature. I'm wondering if this is the right way to go.

brooksmtownsend (commented):
@ceejimus thanks for your patience on this one! The biggest outstanding questions I have (which is for @thomastaylor312) is if we can consider exposing some function from `oci-distribution` to do the parsing and to reduce code duplication. If it's not a goal of the crate then I think this approach works well and we can get this approved 😄

crates/wash-lib/src/registry.rs
thomastaylor312: Yeah, this is _really_ big. In fact I was planning on coming up with a way to use a different regex library to take up less space with this. No need to bring in lazy static at this point I think

crates/wash-lib/src/registry.rs
thomastaylor312: Yeah, this can easily be requested

thomastaylor312:
@brooksmtownsend I'd suggest the following here: We leave this as is and merge, but before we do so, we open up an issue against oci-distribution, and then our own issue that says we should use this if merged into oci-distribution

brooksmtownsend (dismissed):
From @thomastaylor312 's suggestion, I like merging this in alongside an issue in oci-distribution to make it a little nicer 😄 

FWIW the current (0.15.1) `wash` clocks in at 24Mb on my mac and this comes in at 23mb so I don't think regex makes it too large :D 

@ceejimus just need a rebase from ya and we can get this going. Would you prefer to write the oci-distribution issue, or would you like one of us to handle it?

ceejimus:
Hey, sorry for the delays.

I could make an issue on the other repo. I'd like to clarify what we intend to make the issue about, though. The real problem I ran into in this fix is that there's no way to pull the user-supplied tag from the parsing process. If the user supplies no tag, the `from_str` (on the `TryFrom` impl) method defaults to "latest" and returns that (see [this line](https://github.com/krustlet/oci-distribution/blob/55721aa97d36f52a59625366a24ff75715bdee47/src/reference.rs#L198)).  Even though the `tag` member of the `Reference` struct is an `Option`, there's no way the parse can return a None if you go through `from_str`. If I had it my way, the result from parse would have an `Option::None` result for `tag` when the user-supplied tag is empty. I think the logic for replacing empty tags with the default belongs in the `whole()` method which converts a reference to a string. This is a breaking change though so...

In this context the minimum request would be to add some additional method to extract the just tag from a request. This solution is sub-optimal in my view, since the structure for handling this more naturally already exists in the types (namely the `Reference.tag` being an `Option`).

Now, that I've written it, this could be the blurb for the new issue? :D

brooksmtownsend (approved):
@thomastaylor312 any other thoughts for this one? If it's good for you, merge away

thomastaylor312:
@ceejimus I think something like what you just said would make a good issue for oci-distribution

PR #381, opened 2023-02-08
Bumped to wasmCloud 0.60.0

brooksmtownsend:
This PR bumps to 0.15.0 including the wascap bump from #361 and wasmCloud 0.60.0 which are meant to interact with each other in terms of claims signing on actors.

In a separate commit I also bumped lock deps with `cargo update`, happy to back that out if desired.

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -1,15 +1,16 @@
-use anyhow::{anyhow, Result};
-use async_compression::tokio::bufread::GzipDecoder;
-#[cfg(target_family = "unix")]
-use command_group::AsyncCommandGroup;
-use futures::future::join_all;
-use log::warn;
 use std::collections::HashMap;
 use std::io::Cursor;
 #[cfg(target_family = "unix")]
 use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+
+use anyhow::{anyhow, Result};
+use async_compression::tokio::bufread::GzipDecoder;
+#[cfg(target_family = "unix")]
+use command_group::AsyncCommandGroup;
+use futures::future::join_all;
+use log::warn;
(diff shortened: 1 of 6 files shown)


connorsmith256 (dismissed):

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -84,9 +89,9 @@ where
     P: AsRef<Path>,
 {
     check_version(version)?;
-    if is_wasmcloud_installed(&dir).await {
+    if is_wasmcloud_installed(&dir, version).await {
         // wasmCloud already exists, return early
-        return Ok(dir.as_ref().join(WASMCLOUD_HOST_BIN));
+        return Ok(dir.as_ref().join(version).join(WASMCLOUD_HOST_BIN));
```
connorsmith256: Nit: we're breaking encapsulation of `is_wasmcloud_installed` by assuming the directory it's checking is `[dir]/[version]/[bin]`. It would be safer to rename the function to something like `find_wasmcloud_binary` and have it return an `Option<Path>` containing the install path

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -164,7 +170,7 @@ where
             )
         })?;
         if let Ok(path) = entry.path() {
-            let file_path = dir.as_ref().join(path);
+            let file_path = dir.as_ref().join(version).join(path);
```
connorsmith256: The doc comment on this function should be updated to specify `[dir]/[version]` will be used as the install dir, rather than just `dir`

connorsmith256 (changes requested):

crates/wash-lib/src/start/wasmcloud.rs
thomastaylor312: Ok, this is now addressed

brooksmtownsend (commented):
✅

connorsmith256 (approved):
🎉

PR #385, opened 2023-02-10
Consolidating claims inspect and par inspect command into a new inspect command

### LINKED ISSUE #369: [FEATURE] Consolidate `claims inspect` and `par inspect` commands
When inspecting actors or capability providers, developers use the `wash claims inspect <actor>` or `wash par inspect <provider>`, respectively. The argument to this function can either be a path to a local file or a valid OCI reference.

The problem comes from mixing up these commands due to confusion. When inspecting a capability provider, you may want to view its embedded claims (valid sentence) and you use `wash claims inspect`. However, since we're expecting a Wasm file, `wash` throws a parse error and you're left confused.

I'd like to consolidate these commands into a single `wash inspect` command. The semantics would be `wash inspect <file_or_reference>` where the file/reference could be an actor OR a capability provider, and we return the relevant information either way. Ideally, in the future, we could support more information than just claims here too, like Wasm imports, WASI support, etc. But for this ticket, simply merging these commands would be the definition of done.

aish-where-ya:
Created a new command named `inspect` that will perform the functionalities of `par inspect` and `claims inspect`.
Added deprecation warnings for `par inspect` and `claims inspect`.
Added unit and integration tests.

Fixes #369

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -516,12 +516,12 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.1.6"
+version = "4.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec0b0588d44d4d63a87dbd75c136c166bbfd9a86a31cb89e09906521c7d3f5e3"
+checksum = "c3d7ae14b20b94cb02149ed21a86c423859cbe18dc7ed69845cace50e52b40a5"
 dependencies = [
  "bitflags",
- "clap_derive 4.1.0",
+ "clap_derive 4.1.8",
  "clap_lex 0.3.0",
  "is-terminal",
  "once_cell",
@@ -535,7 +535,7 @@ version = "4.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bd125be87bf4c255ebc50de0b7f4d2a6201e8ac3dc86e39c0ad081dc5e7236fe"
 dependencies = [
- "clap 4.1.6",
+ "clap 4.1.8",
 ]
 
 [[package]]
@@ -553,9 +553,9 @@ dependencies = [
 
 [[package]]
 name = "clap_derive"
-version = "4.1.0"
+version = "4.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "684a277d672e91966334af371f1a7b5833f9aa00b07c84e92fbce95e00208ce8"
+checksum = "44bec8e5c9d09e439c4335b1af0abaab56dcf3b94999a936e1bb47b9134288f0"
 dependencies = [
  "heck 0.4.0",
  "proc-macro-error",
@@ -3944,7 +3944,7 @@ dependencies = [
  "bytes",
  "cargo_atelier",
  "chrono",
- "clap 4.1.6",
+ "clap 4.1.8",
  "clap_complete",
  "cloudevents-sdk",
  "console",
@@ -3997,7 +3997,7 @@ dependencies = [
  "async-compression",
  "cargo_toml",
  "claims",
- "clap 4.1.6",
+ "clap 4.1.8",
  "command-group",
  "config",
  "console",
@@ -4340,7 +4340,7 @@ dependencies = [
  "atelier_smithy",
  "bytes",
  "cfg-if",
- "clap 4.1.6",
+ "clap 4.1.8",
  "directories",
  "downloader",
  "handlebars",


crates/wash-lib/src/cli/claims.rs
```diff
@@ -672,11 +673,12 @@ async fn render_caps(cmd: InspectCommand) -> Result<CommandOutput> {
         }
         None => bail!("No capabilities discovered in : {}", module_name),
     };
+    warn!("claims inspect will be deprecated in future versions. Use inspect instead.");
```
ghost: I'd just remove the old command entirely but will defer if anyone has a stronger opinion. I don't think we're at a stage where we need to support deprecated commands

(@brooksmtownsend)

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+    password: Option<String>,
+
+    /// Allow insecure (HTTP) registry connections
+    #[clap(long = "insecure")]
+    insecure: bool,
+
+    /// skip the local OCI cache
+    #[clap(long = "no-cache")]
+    no_cache: bool,
+
```
ghost: nit - extra newline

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+fn is_wasm(input: &[u8]) -> Result<bool> {
+    if input.len() < 4 {
+        bail!("Not enough bytes to be a valid claim file")
```
ghost: Could this just return a bool (with no Result) and the `if input.len() < 4` block return false?

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+            Alignment::Left,
+        )]));
+
+        table.render()
+    };
+
+    Ok(CommandOutput::new(text_table, map))
+}
+
+pub async fn handle_command(
```
ghost: nit: put this at the top as it's the "entry point" of this command.

src/par.rs
```diff
@@ -355,6 +356,7 @@ pub(crate) async fn handle_inspect(cmd: InspectCommand) -> Result<CommandOutput>
 
         table.render()
     };
+    warn!("par inspect will be deprecated in future versions. Use inspect instead.");
```
ghost: Same here, I'd just remove outright unless someone has a stronger opinion

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
+use clap::{Parser};
```
ghost: Formatting of imports is a little weird, there's extra commas and spaces. I think rustfmt will fix this stuff?

Also, could you add some top-level comments to the functions? I get what's going on but it might be hard to parse for newer wash developers.

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+    )
+    .await?;
+
+    // Extract will return an error if it encounters an invalid hash in the claims
+    Ok(wascap::wasm::extract_claims(artifact_bytes)?)
+}
+
+async fn handle_actor_module(cmd: InspectCliCommand) -> Result<CommandOutput> {
+    let module_name = cmd.module.clone();
+    let jwt_only = cmd.jwt_only;
```
ghost: No need to destructure this if you're only using it once

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+    let out = match caps {
+        Some(token) => {
+            if jwt_only {
+                CommandOutput::from_key_and_text("token", token.jwt)
+            } else {
+                let validation = wascap::jwt::validate_token::<Actor>(&token.jwt)?;
+                render_actor_claims(token.claims, validation)
+            }
+        }
+        None => bail!("No capabilities discovered in : {}", module_name),
```
ghost: `caps.ok_or_else(|| anyhow!(...))?`

Would make this easier to read IMO

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+    command: InspectCliCommand,
+    _output_kind: OutputKind,
+) -> Result<CommandOutput> {
+    let mut buf = Vec::new();
+    let mut f = File::open(command.module.clone())?;
+    f.read_to_end(&mut buf)?;
+
+    let jwt_only = command.jwt_only;
+
+    if is_wasm(&buf).is_ok() || jwt_only {
```
ghost: is the `|| jwt_only` necessary?

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+
+fn is_wasm(input: &[u8]) -> Result<bool> {
+    if input.len() < 4 {
+        bail!("Not enough bytes to be a valid claim file")
+    }
+    Ok(input[0..4] == WASM_MAGIC)
+}
+
+async fn get_caps(cmd: InspectCliCommand) -> Result<Option<Token<Actor>>> {
+    let cache_path = (!cmd.no_cache).then(|| cached_oci_file(&cmd.module));
```
ghost: TIL about using .then() on bools to create `Option`s! I thought this was a future for a minute and got really confused lol

src/inspect.rs
```diff
@@ -0,0 +1,391 @@
...
+        .await
+        .map_err(|e| anyhow!("{}", e))?;
+    let claims = artifact
+        .claims()
+        .ok_or_else(|| anyhow!("No claims found in artifact"))?;
+    let metadata = claims
+        .metadata
+        .ok_or_else(|| anyhow!("No metadata found"))?;
+
+    let friendly_rev = match metadata.rev {
```
ghost: `.map_or("None".to_string(), |rev| rev.to_string())` would be easier to read here.

ghost (changes requested):
Hi @aish-where-ya! Thanks for the contribution!

Left some comments below. Let me know if you need more info on any of these.

crates/wash-lib/src/cli/claims.rs
brooksmtownsend: I feel like we've got a few scripts or actions laying around that inspect an actor to pull out the public key, I just wouldn't want to break those immediately and I'd rather give a chance to discover scripts that had that command

ghost: That makes sense. Current behavior looks good then!

tests/integration_inspect.rs
```diff
@@ -0,0 +1,223 @@
...
+#[test]
+fn integration_inspect_provider() {
+    const SUBFOLDER: &str = "inspect_test";
+    const HTTP_OCI: &str = "wasmcloud.azurecr.io/httpclient:0.3.5";
+    const HTTP_ISSUER: &str = "ACOJJN6WUP4ODD75XEBKKTCCUJJCY5ZKQ56XVKYK4BEJWGVAOOQHZMCW";
+    const HTTP_SERVICE: &str = "VCCVLH4XWGI3SGARFNYKYT2A32SUYA2KVAIV2U2Q34DQA7WWJPFRKIKM";
+    let inspect_dir = test_dir_with_subfolder(SUBFOLDER);
+    let httpclient_inspect = &format!("{}/httpclient:inspect", LOCAL_REGISTRY);
+
+    // Pull the echo module and push to local registry to test local inspect
```
brooksmtownsend: nit: we're pulling the httpclient provider here

brooksmtownsend: Additionally, below, some of the variables are still named from the copy-paste from the other tests. For clarity, I think we should rename those to match this provider and test

crates/wash-lib/src/cli/claims.rs
```diff
@@ -672,14 +673,12 @@ async fn render_caps(cmd: InspectCommand) -> Result<CommandOutput> {
...
```
brooksmtownsend: I think it would be good at this point to remove the code from `claims inspect` that actually performs the inspect and instead call the function in `src/inspect.rs` that you wrote. That way we only have one piece of code that's performing the inspect, and the `wash claims inspect` here is simply a thin (deprecated) wrapper around `inspect`.

src/par.rs
```diff
@@ -265,7 +266,7 @@ pub(crate) async fn handle_create(
 }
 
 /// Loads a provider archive and outputs the contents of the claims
-pub(crate) async fn handle_inspect(cmd: InspectCommand) -> Result<CommandOutput> {
+pub(crate) async fn handle_par_inspect(cmd: InspectCommand) -> Result<CommandOutput> {
```
brooksmtownsend: Same comment here as claims, I think the `par inspect` function should just be a thin wrapper around calling the `src/inspect.rs` function that you wrote

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+use serde_json::json;
+use std::{collections::HashMap, fs::File, io::Read, path::PathBuf};
+use term_table::{row::Row, table_cell::*, Table};
+use wascap::jwt::{Actor, Token};
+use wash_lib::{
+    cli::{cached_oci_file, claims::render_actor_claims, CommandOutput, OutputKind},
+    registry::OciPullOptions,
+};
+
+const WASM_MAGIC: [u8; 4] = [0x00, 0x61, 0x73, 0x6D];
```
brooksmtownsend: A comment indicating where you got this magic number would be helpful for future code readers

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+
+#[derive(Debug, Parser, Clone)]
+pub struct InspectCliCommand {
+    /// Path to signed actor module or provider archive or OCI URL of signed actor module or provider archive
+    pub module: String,
```
brooksmtownsend: nit: this might read better as `Path or OCI URL to signed actor...`

brooksmtownsend: also, it might be better to call this `artifact` or `inspect_target` or something that's actor/provider oci/path agnostic. Happy to entertain suggestions here

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+    command: InspectCliCommand,
+    _output_kind: OutputKind,
+) -> Result<CommandOutput> {
+    let mut buf = Vec::new();
+    if PathBuf::from(command.module.clone()).as_path().is_dir() {
+        let mut f = File::open(command.module.clone())?;
+        f.read_to_end(&mut buf)?;
+    } else {
+        let cache_file =
+            (!command.no_cache.clone()).then(|| cached_oci_file(&command.module.clone()));
```
brooksmtownsend: nit: these `.clone()`s are probably not necessary and might drop this down to one line if removed

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+        buf = wash_lib::registry::get_oci_artifact(
+            command.module.clone(),
+            cache_file,
+            OciPullOptions {
+                digest: command.digest.clone(),
+                allow_latest: command.allow_latest.clone(),
+                user: command.user.clone(),
+                password: command.password.clone(),
+                insecure: command.insecure.clone(),
```
brooksmtownsend: These booleans probably don't need to be cloned

src/inspect.rs
brooksmtownsend: @aish-where-ya unresolving this conversation since it seems to be present still

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+        cmd.module,
+        cache_file,
+        OciPullOptions {
+            digest: cmd.digest,
+            allow_latest: cmd.allow_latest,
+            user: cmd.user,
+            password: cmd.password,
+            insecure: cmd.insecure,
+        },
+    )
```
brooksmtownsend: I think it would be better to pass the bytes down from the initial inspect function to save downloading the OCI artifact twice here

src/inspect.rs
```diff
@@ -0,0 +1,405 @@
...
+        cache_path,
+        OciPullOptions {
+            digest: cmd.digest,
+            allow_latest: cmd.allow_latest,
+            user: cmd.user,
+            password: cmd.password,
+            insecure: cmd.insecure,
+        },
+    )
+    .await?;
```
brooksmtownsend: Let's make the artifact bytes an input for this function since you already pull the bytes down as a part of inspecting the first few to see if it's a Wasm file

brooksmtownsend (changes requested):
The `inspect` module is looking great! I have a few requests for refactoring opportunities since the code is largely contained in `src/inspect` now, and to remove some duplicate `get_oci_bytes` calls that are inefficient since you already download the bytes in the first step.

src/inspect.rs
aish-where-ya: I had destructured this to make the difference between handling actors and providers a bit clearer. But that is okay, I will collate it into a single block.

aish-where-ya:
This is one of those "the checks run successfully on my local machine" errors.

PR #386, opened 2023-02-14
Added devcontainer for wash

brooksmtownsend:
Simple devcontainer that installs `rust` and bind mounts this project into a container. Should work with codespaces and make it easy to develop `wash` even on machines with no Rust tooling!

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -1,0 +1,7 @@
+# wash devcontainer
+
+A simple devcontainer that has `rust` installed. Try it out with `devcontainer open` at the root of this repository!
+
+## Prerequisites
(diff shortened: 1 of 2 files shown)


ghost:
From the way you've written the PR it sounds like you haven't tested the container? "should work with codespaces".

I don't think we should add extra config files to the repo if they haven't been tested / aren't being used. Then it's on us to maintain them later. However if this is being used by you or someone else then I'll approve

brooksmtownsend:
@mattwilkinsonn I tested this locally with devcontainer open, which worked swimmingly! I didn't quite figure out a good workflow for starting a codespace using this devcontainer, but it turns out that just navigating to the branch and creating a codespace will use this by default.

image

Works great 😄

ghost (approved):
LGTM!

PR #389, opened 2023-02-15
ci: setup dependabot

rvolosatovs:
Setup Dependabot to keep Rust crates and GitHub actions themselves up-to-date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: "cargo"
+  directory: "/"
+  schedule:
+    interval: "daily"
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: "cargo"
+  directory: "/"
```
ghost, line 4: Will this also update the dependencies of the internal `wash-lib` crate? (`crates/wash-lib`). Not sure if we need additional configuration for that

ghost (commented):
Hi @rvolosatovs! Thanks for the contribution 🎉!

Had one quick question below.

.github/dependabot.yml
rvolosatovs: I was under the impression that it would, but it appears it does not https://github.com/rvolosatovs/wash/pulls (at least for this particular repository). The docs are pretty vague and I do remember this working in other repositories for subcrates

In any case, it appears that the right approach would be to utilize workspace dependencies in the repository anyway, let me file a PR for that first.

brooksmtownsend: Agreed with Matt that it would be best to get both projects bumping, and that should set a good precedent if we bring in more crates to this repository 😄

rvolosatovs: > Agreed with Matt that it would be best to get both projects bumping, and that should set a good precedent if we bring in more crates to this repository 😄 

Let me introduce workspace deps and once that's done, this will not be an issue, since all dependency versions for all sub-crates will be specified in a single top-level `Cargo.toml`.
I will only address this next week though

rvolosatovs: See #407 , once that one is merged - this will be an non-issue

rvolosatovs: #407 is merged. @mattwilkinsonn @brooksmtownsend PTAL

ghost (approved):
LGTM!

brooksmtownsend (approved):
LGTM 😄

Issue #341, opened 2022-11-18
[BUG] Actor Naming does not respect " - "


andar1an:
**Describe the bug**
Create an actor with a dash in the name, and try to build. 
```Could not find compiled wasm file to sign: target/wasm32-unknown-unknown/release/actor-name.wasm```

**To Reproduce**
1. Run command `wash new actor actor-name -t hello`
1. Run other command `wash build`
1. See error

**Expected behavior**
Wash respects dash in an actor name, and build without needing to change the name in wasmcloud.toml to use an underscore (@jordan-rash discovered this workaround).

**Screenshots**

**Environment (please complete the following information) **
 - OS: Manjaro 22.0.0
 - Shell: zsh
 - wash Version: 13

**Comment by stevelr** [comment]
Some additional background information and how I think we could fix it:

There are a couple limitations that restrict how we can handle hyphens:
- cargo won't compile a library name with a hyphen in it, so the project name needs to be changed to have an underscore. The generated wash file name will have an underscore in the name. 
- it is possible to have a actor name that contains a dash, even though the wasm file contains underscores, by using the `--name` parameter to `wash sign`. That would let the actor show up in host inventory and the washboard with a hyphen, but that's a bit limited: Call aliases also can't have underscores (currently), so if you were planning to call the actor with a friendly name you'd have to change the hyphens to underscores. 

Options:

(1) replace the project name with underscore, so at least wash new && wash build don't fail.  This can be done with a change to the project templates: Use  `name = {{to_snake_case project-name}}` (whoever implements this, it should be done for all project templates). We may want to print a notice on wash's output to tell the user that the name was changed. 

(2) change the project name and also rename the actor at signing time. In addition to the change (1) on the template, this would require a change to the wasmcloud.toml parser and adding the `--name` parameter to `wash sign`.

(3) reject project names with hyphens and make the user re-submit the wash new command with a non-hyphenated name. 

IMO (1) is preferable to (3) since (3) is an abrupt failure and (1) succeeds with a minor change, plus a notice to train the user how to do it right the next time.

**Comment by andar1an** [comment]
I also like (1), especially with notice and auto-renaming to underscore.

**Comment by brooksmtownsend** [comment]
Agreed with above, `1` sounds like a built-in way to handle the problem!

**Comment by ricochet** [comment]
Another case to be handled that seems related:

```bash
wash new actor blobs3
cd blobs3
wash build
```

Compiles but then this error:
[...]

PR #390, opened 2023-02-15
fix: use lib.name from cargo.toml for rust wasm binary name

vados-cosmonic:
This PR adds code to use `lib.name` from `Cargo.toml` for Rust projects, to avoid issues with Rust project builds when actor name does not match Rust's requirements on output artifact names. 

See #341 for for more discussion around this issue, and the [RFC on naming](https://github.com/wasmCloud/wasmCloud/issues/271) that should standardize naming across libraries in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,17 +28,29 @@ build-watch: ## Continuously build the project
 
 ##@ Testing
 
-test: ## Run the entire unit test suite
-	@$(CARGO) nextest run --no-fail-fast -p wash-lib
-	@$(CARGO) nextest run --no-fail-fast --bin wash
+# Target to focus on for tests (enables running only one test)
+CARGO_TEST_TARGET ?= ""
+
+test: ## Run unit test suite
+	@$(CARGO) nextest $(CARGO_TEST_TARGET) --no-fail-fast --bin wash
+	@$(CARGO) nextest $(CARGO_TEST_TARGET) --no-fail-fast -p wash-lib
+
+test-watch: ## Run unit tests continously, can optionally specify a target test filter.
+	@$(CARGO) watch -- $(CARGO) nextest run $(TARGET)
 
 test-integration: ## Run the entire integration test suite
 	@$(DOCKER) compose -f ./tools/docker-compose.yml up --detach
 	@$(CARGO) nextest run --profile integration -E 'kind(test)'
 	@$(DOCKER) compose -f ./tools/docker-compose.yml down
 
-test-watch: ## Run unit tests continously, can optionally specify a target test filter.
-	@$(CARGO) watch -- $(CARGO) nextest run $(TARGET)
+test-integration-watch: ## Run integration test suite continuously
+	@$(CARGO) watch -- $(MAKE) test-integration
+
+test-unit: ## Run one or more unit tests
+	@$(CARGO) nextest $(CARGO_TEST_TARGET)
+
+test-unit-watch: ## Run tests continuously
+	@$(CARGO) watch -- $(MAKE) test-unit
 
 rust-check: ## Run rust checks
 	@$(CARGO) fmt --all --check
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -134,6 +134,12 @@ fn build_rust_actor(
         bail!("Compiling actor failed: {}", result.to_string())
     }
 
+    // Determine the wasm binary name
+    let wasm_bin_name = common_config
+        .wasm_bin_name
+        .as_ref()
+        .unwrap_or(&common_config.name);
+
     let wasm_file = PathBuf::from(format!(
         "{}/{}/release/{}.wasm",
         rust_config
@@ -142,7 +148,7 @@ fn build_rust_actor(
             .unwrap_or_else(|| PathBuf::from("target"))
             .to_string_lossy(),
         actor_config.wasm_target,
-        common_config.name,
+        wasm_bin_name,
     ));
 
     if !wasm_file.exists() {
@@ -153,7 +159,7 @@ fn build_rust_actor(
     }
 
     // move the file out into the build/ folder for parity with tinygo and convienience for users.
-    let copied_wasm_file = PathBuf::from(format!("build/{}.wasm", common_config.name));
+    let copied_wasm_file = PathBuf::from(format!("build/{}.wasm", wasm_bin_name));
     if let Some(p) = copied_wasm_file.parent() {
         fs::create_dir_all(p)?;
     }
--- a/tests/common.rs
+++ b/tests/common.rs
@@ -31,11 +31,21 @@ pub(crate) fn get_json_output(output: std::process::Output) -> Result<serde_json
 
 #[allow(unused)]
 /// Creates a subfolder in the test directory for use with a specific test
+pub(crate) fn test_dir() -> PathBuf {
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR").expect("$CARGO_MANIFEST_DIR");
+    let test_dir = PathBuf::from(manifest_dir).join("tests/fixtures");
+    remove_dir_all(&test_dir);
+    create_dir_all(&test_dir);
+    test_dir
+}
+
+#[allow(unused)]
+/// Creates a subfolder in the test directory for use with a specific test
 /// It's preferred that the same test that calls this function also
 /// uses std::fs::remove_dir_all to remove the subdirectory
 pub(crate) fn test_dir_with_subfolder(subfolder: &str) -> PathBuf {
-    let root_dir = &env::var("CARGO_MANIFEST_DIR").expect("$CARGO_MANIFEST_DIR");
-    let with_subfolder = PathBuf::from(format!("{root_dir}/tests/fixtures/{subfolder}"));
+    let test_dir = test_dir();
+    let with_subfolder = test_dir.join(subfolder);
     remove_dir_all(with_subfolder.clone());
     create_dir_all(with_subfolder.clone());
     with_subfolder
@@ -46,6 +56,5 @@ pub(crate) fn test_dir_with_subfolder(subfolder: &str) -> PathBuf {
 /// to the test fixtures directory. This does _not_ create the file,
 /// so the test is responsible for initialization and modification of this file
 pub(crate) fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
-    let root_dir = &env::var("CARGO_MANIFEST_DIR").expect("$CARGO_MANIFEST_DIR");
-    PathBuf::from(format!("{root_dir}/tests/fixtures/{subfolder}/{file}"))
+    test_dir_with_subfolder(subfolder).join(file)
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -437,12 +437,12 @@ dependencies = [
 
 [[package]]
 name = "cargo_toml"
-version = "0.13.3"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "497049e9477329f8f6a559972ee42e117487d01d1e8c2cc9f836ea6fa23a9e1a"
+checksum = "7f83bc2e401ed041b7057345ebc488c005efa0341d5541ce7004d30458d0090b"
 dependencies = [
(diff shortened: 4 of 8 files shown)


vados-cosmonic:
Fixed lint, rebased to newest master to resolve the conflicts!

vados-cosmonic:
## :warning: some tests are failing locally

Weird -- the following tests are failing for me on linux locally but pass in CI:

`folder_path`
`minimal_rust_actor`
`rust_actor`

Please don't merge yet! Tracking down what's happening, it looks like some of the tests depend on the  [the non-language specific short circuit behavior](https://github.com/wasmCloud/wash/blob/main/crates/wash-lib/src/parser/mod.rs#L338)...

[EDIT] - converting back to draft for now

vados-cosmonic:
Refactored this to discover the binary name right before use when building (in `wash-lib/src/build.rs`). I originally wrote code to add it to `CommonConfig` so we could know the name for more of the process, but the more minimal change is to have it in build.rs.

@mattwilkinsonn would you mind taking a look at this?

tests/common.rs
```diff
@@ -29,13 +31,33 @@ pub(crate) fn get_json_output(output: std::process::Output) -> Result<serde_json
     Ok(json)
 }
 
+#[allow(unused)]
+/// Creates a temporary directory unique to the given test name
+pub(crate) fn tmp_dir_for_test(test_name: &str) -> PathBuf {
```
ghost: What's the reasoning behind the test directory changes?

tests/integration_new.rs
```diff
@@ -0,0 +1,68 @@
...
+        .args(["new", "actor", "dashed-actor", "-t", "hello"])
+        .current_dir(&root_dir)
+        .stdout(stdout.try_clone()?)
+        .spawn()?;
+    assert!(new_cmd.wait()?.success());
+
+    // Ensure that the actor dir was created as expected
+    assert!(actor_dir.exists());
+
+    stdout.write_all(
```
ghost: I'd prefer against adding extra print statements to tests, could get confusing when trying to debug them

tests/integration_new.rs
```diff
@@ -0,0 +1,68 @@
...
+// This test ensures that names that are invalid for languages that don't support dashed names
+// are detected and corrected by wash
+//
+// The failure that we are avoiding/correcting for, when printed normally:
+//
+// Could not find compiled wasm file, please ensure target/wasm32-unknown-unknown/release/dashed-actor.wasm exists
+// thread 'integration_new_handles_dashed_names' panicked at 'assertion failed: build_cmd.wait()?.success()', tests/integration_new.rs:65:5
+#[test]
+#[serial]
+fn integration_new_handles_dashed_names() -> Result<(), anyhow::Error> {
```
ghost: This test is really testing that `wash build` can successfully handle names that are different in `Cargo.toml` and `wasmcloud.toml`, correctly handling dashed names is a side effect of that being fixed.

With that in mind, I think we should test that behavior directly in `integration_build.rs`.

ghost: Quick tip: import `anyhow::Result`, and then you can write this Result type like `Result<()>`!

tests/integration_new.rs
```diff
@@ -0,0 +1,68 @@
+mod common;
+use common::{tmp_dir_for_test, wash};
+use serial_test::serial;
+use std::fs::{remove_dir_all, File};
+use std::io::Write;
+
+use dispose::defer;
```
ghost: `scopeguard` already does this, see `integration_build.rs`. If dispose is better for some reason we can use that, but i'd prefer not to have two ways of doing the same thing

crates/wash-lib/src/build.rs
```diff
@@ -134,6 +135,19 @@ fn build_rust_actor(
         bail!("Compiling actor failed: {}", result.to_string())
     }
 
+    // Determine the wasm binary name
+    let mut cargo_toml = Manifest::from_path(common_config.path.join("Cargo.toml"))?;
+    cargo_toml.complete_from_path(&common_config.path)?;
```
ghost: Why do you need `::from_path` and then `.complete_from_path` afterwards? Seems like a weird API design

crates/wash-lib/src/build.rs
```diff
@@ -134,6 +135,19 @@ fn build_rust_actor(
...
+
+    // Determine the wasm module name from the [lib] section of Cargo.toml
+    let wasm_bin_name = match cargo_toml.lib {
```
ghost: Does `cargo_toml.name` not work here? That is what [the parser uses](https://github.com/wasmCloud/wash/blob/main/crates/wash-lib/src/parser/mod.rs#L361)

Makefile
```diff
@@ -1,34 +1,54 @@
...
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_\-.*]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+clean: ## Clean all tests
+	cargo clean
+	wash drain all
+
+##@ Building
+
+build: ## Build the project
+	@$(CARGO) build
```
ghost: What's the reasoning behind these Makefile changes?

ghost (commented):
Left some comments! On the `CommonConfig` stuff - i'm not against either option but perhaps it could be useful to add an extra field like `language_package_name` and then just use that in `build.rs`? Adding it inside the config parsing would let other consumers of `wasmcloud.toml` get that information more easily.

tests/integration_new.rs
vados-cosmonic: I can remove this -- I wanted to break up the output text that was getting to the log file if anyone was ever going to read it, but I'm not tied to it at all

- [x] remove extra logging to output during test

Makefile
vados-cosmonic: A few things:
- no easy way to do a continuous build and watch of the project (which I personally find helpful for development, and thought others might as well)
- no easy way to do a continuous build and watch on a specific test (for example if you develop in a TDD style)
- consistent use of swappable binary commands (being able to easily influence `docker`/`cargo`/etc is a nice-to-have
- `help` as the first (default) target is a nice default

Definitely nice-to-haves though, I'm not tied to the changes here, so if they're obscuring the rest  I can certainly remove!

crates/wash-lib/src/build.rs
vados-cosmonic: So it turns out that some portions of the structure are actually not filled out by default -- maybe I'm missing something but [that seems like how to do it](https://docs.rs/cargo_toml/0.15.2/cargo_toml/struct.Manifest.html#method.complete_from_path)

crates/wash-lib/src/build.rs
vados-cosmonic: I found that was the `package` name, which is looser than the WASM binary name (under `[lib]`) is allowed to be. As of right now, the template correctly sets the `lib` name but the package name stays unusable/not the right one.

tests/common.rs
vados-cosmonic: While I was running the tests, if you try to `wash build` anywhere under the repo (the default test folder starting point), `cargo` gets confused about workspaces with the default directory that was being used, and doesn't let you build.

The project needed to be templated out somewhere completely different, and normally what you'd reach for is a system temp dir so I sprung for that.

I don't think there was a test of this type (that tries to actually template out, build and check), so I think that's why no one has run into it before

tests/integration_new.rs
vados-cosmonic: Can do -- wasn't aware of `scopeguard`, I missed that test, will switch

- [x] switch to `scopeguard`, remove `dispose`

tests/integration_new.rs
vados-cosmonic: Ah that's a good point, I'll move this test over to `integration_build.rs`

- [x] move test to `integration_build.rs`

vados-cosmonic: Will do, thanks for the tip!

- [x] Use `anyhow::Result` for more succinct return type

vados-cosmonic:
Apologies for the delay @mattwilkinsonn -- just responded to the comments and will make the changes by tomorrow

> Left some comments! On the CommonConfig stuff - i'm not against either option but perhaps it could be useful to add an extra field like language_package_name and then just use that in build.rs? Adding it inside the config parsing would let other consumers of wasmcloud.toml get that information more easily.

Funny, this is [exactly how I did it](https://github.com/wasmCloud/wash/pull/390/commits/37733d770fc4dd34a2aaa2c7de60a0a4e3734c44#) (called it `wasm_bin_name`) then ripped it out.

I can definitely switch back to this -- I personally like the `CommonConfig` approach with the `Option` because other languages could use it and fallback to the regular name, but it felt like it might be too many changes at the same time.

Makefile
thomastaylor312: Yeah let's pop Makefile changes into a separate PR I think

tests/common.rs
thomastaylor312: Highly recommend just using the tempdir crate instead: https://docs.rs/tempdir/latest/tempdir/ as it comes with a guard that cleans it up

Makefile
vados-cosmonic: :+1: 

- [x] remove Makefile changes

tests/common.rs
vados-cosmonic: Done! I can add `tempdir` instead of doing it manually (this will also remove the need for the `dispose` stuff) if we're OK with this addition in general -- @mattwilkinsonn just want to make sure that this change made sense/is reasonable to add.

One benefit of the manual approach @thomastaylor312 -- I like being able to allow the tests to leave the generated content behind -- for example in the case the test failed -- I'm not sure `tempdir` (now `tempfile`?) offers an ability to *not* clean up given normal execution.

ghost: Personally i'd lean towards using `tempfile::TempDir` and just deleting the directory. I get wanting to look at the generated content but that *should* not be needed, can just add more `assert`s to check what is going on inside the test dir at each step.

vados-cosmonic: I'm open to that -- if something is going wrong in a test it's reasonable for the debug loop can be a bit more manual (or if it's something that should be checked the whole time, then an `assert` should be present)

vados-cosmonic: OK, I've addressed this now -- I've added the field `wasm_bin_name` to `CommonConfig`

thomastaylor312 (dismissed):
@mattwilkinsonn any other thoughts here before merge?

brooksmtownsend:
I'm good with this as long as it handles falling back to the package name if the `[lib]` isn't present, which the code appears to handle! Just wanted to make sure we tested that case.

vados-cosmonic:
Hey @thomastaylor312 one more review please. 

The tests that expect to handle a rust project with no `Cargo.toml` (using the `name` & `version` only) were breaking -- I did a refactor while I was in there, because the fn was too long.

[EDIT] also fixed lint

vados-cosmonic:
[Integration tests that *seem* unrelated are failing](https://github.com/wasmCloud/wash/actions/runs/4299433578/jobs/7494635417) -- investigating:

```
failures:

failures:
error: test failed, to rerun pass `--test integration_reg`
     Running tests/integration_up.rs (target/debug/deps/integration_up-fc47ecdd757b97af)
    integration_pull_basic
    integration_pull_comprehensive
    integration_push_basic
    integration_push_comprehensive

test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.28s


running 3 tests
test doesnt_kill_unowned_nats ... ok
test can_stop_detached_host ... ok
error: 4 targets failed:
test integration_up_can_start_wasmcloud_and_actor ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 35.37s
    `--test integration_claims`

    `--test integration_drain`
    `--test integration_par`
    `--test integration_reg`
```

ghost (dismissed):
LGTM! The integration test failures are almost certainly flakes btw.

brooksmtownsend (approved):

PR #301, opened 2022-08-30
feature: shell completions

stevelr:
Signed-off-by: stevelr <steve@cosmonic.com>

PR #392, opened 2023-02-16
feat: shell completions

vados-cosmonic:
This PR rebases/rebuilds the work done in https://github.com/wasmCloud/wash/pull/301

The code builds but testing on Windows is still not yet completed (trying this now in a VM)

### DIFF
--- a/src/completions.rs
+++ b/src/completions.rs
@@ -1,0 +1,99 @@
+//! Generate shell completion files
+//!
+use crate::{cfg::cfg_dir, CommandOutput};
+use anyhow::anyhow;
+use anyhow::Result;
+use clap::{Args, Subcommand};
+use clap_complete::{generator::generate_to, shells::Shell};
+use std::collections::HashMap;
+use std::path::PathBuf;
+
+const TOKEN_FILE: &str = ".completion_suggested";
+const COMPLETION_DOC_URL: &str = "https://github.com/wasmCloud/wash/blob/main/Completions.md";
+
+fn instructions() -> String {
+    format!(
+        "For instructions on setting up auto-complete for your shell, please see '{}'",
+        COMPLETION_DOC_URL
+    )
+}
+
+#[derive(Debug, Clone, Args)]
+pub(crate) struct CompletionOpts {
+    /// Output directory (default '.')
+    #[clap(short = 'd', long = "dir")]
+    dir: Option<PathBuf>,
+
+    /// Shell
+    #[clap(name = "shell", subcommand)]
+    shell: ShellSelection,
+}
+
+#[derive(Subcommand, Debug, Clone)]
+pub(crate) enum ShellSelection {
+    /// generate completions for Zsh
+    Zsh,
+    /// generate completions for Bash
+    Bash,
+    /// generate completions for Fish
+    Fish,
+    /// generate completions for PowerShell
+    PowerShell,
+}
+
+/// Displays a message one time after wash install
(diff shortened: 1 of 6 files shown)


vados-cosmonic:
Was able to confirm that the completion works in a windows VM:

wash-completion-on-windows

That said -- I'm not sure the amount of documentation there is normal... Are the other steps (execution policy, profile additions, enabling completion/etc) expected knowledge for the average windows command line user? 

Unfortunately Windows is not my forte but I muddled through it -- could use a line or two on the documentation side to make it easier to actually install this functionality. @brooksmtownsend would you mind?

autodidaddict (dismissed):

src/completions.rs
```diff
@@ -0,0 +1,99 @@
...
+pub(crate) fn first_run_suggestion() -> Result<Option<String>> {
+    let token = cfg_dir()?.join(TOKEN_FILE);
+    if token.is_file() {
+        return Ok(None);
+    }
+    let _ = std::fs::File::create(token).map_err(|e| {
+        anyhow!(
+            "can't create completion first-run token in {}: {}",
+            // unwrap() ok because cfg_dir worked above
+            cfg_dir().unwrap().display(),
```
thomastaylor312, line 54: Nit: I'd just capture the path above `let token` since `.join(TOKEN_FILE)` returns a new `PathBuf`

src/main.rs
```diff
@@ -153,13 +162,28 @@ async fn main() {
                     let mut map = out.map;
                     map.insert("success".to_string(), json!(true));
                     println!("\n{}", serde_json::to_string_pretty(&map).unwrap());
+                    0
                 }
                 OutputKind::Text => {
                     println!("\n{}", out.text);
+                    // on the first non-error, non-json use of wash, print info about shell completions
```
thomastaylor312, line 169: Although I find this helpful, I worry about it in cases where someone is installing wash in pipeline and this prints during the first run on every pipeline. The thing I learned from maintaining Helm is no matter how many computer-friendly outputs you have, people will still try to parse plaintext output

thomastaylor312 (changes requested):
This is pretty much good to go, blocking purely so we can have a discussion on the non-nit comment here. cc @brooksmtownsend @mattwilkinsonn

src/main.rs
vados-cosmonic: I only refactored and rebased this code, but in the interest of pushing this forward -- @brooksmtownsend @mattwilkinsonn any thoughts?

brooksmtownsend: I don't have a strong opinion, it's a nice touch cause people probably wouldn't know about it otherwise. I'll defer to @thomastaylor312, to be fair we probably will use `wash` in CI but it's more likely that the proper strategy is to use JSON output

thomastaylor312: Since this is just a refactor and rebase, not gonna block on it.

thomastaylor312 (dismissed):

thomastaylor312:
@vados-cosmonic I'll merge this after rebase!

brooksmtownsend (dismissed):

vados-cosmonic:
@thomastaylor312 done, thanks for taking another look :bow:

brooksmtownsend (dismissed):

vados-cosmonic:
Rebased to match changes on `main`!

thomastaylor312 (approved):

PR #396, opened 2023-02-18
Add `nextest` as the test runner for all wash tests.

ghost:
This will have an improvement on the speed of our test runs (due to [nextest's design](https://nexte.st/index.html)), but the main benefit is that it lets us automatically retry tests locally and on CI. This fixes a pain point for many `wash` PRs lately where one of the integration tests flakes out. Retrying reduces the impact of these flakes significantly.


README.md
```diff
@@ -121,3 +121,11 @@ Perform validation checks on .smithy models, ensuring that your interfaces are v
 ## Contributing to wash
 
 If you have any feature suggestions, find any bugs, or otherwise have a question, please submit an issue [here](https://github.com/wasmcloud/wash/issues/new/choose). Forking & submitting Pull Requests are welcome, and the [good first issue](https://github.com/wasmcloud/wash/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) label is a great way to find a place to start if you're looking to contribute.
+
+### Setting up for wash development
+
+Developing wash and running it's integration tests requires a number of extra compilers and build tools. The `./tools/deps_check.py` script will check you have all of the requirements to develop wash successfully! You can also run this script using `make check-deps`.
```
brooksmtownsend: I would rather make this purely Makefile driven rather than asking for Python as well (even though it's installed on many systems)

tools/check_deps.py
```diff
@@ -0,0 +1,29 @@
...
+
+targets = subprocess.run("rustup target list --installed", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True).stdout
+if "wasm32-unknown-unknown" not in targets:
+    print('Rust wasm32-unknown-unknown target not found. Installing..."')
+    subprocess.run('rustup target add wasm32-unknown-unknown', shell=True)
+
+nextest_output = subprocess.run("cargo nextest --version", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True).stdout
+if "error: no such command" in nextest_output:
+    print('cargo nextest not found. Installing..."')
+    subprocess.run('cargo install cargo-nextest --locked', shell=True)
```
brooksmtownsend: Though I like the inclusion of more tech here, I don't want to give the impression people _have_ to install all of these things to contribute to `wash`. All people really need is a Rust toolchain

Makefile
```diff
@@ -29,6 +29,9 @@ clean:
 	cargo clean
 	wash drain all
 
+# Checks dependencies necessary for developing wash.
+check-deps:
+	python3 tools/check_deps.py
```
brooksmtownsend: I like this rule, would just prefer it checks deps within the Makefile

Makefile
```diff
@@ -5,12 +5,12 @@
...
 test: ## Run unit test suite
-	cargo test --no-fail-fast --bin wash -- --nocapture
-	cargo test --no-fail-fast -p wash-lib -- --nocapture
+	cargo nextest run --no-fail-fast -p wash-lib
+	cargo nextest run --no-fail-fast --bin wash
 
 test-integration: ##Run integration test suite
 	docker compose -f ./tools/docker-compose.yml up --detach
-	cargo test --no-fail-fast --test "integration*" -- --nocapture
+	cargo nextest run --profile integration -E 'kind(test)'
```
brooksmtownsend: I like the promise of `nextest`, though I would rather either reserve separate rules for it or keep it optional in order to run tests locally

brooksmtownsend (changes requested):
I'm fan of the nextest promise! Requests are to keep tools and scripts simple

tools/check_deps.py
ghost: I could clarify that in the README that this is just if you want to run the integration tests?

README.md
ghost: TBH I super hate writing Make for anything beyond just dispatching to other CLI tools and I think it's really hard to read at the best of times. I also think that Python is installed by default on most major OS's at this point, even Windows gives you a prompt to install it if you try to run `python` from the command line. This is also an optional script, we could just also list the dependencies in the README section if someone doesn't want to run it / doesn't have Python installed

Makefile
ghost: IMO I don't think it's necessary to support both `cargo test` and `nextest`. I get the annoyance of a first-time contributor having to install `nextest` first, but that will quickly become much more painful for us as maintainers if a test starts failing on one and not the other.

If the end decision is to just stick with `cargo test`, that's also fine with me but I'd rather not support both

README.md
ghost: All that is to say, please don't make me write Make 😂

But I think that my point still stands on Python being installed by default and being easier to maintain

tools/check_deps.py
ghost: Resolving the readme/python script related comments as those were moved to #403.

Makefile
thomastaylor312: I just worry about having yet another tool for users to install. Example from my past in Krustlet. We used `just` instead of `make` because we really liked it, but it was such a confusing thing for people. There were even people who would straight give up rather than submitting a small patch because they had to install another tool. The ideal would be to support both, but I also understand the concern there.

Any ideas we could use to maybe mitigate some of the "yet another tool" pain?

ghost: I was hoping the added info in #403 as well as the deps check script would ease this pain a little, but I see the "just one more tool" pain point.

We could just support both but there's a key difference in that `nextest` doesn't use the `#[serial]` macro to define tests that must run serially, you have to define a [test group](https://nexte.st/book/test-groups.html) to do so. This would get very confusing if someone adds a new serial-only test and uses `cargo test` locally. They'd just need the `#[serial]` macro to get the test to pass on their machine but then it would fail on CI because the test group isn't set up. Worried we'd have this issue often (as we have a lot of network-reliant tests in `wash`) and also any other differences that might come up between the two.

It kinda comes down to how much we value automatic retries vs having to install a different test runner. Personally i'm on the side of asking contributors to install `nextest` as we have a lot of network/IO reliant tests that are likely to flake out and we're probably only going to add more, but will defer to everyone's opinion

brooksmtownsend (dismissed):
@mattwilkinsonn would you mind just adding a snippet into the README or a `make` rule or something that can prompt people to install nextest?

ghost:
> @mattwilkinsonn would you mind just adding a snippet into the README or a `make` rule or something that can prompt people to install nextest?

Done! Added to `DEVELOPMENT.md`. I messed around with trying to check if it was there in the Makefile but I don't think there's an easy way to do it

brooksmtownsend (approved):

PR #397, opened 2023-02-18
Add --kill-in-use flag to `wash up`.

ghost:
Adds a flag to `up` to kill any processes using the requested wasmCloud & NATS ports. This is especially useful during testing where a test might fail to clean up a server it started, resulting in the next test failing. Now the next test can just kill off the old process, but making it a CLI flag also gives this benefit to end users!

### DIFF
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -5,6 +5,7 @@ use std::os::unix::prelude::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::{ffi::OsStr, io::Cursor};
+use std::{thread, time};
 use tokio::fs::{create_dir_all, metadata, write, File};
 use tokio::process::{Child, Command};
 use tokio_stream::StreamExt;
@@ -309,7 +310,12 @@ jetstream {{
 /// * `bin_path` - Path to the nats-server binary to execute
 /// * `stderr` - Specify where NATS stderr logs should be written to. If logs aren't important, use std::process::Stdio::null()
 /// * `config` - Configuration for the NATS server, see [NatsConfig] for options. This config file is written alongside the nats-server binary as `nats.conf`
-pub async fn start_nats_server<P, T>(bin_path: P, stderr: T, config: NatsConfig) -> Result<Child>
+pub async fn start_nats_server<P, T>(
+    bin_path: P,
+    stderr: T,
+    config: NatsConfig,
+    kill_in_use: bool,
+) -> Result<Child>
 where
     P: AsRef<Path>,
     T: Into<Stdio>,
@@ -319,11 +325,27 @@ where
         .await
         .is_ok()
     {
-        return Err(anyhow!(
-            "Could not start NATS server, a process is already listening on {}:{}",
-            config.host,
-            config.port
-        ));
+        if !kill_in_use {
+            return Err(anyhow!(
+                "Could not start NATS server, a process is already listening on {}:{}. Use the --kill-in-use flag to kill the process listening on this port",
+                config.host,
+                config.port
+            ));
+        }
+
+        port_killer::kill(config.port)?;
+        // Port killing isn't an exact science. Wait a few seconds to ensure the port is free.
+        thread::sleep(time::Duration::from_secs(3));
+        if tokio::net::TcpStream::connect(format!("{}:{}", config.host, config.port))
+            .await
+            .is_ok()
+        {
+            return Err(anyhow!(
+                "Could not start NATS server, unable to kill process is already listening on {}:{}",
+                config.host,
+                config.port
+            ));
+        }
     }
     if let Some(parent_path) = bin_path.as_ref().parent() {
         let config_path = parent_path.join(NATS_SERVER_CONF);
@@ -420,8 +442,13 @@ mod test {
         let log_file = tokio::fs::File::create(&log_path).await?.into_std().await;
 
         let config = NatsConfig::new_standalone("127.0.0.1", 10000, None);
-        let child_res =
-            start_nats_server(&install_dir.join(NATS_SERVER_BINARY), log_file, config).await;
+        let child_res = start_nats_server(
+            &install_dir.join(NATS_SERVER_BINARY),
+            log_file,
+            config,
+            /* kill_in_use= */ true,
+        )
+        .await;
         assert!(child_res.is_ok());
 
         // Give NATS max 5 seconds to start up
@@ -461,6 +488,7 @@ mod test {
             &install_dir.join(NATS_SERVER_BINARY),
             std::process::Stdio::null(),
             config.clone(),
+            /* kill_in_use= */ true,
         )
         .await;
         assert!(nats_one.is_ok());
@@ -469,7 +497,13 @@ mod test {
         tokio::time::sleep(std::time::Duration::from_millis(5000)).await;
         let log_path = install_dir.join("nats.log");
         let log = std::fs::File::create(&log_path)?;
-        let nats_two = start_nats_server(&install_dir.join(NATS_SERVER_BINARY), log, config).await;
+        let nats_two = start_nats_server(
+            &install_dir.join(NATS_SERVER_BINARY),
+            log,
+            config,
+            /* kill_in_use= */ true,
+        )
+        .await;
         assert!(nats_two.is_err());
 
         nats_one.unwrap().kill().await?;
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -501,7 +501,7 @@ mod test {
             &install_dir.join(NATS_SERVER_BINARY),
             log,
             config,
-            /* kill_in_use= */ true,
+            /* kill_in_use= */ false,
         )
         .await;
         assert!(nats_two.is_err());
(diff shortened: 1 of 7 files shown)


crates/wash-lib/src/start/nats.rs
```diff
@@ -319,11 +325,27 @@ where
...
-        ));
+        if !kill_in_use {
+            return Err(anyhow!(
+                "Could not start NATS server, a process is already listening on {}:{}. Use the --kill-in-use flag to kill the process listening on this port",
+                config.host,
+                config.port
+            ));
+        }
+
+        port_killer::kill(config.port)?;
```
thomastaylor312, line 336: I think I'd feel safer with this feature if there was a confirmation dialog. Something like "Found process $Process_path_or_name running. Would you like to kill it? (y|n)". We could couple that with a `--force` flag (bonus points if `--yolo` is an alias) when you really don't care.

thomastaylor312: Related: What happens if this is a docker image running NATS or something on 4222?

thomastaylor312 (commented):
Left some discussion items that also apply to the wasmcloud start stuff

brooksmtownsend (commented):
I'm a little concerned about this feature, and if the biggest issue was leaving things around between test runs I think we should address that with better scripting and cleanup tasks. Just like the consensus in #402 I don't think wash should be responsible or able to kill other processes that it didn't start. For example, if you were to take this feature and run `wash up --nats-port 3000 --kill-in-use` we could kill a completely arbitrary process

crates/wash-lib/src/start/nats.rs
ghost: IMO, `--kill-in-use` is enough to convey that this is gonna kill processes on the requested ports if you pass it. If a user doesn't want that behavior then they can just not pass this flag.

I think the confirmation dialog could be useful when they don't pass the flag. Instead of just instantly failing out if a process is on the port, we print the message you suggested.

Seems like you guys have more reservations about this though. The worst case I can see here is someone forgets they are passing this flag (from reusing the CLI command thru history or copy/paste etc) and accidentally kills off the wasmCloud host and/or NATS server they were using to develop on. I didn't think this would be that painful of an occurrence given you can just start the actors and providers again, but if you guys have a different opinion on this or a worst-er case LMK

On the docker image, i don't know atm. My guess is that it will kill that container, but i'll go check

ghost:
> I'm a little concerned about this feature, and if the biggest issue was leaving things around between test runs I think we should address that with better scripting and cleanup tasks. Just like the consensus in #402 I don't think wash should be responsible or able to kill other processes that it didn't start. For example, if you were to take this feature and run `wash up --nats-port 3000 --kill-in-use` we could kill a completely arbitrary process

The user could also run `kill $(lsof -t -i:3000)` to have the same effect though. I just think its different for `wash` to kill processes it didn't start without being told to do so by the user, vs the user explicitly passing a flag telling `wash up` to kill processes on the ports it needs to start.

thomastaylor312:
Yeah, after thinking about it some more, I agree more with @brooksmtownsend. I don't think I've seen a feature like this before in the CLIs I've used (doesn't mean that there aren't any) and it feels more like testing script cleanup being better

ghost:
> Yeah, after thinking about it some more, I agree more with @brooksmtownsend. I don't think I've seen a feature like this before in the CLIs I've used (doesn't mean that there aren't any) and it feels more like testing script cleanup being better

Anecdotal but I do use this same feature in a CLI for my day job. I think it's quite convenient to use when you don't really care about any old development instances you might be running and just want to start with a clean slate. 

I do think we should set up better "clean up" practices for the integration tests though. I'll open a separate issue for that.

I think the flag is a nice-to-have addition, so not really tied to it if everyone isn't in favor, but I'm not really understanding what the negative impacts of adding it are atm. Could you guys elaborate on what those could be?

thomastaylor312:
My main concern is just around the footgun this gives to users to unintentionally kill something they didn't mean to. For example, if someone is debugging a wasmCloud host they are running in production and they have that command in their history. They could end up killing the host (not too terrible, but could be annoying for debugging) or killing their NATS server (could be real bad). I don't think that will be particularly common, but that could really ruin someone's day

brooksmtownsend:
> Anecdotal but I do use this same feature in a CLI for my day job. I think it's quite convenient to use when you don't really care about any old development instances you might be running and just want to start with a clean slate.

I agree with this! My concern was more that this just checks a port, rather than looking for any possible old _wash_ launched resources. If this PR checked the pid files that we output to ~/.wash/downloads, I'd be much more in favor

ghost:
> if someone is debugging a wasmCloud host they are running in production and they have that command in their history. They could end up killing the host (not too terrible, but could be annoying for debugging) or killing their NATS server (could be real bad).

I didn't really consider that people might be debugging on the same machine they are running the prod NATS server. That seems kind of crazy but I see how adding this flag could be bad in that case!

> I agree with this! My concern was more that this just checks a port, rather than looking for any possible old wash launched resources. If this PR checked the pid files that we output to ~/.wash/downloads, I'd be much more in favor

Oh that's a good idea! I'll see if I can change it to do that. Might need to fork the `port_killer` crate to add a more specific API or find another crate

brooksmtownsend:
@mattwilkinsonn I think the biggest issue as it stands today is that this can be used to kill _any_ process listening on any port, so long as it's specified on the NATS port (which doesn't have to be 4222 since it's user input)

ghost:
> @mattwilkinsonn I think the biggest issue as it stands today is that this can be used to kill any process listening on any port, so long as it's specified on the NATS port (which doesn't have to be 4222 since it's user input)

I don't agree with this being an issue. A user can kill whatever process they want with `kill`, this is just a shorthand to do so within `wash up` to make starting a development/testing environment easier.

But I think there's an important use case difference to understand here. Is `wash up` intended to be used to bootstrap a production wasmCloud instance? I was under the impression it was our equivalent to `docker compose up` (used for local development but not in production). I get the fears of a footgun if this is "used in production" and I'll just close this PR if that's the case lol. But if it's just a development tool I don't understand the fear of killing processes on a development machine.

crates/wash-lib/src/start/nats.rs
brooksmtownsend: Did we figure out what happens here regarding the docker image?

brooksmtownsend:
> A user can kill whatever process they want with kill, this is just a shorthand to do so within wash up to make starting a development/testing environment easier.

I think the key is that it becomes a little too easy with this flag, e.g. if someone were to select port 5000 on a mac, they might end up killing (or trying to) the Airdrop utility that runs on that port and then after trying to use `wash` their airdrop doesn't work.

I don't have a huge issue with this, though I think it's a little hit-a-small-nail-with-a-sledgehammer and most dev tools will fail and let you know there's a process listening on another port. `npm` and React hot reloading do this btw, they let you know what process is listening on the port you're trying to use and it asks if you want to try a different port. As it stands today a script like `wash down; wash up` should take care of dropping any leftover dev resources, and I think if it's not sufficient we should update `wash down`

ghost:
> > A user can kill whatever process they want with kill, this is just a shorthand to do so within wash up to make starting a development/testing environment easier.
> 
> I think the key is that it becomes a little too easy with this flag, e.g. if someone were to select port 5000 on a mac, they might end up killing (or trying to) the Airdrop utility that runs on that port and then after trying to use `wash` their airdrop doesn't work.
> 
> I don't have a huge issue with this, though I think it's a little hit-a-small-nail-with-a-sledgehammer and most dev tools will fail and let you know there's a process listening on another port. `npm` and React hot reloading do this btw, they let you know what process is listening on the port you're trying to use and it asks if you want to try a different port. As it stands today a script like `wash down; wash up` should take care of dropping any leftover dev resources, and I think if it's not sufficient we should update `wash down`

That's fair, its definitely a sledgehammer approach. Personally I want a sledgehammer approach most of the time because if i'm actively working on developing something I just want it to start, regardless of whatever else is running on the ports on my computer. Not everyone thinks like this though lol.

But I do think the opt-in nature lets people choose if they want this or not, so we're not imposing this on people who don't want it. We're really just talking about edge cases here where someone has already opted in and then forgot they did so when rerunning the command from history, but really didn't want to kill something this time, and the thing they didn't want to kill is running on one of the exact 2 ports wasmCloud needed.

I personally don't think this edge case is going to be common, and when it does happen, be particularly painful for the user. However I'll also defer to the broader consensus on this, it's not something really worth continuing to go back and forth about given this is just a nice-to-have feature. If you guys still have objections i'll just close this one for now.

thomastaylor312:
Re this:

> I didn't really consider that people might be debugging on the same machine they are running the prod NATS server. That seems kind of crazy but I see how adding this flag could be bad in that case!

Wash can be used for both local dev and prod work IMO (just like many other tools, think `kubectl`, `nomad`, etc). People will likely use the `wash app` and `wash ctl` commands quite often in prod, even if they might not use `wash up`. 

For sake of making this easy, I propose the following: We merge this (preferably with the pid file changes that @brooksmtownsend suggested) and if for some reason we see footgun instances we can remove since we're pre-1.0. If we were in a 1.0 state where we couldn't remove until breaking version, I'd feel a bit more cautious, but since we aren't I think it is ok to try

brooksmtownsend:
Gonna close this one since it's getting a little stale, but feel free to re-open if we can agree on the discussion above

PR #398, opened 2023-02-18
Fix `wash up --nats-connect-only` killing NATS servers it didn't start

### LINKED ISSUE #388: [BUG] wash holdeth no prisioners!!!!  (it kills all the nats)
**Describe the bug**
wash is killing off nats-servers too aggressively.  If it can not find the pid file, its sending a blanket stop signal, which will shut down a nats-server that is not managed by `wash`

**To Reproduce**

1. `nats-server -js`
2. [different window] `wash up --nats-connect-only`
3. [in wash window] ctl-c
4. nats-server shuts down

**Expected behavior**
Do no kill externally managed nats servers when shutting down wasmCloud

**Screenshots**
Check thread...theres a video!

 - OS: MacOS
 - wash Version 15.1

**Additional context**
🧵 https://wasmcloud.slack.com/archives/CS38R7N9Y/p1676411856645259

ghost:
Fixes #388.

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -320,7 +320,9 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
         Ok(child) => child,
         Err(e) => {
             // Ensure we clean up the NATS server if we can't start wasmCloud
-            stop_nats(install_dir).await?;
+            if !cmd.nats_opts.connect_only {
+                stop_nats(install_dir).await?;
+            }
             return Err(e);
         }
     };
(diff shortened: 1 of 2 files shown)


brooksmtownsend (approved):
Good stuff! I think with this a `wash down` will still take out a NATS server, but that should be intended and I don't see as much of an issue with it

ghost:
> Good stuff! I think with this a `wash down` will still take out a NATS server, but that should be intended and I don't see as much of an issue with it

I think a user would expect `wash down` to only stop the stuff that `wash up` started? But that's complicated for us and relies on us using a file or some other kind of state to track what we started. I'll file this as an issue and we can think about it more in there

PR #399, opened 2023-02-18
Use exact imports instead of globs

ghost:
This originally started because I was finding it hard to track down what `wash up` was using from `wash_lib::start`, but I ended up doing a quick pass through all the globs and removing the ones where I think it is more appropriate to directly import what is needed.

### DIFF
--- a/crates/wash-lib/src/cli/claims.rs
+++ b/crates/wash-lib/src/cli/claims.rs
@@ -19,7 +19,7 @@ use term_table::{
     Table,
 };
 use wascap::{
-    caps::*,
+    caps::capability_name,
     jwt::{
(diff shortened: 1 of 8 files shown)


stevelr (approved):
Much better. IMO avoiding globs in `use` statements should be standard practice, with rare exceptions.

ghost:
> Much better. IMO avoiding globs in `use` statements should be standard practice, with rare exceptions.

Agreed! I think we should only do it in local scoped packages with tons of const imports like `crates/wash-lib/src/generate/mod.rs` importing the consts from `project_variables`. Not sure if we can enforce that as a lint or something though

PR #395, opened 2023-02-18
Refactor `integration_build` tests and remove `cmd_lib` macro.

ghost:
Fixes #394.

The macro wasn't really adding any value and was obscuring why the test was failing. Reverted back to the normal Command usage. I also split these tests out again - I think this is easier to read than passing everything into the same function.

PR #403, opened 2023-02-21
Add Development Prerequistes section to `README` and create `deps_check.py` script.

ghost:
Taken from feedback on #395 & #396, this PR adds a section on what tools are needed to contribute to `wash` and optionally run it's tests.

If we do switch to `nextest` (#396) i'll then update this section in that PR to include the requirement.

I created the table of contents from the [Markdown All In One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) extension in VSCode. Shouldn't be too hard to keep up to date even if you don't have the extension though.


connorsmith256:
Assuming `deps_check` is intended to remain small and only check if deps are available on the path, why not convert it to a shell script? This avoids the need for a user to have python (even if that's unlikely these days), and leverages the strength of a shell script. I.e.

```python
cargo = shutil.which('cargo')
if cargo is None:
    print('cargo not found. Please install Rust from https://rustup.rs/')
    exit(1)
```

Would become:
```bash
command -v cargo >/dev/null 2>&1 || echo "cargo not found. Please install Rust from https://rustup.rs/"
```

tools/deps_check.py
```diff
@@ -0,0 +1,29 @@
...
+    exit(1)
+
+tinygo = shutil.which('tinygo')
+if tinygo is None:
+    print('tinygo not found. Please install it from https://tinygo.org/')
+    exit(1)
+
+targets = subprocess.run("rustup target list --installed", shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True).stdout
+if "wasm32-unknown-unknown" not in targets:
+    print('Rust wasm32-unknown-unknown target not found. Installing..."')
```
rvolosatovs, line 21: This will be redundant if #406 is merged

README.md
```diff
@@ -121,3 +147,19 @@ Perform validation checks on .smithy models, ensuring that your interfaces are v
...
 
 If you have any feature suggestions, find any bugs, or otherwise have a question, please submit an issue [here](https://github.com/wasmcloud/wash/issues/new/choose). Forking & submitting Pull Requests are welcome, and the [good first issue](https://github.com/wasmcloud/wash/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) label is a great way to find a place to start if you're looking to contribute.
+
+### Development Prerequistes
+
+To contribute to `wash`, you just need [Rust](https://rustup.rs/) installed.
+
+#### `build` Integration Tests
+
+However, to run the `wash build` integration tests that compile actors using actual language toolchains, you must have those toolchains installed. Currently the requirements for this are:
```
rvolosatovs: In my experience the integration tests simply fail with no explanation if any of the dependencies are missing, how about improving the error reporting in tests when some dependency is missing, pointing out the missing dependency and add a reference to this section in the doc?

ghost: Yeah we should definitely have better error messages especially when we can't find the `tinygo` toolchain. I'll make that a separate PR though to not pollute this one too much

ghost:
> Assuming `deps_check` is intended to remain small and only check if deps are available on the path, why not convert it to a shell script? This avoids the need for a user to have python (even if that's unlikely these days), and leverages the strength of a shell script. I.e.
> 
> ```python
> cargo = shutil.which('cargo')
> if cargo is None:
>     print('cargo not found. Please install Rust from https://rustup.rs/')
>     exit(1)
> ```
> 
> Would become:
> 
> ```shell
> command -v cargo >/dev/null 2>&1 || echo "cargo not found. Please install Rust from https://rustup.rs/"
> ```

I agree we *could* do this, but the Python version is much more readable and editable. Python is also installed by default on Macs, Ubuntu, and has 1 click installation on Windows. It's also slightly better cross-platform considering `bash` isn't very well supported on Windows.

As a supporting case - the [Rust language repo](https://github.com/rust-lang/rust) also uses Python to setup its development environment & dependencies, so this isn't without precedent.

I'll defer to a majority opinion though! I just think the benefits of Python scripts outweigh those of shell scripts most of the time.

tools/deps_check.py
ghost: What's the error message look like with that? Or does it just install the target automatically?

thomastaylor312:
My main concern is basically the same as stated in https://github.com/wasmCloud/wash/pull/396#discussion_r1117727579. 

However, because this is offered as an optional script, I actually think I am ok with it. We are basically saying "hey, we have this handy script that checks everything for you if you want to use it!" which I think is acceptable

brooksmtownsend (commented):
I still think that this python script would be trivial to create as a Makefile as it stands today, it's even fewer lines of code with @connorsmith256 's suggestion. 

Edit: Actually, given the callout for Windows, I think that installing a python environment is actually simpler than trying to get GNU make to work on Windows. I'll approve this and we can reevaluate the languages and tooling if there are issues down the line. I would rather have it all in one place and keep the tooling requirements simple, so let's just keep that in mind as we go forward. For development purposes, I would rather lean on as much Rust-based tooling as possible

brooksmtownsend (dismissed):

thomastaylor312 (dismissed):

tools/deps_check.py
rvolosatovs: `rustup` will make sure this target is available

brooksmtownsend (approved):

PR #406, opened 2023-02-27
build(rust): add `rust-toolchain.toml`

rvolosatovs:
# Feature or Problem

`rust-toolchain.toml` is the standard Rust toolchain specification file, containing, among other things, the desired Rust toolchain channel, targets and components required for development

# Related Issues

# Release Information

# Consumer Impact

# Testing

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

## Platforms

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,6 @@
+[toolchain]
+channel = "stable"
+components = ["clippy", "rustfmt"]
+targets = [
+    "wasm32-unknown-unknown"
+]


rust-toolchain.toml
```diff
@@ -0,0 +1,6 @@
+[toolchain]
+channel = "stable"
+components = ["clippy", "rustfmt"]
+targets = [
+    "wasm32-unknown-unknown"
+]
```
brooksmtownsend, line 6: I don't think this should be `wasm32-unknown-unknown`, unless I misunderstand the purpose of the targets. We build `wash` for a variety of Native targets but not for Wasm

brooksmtownsend (changes requested):

rust-toolchain.toml
rvolosatovs: I took it from https://github.com/wasmCloud/wash/pull/403/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R160

This does not specify the targets that `wash` is *built for*, but rather the targets that should be *available*

brooksmtownsend (approved):

rust-toolchain.toml
brooksmtownsend: Perfect, that's what I was concerned about

ghost (approved):
LGTM!

Also kinda unrelated but imagine if `nextest` could be added as a `component` here 🤔, would fix the concerns in #396

rvolosatovs:
Looks like the CI still needs a push (same flake as in #407)

thomastaylor312:
Because this is a known flake we are working on and this file doesn't touch code, I am going to merge as is

PR #420, opened 2023-02-27
Adds in a longer wait time for starting an actor

thomastaylor312:
# Feature or Problem

We havin' lotsa flakes with this test, probably thanks to slow GH runners. So this is an attempt to rectify that situation

# Release Information

Pipeline fix, not tied to release

# Consumer Impact

Nothing

# Testing

Just run `make test-all`

## Acceptance or Integration

Small changes to the `wash up` tests

## Manual Verification

Running `make test-all`

### DIFF
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -158,10 +158,10 @@ where
     // Look for nats-server binary and only extract that
     let mut entries = nats_server.entries()?;
     while let Some(res) = entries.next().await {
-        let mut entry = res.map_err(|_e| {
+        let mut entry = res.map_err(|e| {
             anyhow!(
-                "Failed to retrieve file from archive, ensure NATS server {} exists",
-                version
+                "Failed to retrieve file from archive, ensure NATS server {} exists. Original error: {}",
+                version, e
             )
         })?;
         if let Ok(tar_path) = entry.path() {
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -314,18 +314,16 @@ where
     P: AsRef<Path>,
     T: Into<Stdio>,
 {
+    let host_addr = format!("{}:{}", config.host, config.port);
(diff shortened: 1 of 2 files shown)


brooksmtownsend (dismissed):

ghost (dismissed):
🍃

brooksmtownsend (dismissed):

crates/wash-lib/src/start/nats.rs
```diff
@@ -362,6 +361,25 @@ where
         .map_or(false, |m| m.is_file())
 }
 
+async fn wait_for_server(url: &str) -> Result<()> {
+    let mut wait_count = 1;
+    loop {
```
vados-cosmonic, line 366: How do you feel about [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) here? Would it make sense?

thomastaylor312: I think it would just complicate the function in this instance. We just want to try 10 times and then bail. Adding timeout just removes the wait count, but nests the loop inside of a future and you might not actually get 10 retries as it doesn't account for the tcp socket connect time. Either would work fine, but I think this is clear enough

thomastaylor312: I could also switch this to a `for` loop to make it even cleaner in hindsight

crates/wash-lib/src/start/nats.rs
```diff
@@ -362,6 +361,25 @@ where
...
+        // Magic number: 10 + 1, since we are starting at 1 for humans
```
brooksmtownsend, line 367: will need to reconsider magic number if non-humans end up contributing, but I'm willing to approach that later

brooksmtownsend (approved):

crates/wash-lib/src/start/nats.rs
vados-cosmonic: ah thanks for explaining, this makes sense -- it's not adding a ton of value here when the goal is to wait a given number of times. I was thinking more of a "try for this amount of time" behavior.

PR #428, opened 2023-02-28
Add rust cache to try and speed up ci

brooksmtownsend:
# Feature or Problem
Sometimes the CI can be pretty slow running, especially when tests flake or small changes need to run through our 7-13 minute suite. This PR attempts to add rust caching to the process to speed up feedback times.

The initial action run of this PR took 8 minutes. https://github.com/wasmCloud/wash/actions/runs/4294856883
The second run with a small fix and cache took 4 minutes. https://github.com/wasmCloud/wash/actions/runs/4295250596

edit:
After the nextest and integration test things, it still improved from 12 minutes to ~8 minutes

# Related Issues
None

# Release Information
N/A

# Consumer Impact
Contributors may be shocked at faster CI times

# Testing
Testing just involves running through CI, so doing that now

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [x] x86_64-darwin
- [ ] aarch64-darwin
- [x] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [x] x86_64-darwin
- [ ] aarch64-darwin
- [x] x86_64-windows

## Unit Test(s)
Simply changed tests to first run builds in a separate cache step

## Acceptance or Integration
Simply changed tests to first run builds in a separate cache step

## Manual Verification
- [ ] Tests passing with at least a few minutes quicker turnaround is the verification step here


.github/workflows/rust_ci.yml
```diff
@@ -14,7 +14,9 @@ jobs:
     name: Fmt & Clippy Checks
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+      # Cache Rust builds between runs
+      - uses: Swatinem/rust-cache@v2
```
thomastaylor312, line 19: Is the reason we are using this one due to ease of use? Rust caching can be done with the built-in cache action: https://github.com/actions/cache/blob/main/examples.md#rust---cargo

thomastaylor312 (dismissed):
Approved with one minor conversation point

.github/workflows/rust_ci.yml
brooksmtownsend: I just found that one and really enjoyed the fact it can auto discover targets and be a one-line addition in most places. It uses the builtin cache under the hood, but I'd be happy to consider just using that if there's a concern

thomastaylor312: Given the popularity and use of the third party one, I'm not too worried. Just wanted to ask the question

brooksmtownsend: Yeah very fair, I didn't know it was that simple to go with the built in one. Given the popularity and ease I'll keep this action

.github/workflows/rust_ci.yml
```diff
@@ -27,6 +29,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+      # Cache Rust builds between runs
+      - uses: Swatinem/rust-cache@v2
+        with:
+          # Additional key to ensure OS specific caches are separated
```
connorsmith256, line 35: Do we not need this additional key in the other places we cache?

brooksmtownsend: What other places? Might just not understand the question

connorsmith256: Looks like we use this action 3 times, but only one of them has this extra config. Just wondering why it wasn't necessary in the other 2 jobs (`fmt_clippy` and `integration_tests`)

brooksmtownsend: Oh yeah! in those two places we only run it on ubuntu. It wouldn't hurt to add it, just not needed

connorsmith256: Thanks, that's what I was missing :)

ghost (approved):
vroom vroom

(action lgtm, nice and simple!)

Issue #405, opened 2023-02-26
Ensure test resources get cleaned up from integration tests, and don't use `defer!` to do it


ghost:
From discussion on #397, we should make sure all integration tests clean up their resources before (or when) exiting to not cause flaky failures in other tests. #395 also accomplishes this to some extent by using the `tempfile::TempDir` struct for test directories.

The "quick and dirty" approach to this is to use a `defer!` macro to run some code before the test exists, regardless if it panic'ed or not. This is a bit of a code smell though, 1. because it's a macro and 2. you could simply forget to use it or update the existing one when making changes. We should prefer to use `tempfile::TempDir` or a custom type with a custom `Drop` implementation. However given our flakiness situation I'm preferring `defer!` for now, and then will move to the better approach in follow up refactors.

thomastaylor312:
One suggestion is to create a cleanup guard that can call `wash down` for us. I did a similar pattern with that here: https://github.com/wasmCloud/examples/blob/1ce69e6371826a0395d345c78f9ab43901f000be/actor/blobby/testing/tests/setup.rs#L15

PR #441, opened 2023-03-02
Always clean up test directories in `integration_par` and add extra assert for the `reg pull`.

ghost:
Related to #405.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -863,6 +863,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "5.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "907076dfda823b0b36d2a1bb5f90c96660a5bbcd7729e10727f07858f22c4edc"
+dependencies = [
+ "cfg-if",
+ "hashbrown",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "data-encoding"
 version = "2.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


autodidaddict:
This is likely a matter of personal language preference, but the `defer!` macro feels like it has the potential to confuse readers who have experience with different languages, namely Go. Not everyone is going to know what it does or how it works on a quick glance, or that it works differently than Go's defer, which makes people like me more likely to use it the wrong way 😁 

Is there something more idiomatically "Rusty" that we can do here instead like implementing `Drop` or maybe there's a hook we can use to attach into the Rust test runner?

ghost:
> This is likely a matter of personal language preference, but the `defer!` macro feels like it has the potential to confuse readers who have experience with different languages, namely Go. Not everyone is going to know what it does or how it works on a quick glance, or that it works differently than Go's defer, which makes people like me more likely to use it the wrong way 😁
> 
> Is there something more idiomatically "Rusty" that we can do here instead like implementing `Drop` or maybe there's a hook we can use to attach into the Rust test runner?

Agreed 100%. All my homies hate `defer!` and macros in general. I'm just trying to fix these flakes quickly because they are becoming a huge blocker, I'll go and refactor to use `Drop` or `tempfile::TempDir` later

thomastaylor312 (dismissed):
@mattwilkinsonn Let's create an issue so we don't forget to follow up on the defer stuff

ghost:
> @mattwilkinsonn Let's create an issue so we don't forget to follow up on the defer stuff

I updated #405 to reflect this, it's kinda two steps of the same thing. I think I cause too much issue pollution in this repo anyways lol

ghost:
Updated to use `tempfile::TempDir` in everywhere except the cached artifact test, which relies on the directory we use being `/tmp/wasmcloud_ocicache` exactly, so I needed to keep the `defer!`

brooksmtownsend:
Gonna close this one since it's getting a little stale, but feel free to re-open if we can get the conflict fixed and we can agree on the defer stuff

PR #445, opened 2023-03-03
Revert `test_dir` changes and pin `tokio` to =1.24.0

ghost:
Changes to the test dir functions were breaking the `integration_claims` tests, and the `tokio` version update was breaking all of the tests that used `reg pull`. Reverted them both here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,6 +126,7 @@ term-table = "1.3.1"
 test-case = "2.2.1"
 test_bin = "0.4.0"
 thiserror = "1.0"
+# DO NOT UPDATE - https://github.com/wasmCloud/wash/issues/382
 tokio = { version = "=1.24.0", default-features = false }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -126,7 +126,7 @@ term-table = "1.3.1"
 test-case = "2.2.1"
 test_bin = "0.4.0"
 thiserror = "1.0"
-tokio = { version = "1", default-features = false }
+tokio = { version = "=1.24.0", default-features = false }
```
thomastaylor312, line 130: Nit: Probably should add a comment linking back to the weird bug

thomastaylor312 (dismissed):

Cargo.toml
ghost: Done!

brooksmtownsend (approved):

PR #513, opened 2023-04-10
Feat/allow file upload

### LINKED ISSUE #509: [FEATURE] Have `wash up` enable starting from file by default

`wash up` is meant to be a quick start/developer tool. Because of this, it would be really nice if `wash up` by default set `WASMCLOUD_ALLOW_FILE_LOAD` to `1` so that you could start things from file using `wash ctl start actor file://<path>`. 

**Additional context**

This came from my own pain as well as feedback from other users who were trying to start larger actors but hitting the washboard limit

connorsmith256:
## Feature or Problem
Resolves https://github.com/wasmCloud/wash/issues/509 by setting `WASMCLOUD_ALLOW_FILE_LOAD` to true for hosts started via `wash up`

Also fixed a clippy issue

I made the `allow-file-load` flag take an explicit value to account for the default being true. The alternative is to invert the logic in wash and make the flag something like `--disallow-file-load`. If preferred, I can swap that easily

## Related Issues
https://github.com/wasmCloud/wash/issues/509

## Release Information
Next (v0.17.0, since this is a new feature)

## Consumer Impact
After this change, users will be able to `wash up` and then `wash ctl start actor file://foo`

### Manual Verification

Started a host with `wash up` and confirmed `WASMCLOUD_ALLOW_FILE_LOAD` was set and `wash ctl start actor file://foo` returned `:enoent`

Started a host with `wash up --allow-file-load true` and confirmed the same behavior

Started a host with `wash up --allow-file-load false` and confirmed the env var was not set and the ctl command returned `actor file loading is disabled`

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -225,6 +225,10 @@ pub(crate) struct WasmcloudOpts {
     #[clap(long = "config-service-enabled", env = WASMCLOUD_CONFIG_SERVICE)]
     pub(crate) config_service_enabled: bool,
 
+    /// Denotes if a wasmCloud host should allow starting actors from the file system
+    #[clap(long = "allow-file-load", default_value = DEFAULT_ALLOW_FILE_LOAD, env = WASMCLOUD_ALLOW_FILE_LOAD)]
+    pub(crate) allow_file_load: Option<bool>,
+
     /// Enable JSON structured logging from the wasmCloud host
     #[clap(
         long = "enable-structured-logging",
--- a/crates/wash-lib/src/generate/mod.rs
+++ b/crates/wash-lib/src/generate/mod.rs
@@ -32,8 +32,9 @@ type ParamMap = std::collections::BTreeMap<String, serde_json::Value>;
 const PROJECT_NAME_REGEX: &str = r"^([a-zA-Z][a-zA-Z0-9_-]+)$";
 
 /// Type of project to be generated
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Default, Clone, Copy)]
 pub enum ProjectKind {
+    #[default]
     Actor,
     Interface,
     Provider,
@@ -50,12 +51,6 @@ impl fmt::Display for ProjectKind {
                 ProjectKind::Provider => "provider",
             }
         )
-    }
-}
-
-impl Default for ProjectKind {
-    fn default() -> Self {
-        ProjectKind::Actor
     }
 }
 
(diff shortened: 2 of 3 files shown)


ricochet:
We haven't started using the `deny` verb and `allow` fits, so I prefer what you have in the PR.

```bash
./target/debug/wash up -h | grep allow
      --allow-latest
          Determines whether OCI images tagged latest are allowed to be pulled from OCI registries and started [env: WASMCLOUD_OCI_ALLOW_LATEST=]
      --allowed-insecure <ALLOWED_INSECURE>
          A comma-separated list of OCI hosts to which insecure (non-TLS) connections are allowed [env: WASMCLOUD_OCI_ALLOWED_INSECURE=]
      --allow-file-load <ALLOW_FILE_LOAD>
          Denotes if a wasmCloud host should allow starting actors from the file system [env: WASMCLOUD_ALLOW_FILE_LOAD=] [default: true] [possible values: true, false]
```

ricochet (dismissed):

ricochet:
Integration test failure unrelated to these changes (signing go actors) and has been rep'd in other PR's.

src/up/mod.rs
```diff
@@ -225,6 +225,10 @@ pub(crate) struct WasmcloudOpts {
     #[clap(long = "config-service-enabled", env = WASMCLOUD_CONFIG_SERVICE)]
     pub(crate) config_service_enabled: bool,
 
+    /// Denotes if a wasmCloud host should allow starting actors from the file system
+    #[clap(long = "allow-file-load", default_value = DEFAULT_ALLOW_FILE_LOAD, env = WASMCLOUD_ALLOW_FILE_LOAD)]
+    pub(crate) allow_file_load: Option<bool>,
```
brooksmtownsend, line 230: This should just be a bool, it'll default to false

brooksmtownsend (changes requested):
Requesting one ergonomic change. The desired behavior is that it's denied by default, right?

ricochet:
> Requesting one ergonomic change. The desired behavior is that it's denied by default, right?

The original issue requested enabled by default. https://github.com/wasmCloud/wash/issues/509

src/up/mod.rs
connorsmith256: See @ricochet's comment: I intended this to be allowed by default, per the source issue. Do you think that's a bad idea?

brooksmtownsend (approved):

PR #517, opened 2023-04-12
Bumpity bump

connorsmith256:
Leaving as draft, just checking if the tokio bug with pulling from a registry is still there


connorsmith256:
Closing since the issue is still present

PR #520, opened 2023-04-17
feat(*) wadm 0.4 support in `wash app`

### LINKED ISSUE #382: [BUG] Unable to reg pull properly after `cargo update`
During our preparation for the wash 0.15 release in #381, we noticed that all claims tests were failing intermittently. Turns out if you run `cargo update` against wash right now, it introduces a bug.

All of the claims test do a `wash reg pull`. However, the size of the resulting wasm file it pulls down differs almost every single time, and it is generally smaller than the expected number of bytes. For example, if you run `wash reg pull wasmcloud.azurecr.io/echo:0.2.1 --destination echo.wasm` with `RUST_LOG=debug`, you can see in the logs:

```
[2023-02-09T21:14:16Z DEBUG reqwest::async_impl::client] redirecting 'https://wasmcloud.azurecr.io/v2/echo/blobs/sha256:23efc45c62d316001cb19979de071917afa1122c1f1900eef284b3c031f698a2' to 'https://eusmanaged66.blob.core.windows.net/d397f64bd0994f7da3c826aa2007f401-yk1iax15ek//docker/registry/v2/blobs/sha256/23/23efc45c62d316001cb19979de071917afa1122c1f1900eef284b3c031f698a2/data?se=2023-02-09T21%3A33%3A01Z&sig=gVvvBAIbMK2lUO9cMbfbz0F4nudLS%2BnJfQFOgvC4Kc8%3D&sp=r&spr=https&sr=b&sv=2016-05-31&regid=d397f64bd0994f7da3c826aa2007f401&anon=true'
[2023-02-09T21:14:16Z DEBUG hyper::client::pool] reuse idle connection for ("https", eusmanaged66.blob.core.windows.net)
[2023-02-09T21:14:16Z DEBUG hyper::proto::h1::io] flushed 704 bytes
[2023-02-09T21:14:17Z DEBUG hyper::proto::h1::io] parsed 19 headers
[2023-02-09T21:14:17Z DEBUG hyper::proto::h1::conn] incoming body is content-length (420611 bytes)
```

That content length is the correct size, but after the validation step the size on disk can vary (I got it to be correct about 1 out of every 6 times). This causes the actual signing step to fail when we run the tests.

What makes this worse is that it only happens on linux. We tested this on Mac and Windows and it was fine every time. We tested on the following OS/arch combos:

- M1 Mac (Ventura): Just fine
- Windows 11 amd64: Just fine
- Linux amd64 (Ubuntu 20.04): Things break

We need to figure out which dependency is causing this before we `cargo update` anything

brooksmtownsend:
## Feature or Problem
After the many changes going into `wadm` 0.4, there were a few changes to the app subcommand in order to make it compatible with the new types. Additionally, this PR adds in `wadm` as a binary to download as a part of running `up` so that we can immediately interact with `wadm`.

You should consider reviewing this PR as two big sections: the `wash up` wadm support, and the changes to the `app` module.

This PR
- [x] Adds `wadm` as another binary that can be downloaded + started with `wash up`
- [x] Updates the `app` subcommand to match the Wadm NATS API
- [x] Updates async_nats, tokio, and a few other dependencies after figuring out the issue with the OCI downloads in #382 

## Related Issues
(pr) https://github.com/wasmCloud/wadm/pull/89
Fixes #382

## Release Information
v0.18.0

## Consumer Impact
Consumers of the Elixir implementation of wadm will see breaking changes here, as expected, so with the v0.18.0 release notes we should note that the latest version of wadm should be used.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
Manual tested with https://github.com/wasmCloud/wadm/pull/89

### DIFF
--- a/crates/wash-lib/src/app.rs
+++ b/crates/wash-lib/src/app.rs
@@ -1,0 +1,240 @@
+//! Interact with and manage wadm applications over NATS, requires the `nats` feature
+
+use std::time::Duration;
+
+use anyhow::{bail, Result};
+use async_nats::{Client, Message};
+use wadm::server::{
+    DeleteModelRequest, DeleteModelResponse, DeployModelRequest, DeployModelResponse,
+    GetModelRequest, GetModelResponse, ModelSummary, PutModelResponse, UndeployModelRequest,
+    VersionResponse,
+};
+
+use crate::config::DEFAULT_LATTICE_PREFIX;
+
+/// The NATS prefix wadm's API is listening on
+const WADM_API_PREFIX: &str = "wadm.api";
+
+/// A helper enum to easily refer to wadm model operations and then use the
+/// [ToString](ToString) implementation for NATS topic formation
+pub enum ModelOperation {
+    List,
+    Get,
+    History,
+    Delete,
+    Put,
+    Deploy,
+    Undeploy,
+}
+
+impl ToString for ModelOperation {
+    fn to_string(&self) -> String {
+        match self {
+            ModelOperation::List => "list",
+            ModelOperation::Get => "get",
+            ModelOperation::History => "versions",
+            ModelOperation::Delete => "del",
+            ModelOperation::Put => "put",
+            ModelOperation::Deploy => "deploy",
+            ModelOperation::Undeploy => "undeploy",
+        }
+        .to_string()
+    }
+}
+
+/// Undeploy a model, instructing wadm to no longer manage the given application
+///
+/// # Arguments
+/// * `client` - The [Client](async_nats::Client) to use in order to send the request message
+/// * `lattice_prefix` - Optional lattice prefix that the application is managed on, defaults to `default`
+/// * `model_name` - Model name to undeploy
+/// * `non_destructive` - Undeploy deletes managed resources by default, this can be overridden by setting this to `true`
+pub async fn undeploy_model(
+    client: &Client,
+    lattice_prefix: Option<String>,
+    model_name: &str,
+    non_destructive: bool,
+) -> Result<DeployModelResponse> {
+    let res = model_request(
+        client,
+        ModelOperation::Undeploy,
+        lattice_prefix,
+        Some(model_name),
+        serde_json::to_vec(&UndeployModelRequest { non_destructive })?,
+    )
+    .await?;
+
+    serde_json::from_slice(&res.payload).map_err(|e| anyhow::anyhow!(e))
+}
+
+/// Deploy a model, instructing wadm to manage the application
+///
+/// # Arguments
+/// * `client` - The [Client](async_nats::Client) to use in order to send the request message
+/// * `lattice_prefix` - Optional lattice prefix that the application will be managed on, defaults to `default`
+/// * `model_name` - Model name to deploy
+/// * `version` - Version to deploy, defaults to deploying the latest "put" version
+pub async fn deploy_model(
+    client: &Client,
+    lattice_prefix: Option<String>,
+    model_name: &str,
+    version: Option<String>,
+) -> Result<DeployModelResponse> {
+    let res = model_request(
+        client,
+        ModelOperation::Deploy,
+        lattice_prefix,
+        Some(model_name),
+        serde_json::to_vec(&DeployModelRequest { version })?,
+    )
+    .await?;
+
+    serde_json::from_slice(&res.payload).map_err(|e| anyhow::anyhow!(e))
+}
+
+/// Put a model definition, instructing wadm to store the application manifest for later deploys
+///
+/// # Arguments
+/// * `client` - The [Client](async_nats::Client) to use in order to send the request message
+/// * `lattice_prefix` - Optional lattice prefix that the application manifest will be stored on, defaults to `default`
+/// * `model` - The full YAML or JSON string containing the OAM wadm manifest
+pub async fn put_model(
+    client: &Client,
+    lattice_prefix: Option<String>,
+    model: &str,
+) -> Result<PutModelResponse> {
+    let res = model_request(
+        client,
+        ModelOperation::Put,
+        lattice_prefix,
+        None,
+        model.as_bytes().to_vec(),
+    )
+    .await?;
+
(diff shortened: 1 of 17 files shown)


brooksmtownsend:
https://github.com/wasmCloud/wash/issues/382 is biting me here

src/up/mod.rs
```diff
@@ -397,7 +452,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
             "\n🌐 The wasmCloud dashboard is running at {}\n📜 Logs for the host are being written to {}",
             url, wasmcloud_log_path.to_string_lossy()
         );
-        let _ = write!(out_text, "\n\n🛑 To stop wasmCloud, run \"wash down\"");
+        let _ = write!(out_text, "\n\n⬇️  To stop wasmCloud, run \"wash down\"");
```
brooksmtownsend, line 462: @autodidaddict changed after your comment 😄

crates/wash-lib/src/start/wadm.rs
```diff
@@ -0,0 +1,285 @@
...
+    P: AsRef<Path>,
+    T: Into<Stdio>,
+{
+    let pid_file = bin_path.as_ref().parent().map(|p| p.join(WADM_PID));
+    match pid_file.as_ref() {
+        Some(path) => {
+            if let Ok(previous_pid) = tokio::fs::read_to_string(&path).await {
+                //TODO: bad? What if file just isn't cleared out?
+                // probably should find another way, validate with app API
+                bail!("wadm is already running with pid {previous_pid}, aborting")
```
connorsmith256: IMO this is slightly too aggressive. Consider the case where a user manually killed the wadm process, or restarted their machine, or otherwise shut down wadm without cleaning up the PID file. The next time we try to start wadm on their behalf, it will fail, until they manually clear this PID file.

We cover a similar problem here, with `is_nats_running`: https://github.com/wasmCloud/wash/blob/6c1a94198dfaa65d582e8c040539a009a399f9f2/src/up/mod.rs#L454

There, we collect two bits of info: A) is there a PID file and B) is there a process listening on the expected port. `A && !B` is "okay", in the sense that we should probably ignore the existing PID file and start the process anyway

This logic could/should be generalized across the various processes wash manages, but since the other example is still in PR, we can come back to that later

src/up/mod.rs
```diff
@@ -335,10 +386,16 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
     {
         Ok(child) => child,
         Err(e) => {
-            // Ensure we clean up the NATS server if we can't start wasmCloud
+            // Ensure we clean up the NATS server and wadm if we can't start wasmCloud
             if !cmd.nats_opts.connect_only {
                 stop_nats(install_dir).await?;
             }
+            match (wadm_process, cmd.wadm_opts.disable_wadm) {
```
connorsmith256: Nit: since we only get the process handle if `disable_wadm` is false, I don't think you need the second member of this tuple

connorsmith256 (dismissed):
Approved, but with one suggestion (and a nit)

crates/wash-lib/src/start/wadm.rs
andar1an: From a user perspective - especially if wadm is being folded into wasmcloud host - the process name can be annoying the first time to deal with as it is called beam. If it was possible to update the process name in case a user needed to kill process that would be cool, but I worry if not that Connor's point here is really important.

brooksmtownsend: Yeah I think this is too aggressive, it would be good to try and use the `app` library to query to see if `wadm` is running. Will come up with a solution here

brooksmtownsend: Ended up moving the `app` functionality to `wash-lib` so that I could make a query to see if a wadm is running _anywhere_ 😄

connorsmith256 (dismissed):

crates/wash-lib/src/app.rs
```diff
@@ -0,0 +1,240 @@
...
+/// A helper enum to easily refer to wadm model operations and then use the
+/// [ToString](ToString) implementation for NATS topic formation
+pub enum ModelOperation {
+    List,
+    Get,
+    History,
+    Delete,
+    Put,
+    Deploy,
+    Undeploy,
```
ricochet, line 27: How annoying would it be to change `undeploy` to `destroy`? That's a verb I'm not used to seeing with these kinds of things

ricochet: Or delete/remove. Undeploy isn't a word :/

- `k delete pod foo`
- `vercel remove my-app`
- `flyctl destroy`

connorsmith256: We have `delete`, which undeploys (ignore the red squiggle, it's _totally_ a word!) and removes the model from wash's knowledge

`undeploy` is non-destructive; it "just" tells wadm to stop monitoring the deployment

More here: https://github.com/wasmCloud/wasmcloud.com-dev/blob/main/docs/reference/wadm/usage.md

ricochet: 🎰 ok let's see if it works for folks

ricochet (dismissed):

crates/wash-lib/src/app.rs
brooksmtownsend: The verbage is intentional and I agree that it's not immediately obvious, heck from a wash perspective we could get rid of the verbs entirely and just have "create" and "delete" if we wanted to work around a few edge cases

brooksmtownsend: I think we should keep the app command as experimental and we can def optimize as we find out what feels good

connorsmith256 (approved):

PR #522, opened 2023-04-18
fix(*): Bumps wash-lib version

thomastaylor312:
This was missed and so cargo installing from main causes issues. Also bumps 0.17 so that it can pick up the new version from crates. Once this is published we should yank 0.17.0


brooksmtownsend (approved):
thank you sir 👍🏻

Issue #492, opened 2023-03-23
[BUG] Using --wasmcloud-start-only with wash up does not find locally installed versions


vados-cosmonic:
**Describe the bug**
When trying to test a local `wasmcloud-otp` version, `wash up --wasmcloud-version <local version>` does not find versions that have been installed to `~/.wash/downloads` when `--wasmcloud-start-only` is used.

The [related code](https://github.com/wasmCloud/wash/blob/462364d5d294b3527ebe022f942961a98a63784b/src/up/mod.rs#L289) seems to indicate that using the flag causes a failure to *always* occur.

**To Reproduce**
1. Build a custom version of `wasmcloud-otp` (i.e. run `make release` in `wasmcloud-otp/wasmcloud_host`
1. Install the custom version to `~/.wash/downloads` (ex. `~/.wash/downloads/v0.63.0`)
1. Attempt to run `wash up --wasmcloud-version <local version> --wasmcloud-start-only`

**Expected behavior**
`wash up` should recognize the custom version that has been added to `~/.wash/downloads`, with `--wasmcloud-start-only` specified.

**Screenshots**

N/A

**Environment (please complete the following information) **
 - OS: Linux
 - Shell: Bash
 - `wash` Version `v0.16.2`

**Additional context**

N/A

connorsmith256:
Unmarked as stale. This one is still an issue 🙏

PR #527, opened 2023-04-25
Feat: Resurrect multiple wasmcloud hosts

### LINKED ISSUE #464: [FEATURE] Allow wash up to spin up multiple wasmCloud hosts in quick succession

Right now, when `wash up` runs, it uses the same ports and resources (ports for NATS for example) which means that starting more than one instance requires setting different ENV variables.

Modfiy `wash` such that `wash up` can be run multiple times in quick succession to run multiple wasmCloud hosts.

Alternatively, we could update the documentation to more clearly call out how to use ENV variables properly to run multiple hosts.

**Additional context**

Connor worked on this in the past, so there is some support (ENV vars), may need to be reworked/expanded.

aish-where-ya:
## Feature: Resurrect multiple local wasmcloud hosts successively. 
These hosts can be in detached or interactive mode.
The default NATS port is 4222 and the unless explicitly specified, the hosts will use ports from 4000 to 5000.

Usage: 
To spin a host run `wash up` or `wash up -d`. Specify the host port with `dashboard-port` flag.

To spin a host down, run `wash down --host-id <HOST-ID>` with the respective host-id or `wash down --all` for shutting down all hosts. The host-id can be obtained from `wash ctl get hosts`.

Details:
This host-id is generated using the `WASMCLOUD_HOST_SEED` environment variable. Unless specified by the user, a new n-keys key-pair is generated where the seed is passed on to this variable. A public key is generated and stored as `RELEASE_NODE` which is used to uniquely identify each host. The `RELEASE_NODE` is the host-id that the user gets on running `wash ctl get hosts`. 
When resurrecting consecutive hosts, if a NATS server is already running and a PID file exists, the consecutive hosts attach to this NATS server.

## Related Issues
Fixes #464 

## Release Information
Next

## Consumer Impact
The method to operate `wash up` and `wash down` will change with added options.

## Testing
Locally tested

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
NA

### Acceptance or Integration
NA

### Manual Verification
Manually verified!

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -212,6 +212,26 @@ where
         )),
     }
 }
+
+pub async fn is_wasmcloud_running<P>(bin_path: P, host_id: String) -> Result<bool>
+where
+    P: AsRef<Path>,
+{
+    if let Ok(output) = Command::new(bin_path.as_ref())
+        .arg("pid")
+        .env("RELEASE_NODE", host_id)
+        .output()
+        .await
+    {
+        // Stderr will include :nodedown if no other host is running, otherwise
+        // stdout will contain the PID
+        if String::from_utf8_lossy(&output.stderr).contains(":nodedown") {
+            return Ok(false);
+        }
+    }
+    Ok(true)
+}
+
 /// Helper function to start a wasmCloud host given the path to the elixir release script
 /// /// # Arguments
 ///
@@ -231,10 +251,10 @@ where
     S: Into<Stdio>,
 {
     // If we can connect to the local port, a wasmCloud host won't be able to listen on that port
-    let port = env_vars
-        .get("WASMCLOUD_DASHBOARD_PORT")
-        .cloned()
-        .unwrap_or_else(|| "4000".to_string());
+    let port = match env_vars.get("WASMCLOUD_DASHBOARD_PORT").cloned() {
+        Some(port) => port,
+        None => return Err(anyhow!("wasmCloud host port is not set")),
+    };
     if tokio::net::TcpStream::connect(format!("localhost:{port}"))
         .await
         .is_ok()
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -253,7 +253,7 @@ where
     // If we can connect to the local port, a wasmCloud host won't be able to listen on that port
     let port = match env_vars.get("WASMCLOUD_DASHBOARD_PORT").cloned() {
         Some(port) => port,
-        None => return Err(anyhow!("wasmCloud host port is not set")),
+        None => "4000".to_string(),
     };
     if tokio::net::TcpStream::connect(format!("localhost:{port}"))
         .await
--- a/src/down/mod.rs
+++ b/src/down/mod.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Output, Stdio};
 use std::time::Duration;
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use clap::Parser;
 use serde_json::json;
 use tokio::process::Command;
@@ -13,9 +13,16 @@ use wash_lib::start::{find_wasmcloud_binary, NATS_SERVER_BINARY, NATS_SERVER_PID
 use crate::appearance::spinner::Spinner;
 use crate::cfg::cfg_dir;
 use crate::up::{DOWNLOADS_DIR, WASMCLOUD_PID_FILE};
+use crate::util::get_wasmcloud_hosts;
 
 #[derive(Parser, Debug, Clone)]
-pub(crate) struct DownCommand {}
+pub(crate) struct DownCommand {
+    #[clap(long = "host-id")]
+    pub host_id: Option<String>,
+
+    #[clap(long = "all")]
+    pub all: bool,
+}
 
 pub(crate) async fn handle_command(
     command: DownCommand,
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
     let host_bin = find_wasmcloud_binary(&install_dir, &version)
         .await
         .ok_or_else(|| anyhow::anyhow!("Couldn't find path to wasmCloud binary. Is it running?"))?;
-    if host_bin.is_file() {
-        sp.update_spinner_message(" Stopping host ...".to_string());
-        if let Ok(output) = stop_wasmcloud(host_bin).await {
+
+    let mut stopped_hosts = 0;
+    let hosts = get_wasmcloud_hosts(host_bin.clone()).await?;
+
+    if hosts.is_empty() {
+        sp.println("No local wasmCloud hosts detected.");
+    } else if let Some(host_id) = _cmd.host_id {
+        let host = match hosts.iter().find(|h| h.id == host_id) {
+            Some(host) => host,
+            None => bail!("Host {} is not running", host_id),
+        };
+        sp.update_spinner_message(format!("Stopping host {}...", host.id));
+        if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
             if output.stderr.is_empty() && output.stdout.is_empty() {
                 // if there was a host running, 'stop' has no output.
                 // Give it time to stop before stopping nats
                 tokio::time::sleep(Duration::from_secs(6)).await;
-                out_json.insert("host_stopped".to_string(), json!(true));
-                out_text.push_str("✅ wasmCloud host stopped successfully\n");
+                stopped_hosts += 1;
+                out_json.insert("host_stopped".to_string(), json!(host.id));
+                out_text.push_str(&format!(
+                    "✅ wasmCloud host {} stopped successfully\n",
+                    host_id,
+                ));
             } else {
-                out_json.insert("host_stopped".to_string(), json!(true));
-                out_text.push_str(
-                    "🤔 Host did not appear to be running, assuming it's already stopped\n",
-                );
+                out_json.insert("host_stopped".to_string(), json!(host.id));
+                out_text.push_str(&format!(
+                    "🤔 Host {} did not appear to be running, assuming it's already stopped\n",
+                    host_id
+                ));
+            }
+        };
+    } else if hosts.len() == 1 {
+        let host = hosts.first().unwrap();
+        sp.update_spinner_message(format!("Stopping host {}...", host.id));
+        if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
+            if output.stderr.is_empty() && output.stdout.is_empty() {
+                // if there was a host running, 'stop' has no output.
+                // Give it time to stop before stopping nats
+                tokio::time::sleep(Duration::from_secs(6)).await;
+                stopped_hosts += 1;
+                out_json.insert("host_stopped".to_string(), json!(host.id));
+                out_text.push_str(&format!(
+                    "✅ wasmCloud host {} stopped successfully\n",
+                    host.id,
+                ));
+            } else {
+                out_json.insert("host_stopped".to_string(), json!(host.id));
+                out_text.push_str(&format!(
+                    "🤔 Host {} did not appear to be running, assuming it's already stopped\n",
+                    host.id
+                ));
             }
         }
+    } else if _cmd.all {
+        for (i, host) in hosts.iter().enumerate() {
+            sp.update_spinner_message(format!(
+                "Stopping host {} ({}/{})...",
+                host.id,
+                i + 1,
+                hosts.len()
+            ));
+            if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
+                if output.stderr.is_empty() && output.stdout.is_empty() {
+                    // if there was a host running, 'stop' has no output.
+                    // Give it time to stop before stopping nats
+                    tokio::time::sleep(Duration::from_secs(6)).await;
+                    stopped_hosts += 1;
+                    out_json.insert("host_stopped".to_string(), json!(host.id));
+                    out_text.push_str(&format!(
+                        "✅ wasmCloud host {} stopped successfully\n",
+                        host.id,
+                    ));
+                } else {
+                    out_json.insert("host_stopped".to_string(), json!(host.id));
+                    out_text.push_str(&format!(
+                        "🤔 Host {} did not appear to be running, assuming it's already stopped\n",
+                        host.id
+                    ));
+                }
+            }
+        }
+    } else {
+        bail!(
+            "Detected multiple wasmCloud hosts. Specify a host with --host-id [hostID], or use --all\nTo see your hosts, run wash ctl get hosts."
+        )
     }
 
-    let nats_bin = install_dir.join(NATS_SERVER_BINARY);
-    if nats_bin.is_file() {
-        sp.update_spinner_message(" Stopping NATS server ...".to_string());
-        if let Err(e) = stop_nats(install_dir).await {
-            out_json.insert("nats_stopped".to_string(), json!(false));
-            out_text.push_str(&format!(
-                "❌ NATS server did not stop successfully: {e:?}\n"
-            ));
-        } else {
-            out_json.insert("nats_stopped".to_string(), json!(true));
-            out_text.push_str("✅ NATS server stopped successfully\n");
+    if stopped_hosts == hosts.len() {
+        let nats_bin = install_dir.join(NATS_SERVER_BINARY);
+        if nats_bin.is_file() {
+            sp.update_spinner_message(" Stopping NATS server ...".to_string());
+            if let Err(e) = stop_nats(install_dir).await {
+                out_json.insert("nats_stopped".to_string(), json!(false));
+                out_text.push_str(&format!(
+                    "❌ NATS server did not stop successfully: {e:?}\n"
+                ));
+            } else {
+                out_json.insert("nats_stopped".to_string(), json!(true));
+                out_text.push_str("✅ NATS server stopped successfully\n");
+            }
         }
+    } else {
+        out_json.insert("nats_stopped".to_string(), json!(false));
+        out_text.push_str("NATS server is still running\n");
     }
 
     out_json.insert("success".to_string(), json!(true));
@@ -81,13 +162,14 @@ pub(crate) async fn handle_down(
 }
 
 /// Helper function to send wasmCloud the `stop` command and wait for it to clean up
-pub(crate) async fn stop_wasmcloud<P>(bin_path: P) -> Result<Output>
+pub(crate) async fn stop_wasmcloud<P>(bin_path: P, host_id: Option<String>) -> Result<Output>
 where
     P: AsRef<Path>,
 {
     Command::new(bin_path.as_ref())
         .stdout(Stdio::piped())
         .arg("stop")
+        .env("RELEASE_NODE", host_id.unwrap_or("".to_string()))
         .output()
         .await
         .map_err(anyhow::Error::from)
--- a/src/down/mod.rs
+++ b/src/down/mod.rs
@@ -8,15 +8,49 @@ use clap::Parser;
 use serde_json::json;
 use tokio::process::Command;
 use wash_lib::cli::{CommandOutput, OutputKind};
-use wash_lib::start::{find_wasmcloud_binary, NATS_SERVER_BINARY, NATS_SERVER_PID};
+use wash_lib::config::{DEFAULT_NATS_HOST, DEFAULT_NATS_PORT};
+use wash_lib::start::{find_wasmcloud_binary, nats_pid_path, NATS_SERVER_BINARY};
 
 use crate::appearance::spinner::Spinner;
 use crate::cfg::cfg_dir;
-use crate::up::{DOWNLOADS_DIR, WASMCLOUD_PID_FILE};
-use crate::util::get_wasmcloud_hosts;
+use crate::up::{
+    DEFAULT_LATTICE_PREFIX, DOWNLOADS_DIR, RELEASE_NODE, WASMCLOUD_CTL_CREDSFILE,
+    WASMCLOUD_CTL_HOST, WASMCLOUD_CTL_JWT, WASMCLOUD_CTL_PORT, WASMCLOUD_CTL_SEED,
+    WASMCLOUD_LATTICE_PREFIX, WASMCLOUD_PID_FILE,
+};
+use crate::util::{get_wasmcloud_hosts, nats_client_from_opts};
 
 #[derive(Parser, Debug, Clone)]
 pub(crate) struct DownCommand {
+    /// A lattice prefix is a unique identifier for a lattice, and is frequently used within NATS topics to isolate messages from different lattices
+    #[clap(
+        short = 'x',
+        long = "lattice-prefix",
+        default_value = DEFAULT_LATTICE_PREFIX,
+        env = WASMCLOUD_LATTICE_PREFIX,
+    )]
+    pub(crate) lattice_prefix: String,
+
+    /// An IP address or DNS name to use to connect to NATS for Control Interface (CTL) messages, defaults to the value supplied to --nats-host if not supplied
+    #[clap(long = "ctl-host", env = WASMCLOUD_CTL_HOST)]
+    pub(crate) ctl_host: Option<String>,
+
+    /// A port to use to connect to NATS for CTL messages, defaults to the value supplied to --nats-port if not supplied
+    #[clap(long = "ctl-port", env = WASMCLOUD_CTL_PORT)]
+    pub(crate) ctl_port: Option<u16>,
+
+    /// Convenience flag for CTL authentication, internally this parses the JWT and seed from the credsfile
+    #[clap(long = "ctl-credsfile", env = WASMCLOUD_CTL_CREDSFILE)]
+    pub(crate) ctl_credsfile: Option<PathBuf>,
+
+    /// A seed nkey to use to authenticate to NATS for CTL messages
+    #[clap(long = "ctl-seed", env = WASMCLOUD_CTL_SEED, requires = "ctl_jwt")]
+    pub(crate) ctl_seed: Option<String>,
+
+    /// A user JWT to use to authenticate to NATS for CTL messages
+    #[clap(long = "ctl-jwt", env = WASMCLOUD_CTL_JWT, requires = "ctl_seed")]
+    pub(crate) ctl_jwt: Option<String>,
+
     #[clap(long = "host-id")]
     pub host_id: Option<String>,
 
@@ -32,7 +66,7 @@ pub(crate) async fn handle_command(
 }
 
 pub(crate) async fn handle_down(
-    _cmd: DownCommand,
+    cmd: DownCommand,
     output_kind: OutputKind,
 ) -> Result<CommandOutput> {
     let install_dir = cfg_dir()?.join(DOWNLOADS_DIR);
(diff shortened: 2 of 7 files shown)


aish-where-ya:
To be discussed:
Consider a host is spun up in interactive mode and consecutive hosts are spun in interactive/detached mode.
When Ctrl+C is hit in the terminal that runs the first interactive host (that also spins NATS), NATS is killed as soon as Ctrl+C is hit. The expected behavior here is for the host to stop but NATS to keep running.

This case is encountered only in this specific case.

This results in zombie processes if the interactive host was the first one to be run and NATS was spawned along with this first host. The zombie processes are the other hosts that were connected to this NATS server. This is because NATS was the initial process and the host was spun using `group_spawn` on UNIX or `spawn` on Windows that returns a child process group or child process respectively. More inspection is needed on this and how do we want the expected/ideal behavior here to be.

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
+        let host = match hosts.iter().find(|h| h.id == host_id) {
+            Some(host) => host,
+            None => bail!("Host {} is not running", host_id),
+        };
+        sp.update_spinner_message(format!("Stopping host {}...", host.id));
+        if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
+            if output.stderr.is_empty() && output.stdout.is_empty() {
+                // if there was a host running, 'stop' has no output.
+                // Give it time to stop before stopping nats
+                tokio::time::sleep(Duration::from_secs(6)).await;
```
connorsmith256: We should find another way to do this. If a user has 3 hosts running, taking ~20s to stop all of them is really slow. Also, if 3 hosts are running, and the user just stops one of them, we don't need to wait at all since we're not going to stop NATS

Why was this necessary? We didn't have this sleep before, right? Was the `host_stopped` event not getting emitted before the NATS server shut down or something?

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
         }
+    } else if _cmd.all {
+        for (i, host) in hosts.iter().enumerate() {
+            sp.update_spinner_message(format!(
+                "Stopping host {} ({}/{})...",
+                host.id,
+                i + 1,
+                hosts.len()
+            ));
+            if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
```
connorsmith256: The following if/else is duplicated 3 times, so it feels like a good candidate for a separate function

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
+                    "❌ NATS server did not stop successfully: {e:?}\n"
+                ));
+            } else {
+                out_json.insert("nats_stopped".to_string(), json!(true));
+                out_text.push_str("✅ NATS server stopped successfully\n");
+            }
         }
+    } else {
+        out_json.insert("nats_stopped".to_string(), json!(false));
+        out_text.push_str("NATS server is still running\n");
```
connorsmith256, line 169: I don't think this is right. We shouldn't add `nats_stopped` to the JSON output, and I don't think we need to print "still running" either

src/down/mod.rs
```diff
@@ -81,13 +162,14 @@ pub(crate) async fn handle_down(
...
 /// Helper function to send wasmCloud the `stop` command and wait for it to clean up
-pub(crate) async fn stop_wasmcloud<P>(bin_path: P) -> Result<Output>
+pub(crate) async fn stop_wasmcloud<P>(bin_path: P, host_id: Option<String>) -> Result<Output>
 where
     P: AsRef<Path>,
 {
     Command::new(bin_path.as_ref())
         .stdout(Stdio::piped())
         .arg("stop")
+        .env("RELEASE_NODE", host_id.unwrap_or("".to_string()))
```
connorsmith256: Nit: `RELEASE_NODE` should probably be defined as a const. I noticed we have it in `src/up/config.rs`, so you can pull it in from there. We already depend on `up` from `down` for a couple other consts

src/up/config.rs
```diff
@@ -62,6 +66,7 @@ pub(crate) const DEFAULT_ALLOW_FILE_LOAD: &str = "true";
 pub(crate) async fn configure_host_env(
     nats_opts: NatsOpts,
     wasmcloud_opts: WasmcloudOpts,
+    other_host_config_values: HashMap<String, String>,
```
connorsmith256: Can we add these to `wasmcloud_opts` instead?

src/up/config.rs
```diff
@@ -74,9 +79,26 @@ pub(crate) async fn configure_host_env(
...
         host_config.insert(WASMCLOUD_HOST_SEED.to_string(), seed);
+    } else {
+        host_config.insert(
+            WASMCLOUD_HOST_SEED.to_string(),
+            KeyPair::new_server()
+                .seed()
+                .unwrap_or_else(|_| "None".to_string()),
+        );
     }
+    if let Some(value) = host_config.get("WASMCLOUD_HOST_SEED") {
```
connorsmith256: Nit: use the const here

Also, we should `bail` in an `else` block, since this value needs to exist

src/up/config.rs
```diff
@@ -74,9 +79,26 @@ pub(crate) async fn configure_host_env(
...
+        host_config.insert(
+            RELEASE_NODE.to_string(),
+            nkeys::KeyPair::from_seed(value).unwrap().public_key(),
+        );
```
connorsmith256: Suggestion: a comment here explaining that we're tagging the OTP process with the host's ID so we can find it later would be helpful

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+            spinner.update_spinner_message(" Downloading NATS ...".to_string());
+            let nats_binary = ensure_nats_server(&cmd.nats_opts.nats_version, &install_dir).await?;
+
+            spinner.update_spinner_message(" Starting NATS ...".to_string());
+            start_nats(&install_dir, &nats_binary, cmd.nats_opts.clone()).await?;
+            Some(nats_binary)
+        } else {
+            // If we can connect to NATS, return None as we aren't managing the child process.
+            // Otherwise, exit with error since --nats-connect-only was specified
+            tokio::net::TcpStream::connect(&nats_listen_address)
```
connorsmith256: Do we need this anymore? We've already checked the TCP connection in `is_nats_running`, right?

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
-            })?
-    };
+        spinner.println("Detected a running NATS server");
+    }
 
     // Download wasmCloud if not already installed
     let wasmcloud_executable = if !cmd.wasmcloud_opts.start_only {
         spinner.update_spinner_message(" Downloading wasmCloud ...".to_string());
         ensure_wasmcloud(&cmd.wasmcloud_opts.wasmcloud_version, &install_dir).await?
     } else {
```
connorsmith256, line 336: I think this else block is a bug. `start_only` says
```
/// If enabled, wasmCloud will not be downloaded if it's not installed
```

So I don't understand why we're trying to clean up NATS and return an error whenever this flag is set

Should this else block exist at all?

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
-        if nats_bin.is_some() {
+        if nats_is_running {
             stop_nats(install_dir).await?;
         }
         return Err(anyhow!("wasmCloud was not installed, exiting without downloading as --wasmcloud-start-only was set"));
     };
 
+    // Find an open port for the host
+    let host_port = if let Some(port_number) = cmd.wasmcloud_opts.dashboard_port {
+        port_number
```
connorsmith256: What if the user supplied a port, but that port is already in use? I think we'll need to confirm the port is open

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+    } else {
+        match find_open_port().await {
+            Ok(port_number) => port_number,
+            Err(e) => {
+                // Find number of hosts running. If they are 0, stop nats
```
connorsmith256: Suggestion: we could get around this cleanup code if we check for an open port before we start the NATS server

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+                    }
+                }
+                return Err(anyhow!("Error finding an open port {:?}", e));
+            }
+        }
+    };
+
+    let mut host_config_values = HashMap::new();
+    host_config_values.insert(
+        "WASMCLOUD_DASHBOARD_PORT".to_string(),
```
connorsmith256: Nit: use the const

src/up/mod.rs
```diff
@@ -318,34 +349,32 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
         stderr,
-        host_env,
+        host_env.clone(),
     )
     .await
     {
         Ok(child) => child,
         Err(e) => {
             // Ensure we clean up the NATS server if we can't start wasmCloud
+            // TODO: check number of hosts before stopping nats
```
connorsmith256: Leftover TODO?

src/up/mod.rs
```diff
@@ -318,34 +349,32 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
             }
             return Err(e);
         }
     };
 
-    let url = format!("{}:{}", "127.0.0.1", dashboard_port);
+    let url = format!("{}:{}", LOCALHOST, host_port);
     if wait_for_server(&url, "Washboard").await.is_err() {
-        if nats_bin.is_some() {
+        if nats_is_running {
```
connorsmith256, line 391: What if NATS was started by another command? I think we'll want the same logic we had elsewhere: check if there are zero running hosts before cleaning up. Alternatively, we could only ever clean up the NATS server if it was started by the current command

src/up/mod.rs
```diff
@@ -361,13 +390,22 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
         );
 
         // Terminate wasmCloud and NATS processes
-        let output = stop_wasmcloud(wasmcloud_executable.clone()).await?;
+        let output = stop_wasmcloud(
+            wasmcloud_executable.clone(),
+            host_env.get("RELEASE_NODE").cloned(),
```
connorsmith256: Nit: use const

src/up/mod.rs
```diff
@@ -361,13 +390,22 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+        )
+        .await?;
         if !output.status.success() {
             log::warn!("wasmCloud exited with a non-zero exit status, processes may need to be cleaned up manually")
         }
-        if !cmd.nats_opts.connect_only {
-            stop_nats(&install_dir).await?;
-        }
+        // The following code does nothing as NATS is killed as soon as Ctrl+C is hit.
```
connorsmith256: Let's just delete this commented code, since it's not needed for the reason you put in the comment

src/up/mod.rs
```diff
@@ -381,7 +419,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
     if cmd.detached {
         // Write the pid file with the selected version
         tokio::fs::write(install_dir.join(config::WASMCLOUD_PID_FILE), version).await?;
-        let url = format!("http://localhost:{}", dashboard_port);
+        let url = format!("{}:{}", "http://localhost", host_port);
```
connorsmith256: Nit: use const

src/up/mod.rs
```diff
@@ -403,6 +441,50 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
     Ok(CommandOutput::new(out_text, out_json))
 }
 
+pub(crate) async fn is_nats_running(
```
connorsmith256: I think this should be moved to `start/nats.rs`

src/util.rs
```diff
@@ -1,8 +1,44 @@
...
 use wash_lib::config::DEFAULT_NATS_TIMEOUT_MS;
+use wash_lib::start::is_wasmcloud_running;
+use wasmcloud_control_interface::{ClientBuilder, Host};
+
+// filter based on pid
+pub(crate) async fn get_wasmcloud_hosts<P>(bin_path: P) -> Result<Vec<Host>>
+where
+    P: AsRef<Path>,
+{
+    let dummy_nats_client = async_nats::ConnectOptions::default()
```
connorsmith256: Why `dummy`?

src/util.rs
```diff
@@ -1,8 +1,44 @@
...
+use crate::up::{DEFAULT_NATS_HOST, DEFAULT_NATS_PORT};
 use anyhow::{anyhow, bail, Context, Result};
+use std::path::Path;
 use term_table::{Table, TableStyle};
 use wash_lib::config::DEFAULT_NATS_TIMEOUT_MS;
+use wash_lib::start::is_wasmcloud_running;
+use wasmcloud_control_interface::{ClientBuilder, Host};
+
+// filter based on pid
+pub(crate) async fn get_wasmcloud_hosts<P>(bin_path: P) -> Result<Vec<Host>>
```
connorsmith256: I think this will need to take some args, like the NATS host, NATS port, and JS domain. Those are all in `WasmcloudOpts` so maybe you can just pass that

src/util.rs
```diff
@@ -1,8 +1,44 @@
...
+where
+    P: AsRef<Path>,
+{
+    let dummy_nats_client = async_nats::ConnectOptions::default()
+        .connect(format!("{DEFAULT_NATS_HOST}:{DEFAULT_NATS_PORT}"))
+        .await?;
+
+    let control_client = ClientBuilder::new(dummy_nats_client)
+        // .js_domain(WASMCLOUD_JS_DOMAIN.to_string()) // accept this as a var from wash up or env vars
```
connorsmith256: This will need to be done

src/util.rs
```diff
@@ -1,8 +1,44 @@
...
+                "Failed to construct local control interface client: {err:?}"
+            ))
+        })?;
+    let hosts = control_client.get_hosts().await.map_err(|err| {
+        anyhow!(format!(
+            "Was able to connect to NATS, but failed to get hosts: {err:?}"
+        ))
+    })?;
+    let mut running_hosts = Vec::new();
+    for host in hosts {
```
connorsmith256, line 32: Nit: instead of a mutable for loop, this could be done with `hosts.iter().filter(|host| is_wasmcloud_running...).collect()`

src/util.rs
```diff
@@ -1,8 +1,44 @@
...
+where
+    P: AsRef<Path>,
+{
+    let dummy_nats_client = async_nats::ConnectOptions::default()
+        .connect(format!("{DEFAULT_NATS_HOST}:{DEFAULT_NATS_PORT}"))
+        .await?;
+
+    let control_client = ClientBuilder::new(dummy_nats_client)
+        // .js_domain(WASMCLOUD_JS_DOMAIN.to_string()) // accept this as a var from wash up or env vars
+        .build()
```
connorsmith256, line 19: Optimization: since we're only interested in local hosts, we could potentially set the `auction_timeout` to be lower than the default of 2s. Something in the ballpark of ~500ms might work, possibly even lower (needs testing). That would make `wash down`/ctrl-C go faster

connorsmith256 (changes requested):
So excited for this. Mostly nits, and a few questions

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -231,10 +251,10 @@ where
...
 {
     // If we can connect to the local port, a wasmCloud host won't be able to listen on that port
-    let port = env_vars
-        .get("WASMCLOUD_DASHBOARD_PORT")
-        .cloned()
-        .unwrap_or_else(|| "4000".to_string());
+    let port = match env_vars.get("WASMCLOUD_DASHBOARD_PORT").cloned() {
+        Some(port) => port,
+        None => return Err(anyhow!("wasmCloud host port is not set")),
+    };
```
brooksmtownsend: If this environment variable isn't set, then the host will try to start on port `4000`. I think we should still default to `4000` here

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
-    if host_bin.is_file() {
-        sp.update_spinner_message(" Stopping host ...".to_string());
-        if let Ok(output) = stop_wasmcloud(host_bin).await {
+
+    let mut stopped_hosts = 0;
+    let hosts = get_wasmcloud_hosts(host_bin.clone()).await?;
+
+    if hosts.is_empty() {
+        sp.println("No local wasmCloud hosts detected.");
+    } else if let Some(host_id) = _cmd.host_id {
```
brooksmtownsend: Now that you're using the cmd, I would suggest removing the underscore and just using it as `cmd`

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
+        let host = match hosts.iter().find(|h| h.id == host_id) {
+            Some(host) => host,
+            None => bail!("Host {} is not running", host_id),
+        };
+        sp.update_spinner_message(format!("Stopping host {}...", host.id));
```
brooksmtownsend: nit: space between `{}` and `...`

src/down/mod.rs
brooksmtownsend: Since this is used multiple times in these if/else branches, would suggest moving the sleep to right before stopping the NATS server

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
+                out_json.insert("host_stopped".to_string(), json!(host.id));
+                out_text.push_str(&format!(
+                    "🤔 Host {} did not appear to be running, assuming it's already stopped\n",
+                    host_id
+                ));
+            }
+        };
+    } else if hosts.len() == 1 {
+        let host = hosts.first().unwrap();
+        sp.update_spinner_message(format!("Stopping host {}...", host.id));
```
brooksmtownsend: nit: space between {} and ...

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
+                host.id,
+                i + 1,
+                hosts.len()
+            ));
+            if let Ok(output) = stop_wasmcloud(&host_bin, Some(host.id.to_string())).await {
+                if output.stderr.is_empty() && output.stdout.is_empty() {
+                    // if there was a host running, 'stop' has no output.
+                    // Give it time to stop before stopping nats
+                    tokio::time::sleep(Duration::from_secs(6)).await;
+                    stopped_hosts += 1;
```
brooksmtownsend: Would suggest moving this sleep to right before stopping NATS, we can stop multiple hosts quickly

src/down/mod.rs
```diff
@@ -41,36 +48,110 @@ pub(crate) async fn handle_down(
...
         .ok_or_else(|| anyhow::anyhow!("Couldn't find path to wasmCloud binary. Is it running?"))?;
-    if host_bin.is_file() {
-        sp.update_spinner_message(" Stopping host ...".to_string());
-        if let Ok(output) = stop_wasmcloud(host_bin).await {
+
+    let mut stopped_hosts = 0;
+    let hosts = get_wasmcloud_hosts(host_bin.clone()).await?;
+
+    if hosts.is_empty() {
+        sp.println("No local wasmCloud hosts detected.");
```
brooksmtownsend: I think you can `bail` here

src/down/mod.rs
brooksmtownsend: If you don't supply the host_id here, and we try to stop a local host, does it work? Wondering if it's better to only supply the RELEASE_NODE environment variable if the host_id is provided

src/up/config.rs
```diff
@@ -74,9 +79,26 @@ pub(crate) async fn configure_host_env(
...
+    }
     if let Some(seed) = wasmcloud_opts.host_seed {
         host_config.insert(WASMCLOUD_HOST_SEED.to_string(), seed);
+    } else {
+        host_config.insert(
+            WASMCLOUD_HOST_SEED.to_string(),
+            KeyPair::new_server()
+                .seed()
+                .unwrap_or_else(|_| "None".to_string()),
+        );
```
brooksmtownsend: If for some reason we cannot create a seed here, we shouldn't insert `None` as I think that will break the host. Double checking

brooksmtownsend:
```
14:07:37.256 [error] Failed to obtain host public key from seed: (None). Generating a new host key instead.
```
It won't break, but it will generate its own

src/up/config.rs
```diff
@@ -74,9 +79,26 @@ pub(crate) async fn configure_host_env(
...
     }
+    if let Some(value) = host_config.get("WASMCLOUD_HOST_SEED") {
+        host_config.insert(
+            RELEASE_NODE.to_string(),
+            nkeys::KeyPair::from_seed(value).unwrap().public_key(),
+        );
+    }
```
brooksmtownsend: It might be better to move this block up when you're inserting `WASMCLOUD_HOST_SEED`, since you already have the seed value available if provided by `wasmcloud_opts` or if you generated it on your own

src/up/config.rs
```diff
@@ -74,9 +79,26 @@ pub(crate) async fn configure_host_env(
     }
 
     // Host / Cluster configuration
+    if let Some(port) = other_host_config_values.get(WASMCLOUD_DASHBOARD_PORT) {
+        host_config.insert(WASMCLOUD_DASHBOARD_PORT.to_string(), port.to_string());
+    }
```
brooksmtownsend: I think this should be part of `wasmcloud_opts`, and then you can remove the other host config values

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+    }
 
     // Download wasmCloud if not already installed
     let wasmcloud_executable = if !cmd.wasmcloud_opts.start_only {
         spinner.update_spinner_message(" Downloading wasmCloud ...".to_string());
         ensure_wasmcloud(&cmd.wasmcloud_opts.wasmcloud_version, &install_dir).await?
     } else {
         // Ensure we clean up the NATS server if we can't start wasmCloud
-        if nats_bin.is_some() {
+        if nats_is_running {
```
brooksmtownsend: We want to make sure we close the NATS server if we launched it as a part of this `wash up` and wasmCloud was set to start only

src/up/mod.rs
```diff
@@ -361,13 +390,22 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
-            stop_nats(&install_dir).await?;
-        }
+        // The following code does nothing as NATS is killed as soon as Ctrl+C is hit.
+        // This results in zombie processes if the non-detatched host was the first one to be run and NATS was spawned along with this first host.
+        // The zombie processes are the other hosts that were connected to this NATS server. Ref spawn vs group-spawn for these processes.
+
+        // let hosts = get_wasmcloud_hosts(wasmcloud_executable.clone()).await?;
+        // if !cmd.nats_opts.connect_only && hosts.is_empty() {
+        //     stop_nats(&install_dir).await?;
+        // }
```
brooksmtownsend: Hm, so in this case we don't have any control over NATS in the foreground?

I think that's ok, using `wash up` in the foreground to launch NATS isn't the preferred method when you're running multiple hosts, and users should use the detached mode if they want it to be long running. I think you can remove this comment + code, unless you wanted to leave a note that the CTRL+C stopped the NATS server

src/up/mod.rs
brooksmtownsend: nit: This should use the `LOCALHOST` variable or just have `http://localhost` in the format string

src/up/mod.rs
brooksmtownsend: A doc comment would be helpful here to detail the different return value meanings

src/up/mod.rs
```diff
@@ -403,6 +441,50 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+    }
+}
+
+pub(crate) async fn find_open_port() -> Result<u16> {
+    for i in HOST_PORT_RANGE_START..=HOST_PORT_RANGE_END {
+        if let Ok(conn) = tokio::net::TcpStream::connect((LOCALHOST, i)).await {
+            drop(conn);
+        } else {
+            return Ok(i);
+        }
```
brooksmtownsend: nit: could this be simplified with a `.is_err()` check to make it just an if statement?

src/util.rs
brooksmtownsend: nit: this can just be `nats_client`

src/util.rs
brooksmtownsend: I don't think the JS domain is needed as long as we're just making the ctl request, so this can be left off

src/util.rs
brooksmtownsend: this function will likely have to take some control interface connection variables to ensure you're fetching hosts for the proper Nats server URL, lattice prefix, etc. They should be available on the UpCommand's connectionopts struct, and there should be a util function for creating a NATS client from that struct

src/up/mod.rs
```diff
@@ -269,46 +270,76 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+                return Err(anyhow!("Error finding an open port {:?}", e));
+            }
+        }
+    };
+
+    let mut host_config_values = HashMap::new();
+    host_config_values.insert(
+        "WASMCLOUD_DASHBOARD_PORT".to_string(),
+        host_port.to_string(),
+    );
```
brooksmtownsend: If you move this section inside of the function that configures the host environment, I think you can avoid having an extra hashmap for the one value

brooksmtownsend (changes requested):
This is an awesome change and I love how simple the RELEASE_NODE can be configured by using the host public key. Mostly nits and refactoring opportunities

src/down/mod.rs
brooksmtownsend: I like having some indication of what `wash down` took care of, perhaps we can adjust the wording? I'm trying to think of the JSON output as something we could run operationally, then we could assert that after the down that NATS is no longer running

src/up/config.rs
brooksmtownsend: resolving this as Connor touched on it above

src/up/config.rs
brooksmtownsend: Left a comment a little further down, but I think if we move the logic in this `if` branch up to the above if/else, then we'll always know the host seed

src/up/mod.rs
brooksmtownsend: So, if we _only_ want to start wasmCloud and not download it, we start NATS and then try to start wasmCloud. If for some reason wasmCloud isn't installed, then we would just have a dangling NATS server if we didn't clean it up here

src/up/mod.rs
brooksmtownsend: resolving as Connor said it

src/util.rs
brooksmtownsend: @connorsmith256 resolving your comment, I think with taking the `ConnectionOpts` the JS domain will come with it but I don't think it's necessary

src/util.rs
brooksmtownsend: This might be hard, because of the `.await`'s unless we collect into a bunch of futures

brooksmtownsend:
Regarding this
```
Consider a host is spun up in interactive mode and consecutive hosts are spun in interactive/detached mode.
When Ctrl+C is hit in the terminal that runs the first interactive host (that also spins NATS), NATS is killed as soon as Ctrl+C is hit. The expected behavior here is for the host to stop but NATS to keep running.
```

I think this is actually okay and not much we can do to get around it, the user specifies to run NATS and wasmCloud in the foreground and then they kill it with `CTRL+C`. So I don't think we need to design around this edge case 😄

src/up/mod.rs
connorsmith256: Right, we should clean up the NATS server if we fail to start wasmcloud. But that's not what's happening here. This is just saying "if start_only, stop nats", which I think is a bug

Give `wash up --wasmcloud-start-only` a try. For me it doesn't try to download wasmcloud, it just immediately errors

See https://github.com/wasmCloud/wash/issues/492, which I think is this bug

(Also, since the NATS process is a child of the wash command and not in its own process group, the cleanup happens automatically)

src/down/mod.rs
connorsmith256: I _think_ this is okay, since _usually_ an empty env var is the same as an env var which isn't present

I just double-checked the `bin/wasmcloud_host` script and it has:
```bash
RELEASE_NAME="${RELEASE_NAME:-"wasmcloud_host"}"
...
RELEASE_NODE="${RELEASE_NODE:-"$RELEASE_NAME"}"
```

So an empty string should be fine here

src/down/mod.rs
aish-where-ya: This sleep was a part of the original code but yes, if this has something to do with the `host_stopped` event being delayed, it can be moved to the part before stopping NATS.

src/up/mod.rs
aish-where-ya: You're right. After discussing with @brooksmtownsend , this can be resolved as a separate PR for #492.

src/down/mod.rs
brooksmtownsend: Talked with @connorsmith256 and while this is a little verbose, I think we'll take a little more information here just in case we want to assert `nats_stopped == true` in a future script without worrying about the absence of the field

brooksmtownsend:
Unfortunately I have to close this PR since the number of conflicts is just a little too great to resolve accurately. Given the outstanding comments as well, I've picked the relevant pieces of this PR into https://github.com/wasmCloud/wash/pull/556/commits/873ddf8027e27af088083e8c7299083ea071fd03 where we'll pick up the work. Thank you @aish-where-ya for getting this ball rolling!

PR #556, opened 2023-05-15
feat(*) wash burrito support

### LINKED ISSUE #464: [FEATURE] Allow wash up to spin up multiple wasmCloud hosts in quick succession

### LINKED ISSUE #492: [BUG] Using --wasmcloud-start-only with wash up does not find locally installed versions

brooksmtownsend:
## Feature or Problem
This PR adds support for running https://github.com/wasmCloud/wasmcloud-otp/pull/616 artifacts (burrito releases) with `wash up`. This will be a significant breaking change and will bump the minimum wasmCloud version that `wash` can manage from `0.57.1` (when we stopped using distillery) to `0.63.0`.

PR action items:
- [x] Add support to launch burrito releases with `wash up`
- [x] Stop burrito-release launched hosts with `wash down`
- [x] Add `maintenance uninstall` to the `wash drain` command (I think this can be done in a separate PR since burritos auto-upgrade)
- [x] Test on multiple platforms (Works on aarch64 mac, aarch64 linux, and x86 Windows)
- [x] General code cleanup, TODOs left around the implementation 

## Related Issues
Fixes https://github.com/wasmCloud/wash/issues/464 after pulling in implementation from wasmCloud/wash#527 
Fixed wasmCloud/wash#492 

## Release Information
`wash` v0.18.0
`wash-lib` v0.9.0

~?? We may need to hold on merging this PR until we can release 0.18 as scheduled~
Changed to v0.18.0 as it makes sense to release this with wadm

## Consumer Impact
Consumers won't notice a difference with running the wasmCloud host, only that the downloaded assets are a single binary `wasmcloud_host` instead of a tarball-ed directory

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification
As of creation, I've run `cargo run -- up` and `cargo run -- up -d` which worked great. `down` now works by using NATS to tell hosts to stop instead of executing the `mix release`, and I've verified that a variety of scenarios work well.
1. Stopping one host when multiple are running do not stop NATS/wadm
2. Stopping the only host will stop all resources
3. Not specifying a host ID will exit early and display the running hosts to the user
4. Trying to stop a host when nothing is running will clean up NATS and wadm

### DIFF
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -11,6 +11,8 @@ use log::warn;
 use reqwest::StatusCode;
 use tokio::fs::{create_dir_all, metadata, File};
 use tokio::process::{Child, Command};
+use tokio_stream::StreamExt;
+use tokio_util::io::StreamReader;
 
 const WASMCLOUD_GITHUB_RELEASE_URL: &str =
     "https://github.com/wasmCloud/wasmcloud-otp/releases/download";
@@ -166,7 +168,11 @@ where
             download_response.status()
         ));
     }
-    let wasmcloud_host_burrito = download_response.bytes().await?.to_vec();
+
+    let burrito_bites_stream = download_response
+        .bytes_stream()
+        .map(|result| result.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err)));
+    let mut wasmcloud_host_burrito = StreamReader::new(burrito_bites_stream);
     let version_dir = dir.as_ref().join(version);
     let file_path = version_dir.join(WASMCLOUD_HOST_BIN);
     if let Some(parent_folder) = file_path.parent() {
@@ -185,7 +191,7 @@ where
                 wasmcloud_file.set_permissions(perms).await?;
             }
         }
-        tokio::io::copy(&mut wasmcloud_host_burrito.as_slice(), &mut wasmcloud_file).await?;
+        tokio::io::copy(&mut wasmcloud_host_burrito, &mut wasmcloud_file).await?;
     }
 
     // Return success if wasmCloud components exist, error otherwise
@@ -392,7 +398,7 @@ mod test {
     }
 
     const NATS_SERVER_VERSION: &str = "v2.8.4";
-    const WASMCLOUD_HOST_VERSION: &str = "v0.63.0";
+    const WASMCLOUD_HOST_VERSION: &str = "v0.63.1";
 
     #[tokio::test]
     async fn can_download_and_start_wasmcloud() -> anyhow::Result<()> {
--- a/crates/wash-lib/src/start/wasmcloud.rs
+++ b/crates/wash-lib/src/start/wasmcloud.rs
@@ -9,15 +9,16 @@ use anyhow::{anyhow, Result};
 use command_group::AsyncCommandGroup;
 use futures::future::join_all;
 use log::warn;
+use reqwest::StatusCode;
 use tokio::fs::{create_dir_all, metadata, File};
 use tokio::process::{Child, Command};
 
 const WASMCLOUD_GITHUB_RELEASE_URL: &str =
     "https://github.com/wasmCloud/wasmcloud-otp/releases/download";
 #[cfg(target_family = "unix")]
-pub const WASMCLOUD_HOST_BIN: &str = "bin/wasmcloud_host";
+pub const WASMCLOUD_HOST_BIN: &str = "wasmcloud_host";
 #[cfg(target_family = "windows")]
-pub const WASMCLOUD_HOST_BIN: &str = "bin\\wasmcloud_host.bat";
+pub const WASMCLOUD_HOST_BIN: &str = "wasmcloud_host.exe";
 
 // Any version of wasmCloud under 0.63.0 uses Elixir releases and is incompatible
 // See https://github.com/wasmCloud/wasmcloud-otp/pull/616 for the move to burrito releases
@@ -39,7 +40,7 @@ const DEFAULT_DASHBOARD_PORT: u16 = 4000;
 /// use wash_lib::start::ensure_wasmcloud;
 /// let res = ensure_wasmcloud("v0.63.0", "/tmp/wasmcloud/").await;
 /// assert!(res.is_ok());
-/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/wasmcloud_host".to_string());
+/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/v0.63.0/wasmcloud_host".to_string());
 /// # }
 /// ```
 pub async fn ensure_wasmcloud<P>(version: &str, dir: P) -> Result<PathBuf>
@@ -72,10 +73,9 @@ where
 /// use wash_lib::start::ensure_wasmcloud_for_os_arch_pair;
 /// let os = std::env::consts::OS;
 /// let arch = std::env::consts::ARCH;
-/// TODO: Check to see if this really always stores in a version folder
 /// let res = ensure_wasmcloud_for_os_arch_pair(os, arch, "v0.63.0", "/tmp/wasmcloud/").await;
 /// assert!(res.is_ok());
-/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/bin/wasmcloud_host".to_string());
+/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/v0.63.0/wasmcloud_host".to_string());
 /// # }
 /// ```
 pub async fn ensure_wasmcloud_for_os_arch_pair<P>(
@@ -112,7 +112,7 @@ where
 /// use wash_lib::start::download_wasmcloud;
 /// let res = download_wasmcloud("v0.57.1", "/tmp/wasmcloud/").await;
 /// assert!(res.is_ok());
-/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/bin/wasmcloud_host".to_string());
+/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/v0.63.0/wasmcloud_host".to_string());
 /// # }
 /// ```
 pub async fn download_wasmcloud<P>(version: &str, dir: P) -> Result<PathBuf>
@@ -133,7 +133,7 @@ where
 /// * `arch` - Specifies the architecture of the binary to download, e.g. `amd64`
 /// * `version` - Specifies the version of the binary to download in the form of `vX.Y.Z`
 /// * `dir` - Where to unpack the wasmCloud host contents into. This should be the root level
-///   directory where to store hosts. Each host will be stored in a directory maching its version TODO: verifty this
+///   directory where to store hosts. Each host will be stored in a directory maching its version
 /// # Examples
 ///
 /// ```no_run
@@ -144,7 +144,7 @@ where
 /// let arch = std::env::consts::ARCH;
 /// let res = download_wasmcloud_for_os_arch_pair(os, arch, "v0.63.0", "/tmp/wasmcloud/").await;
 /// assert!(res.is_ok());
-/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/wasmcloud_host".to_string());
+/// assert!(res.unwrap().to_string_lossy() == "/tmp/wasmcloud/v0.63.0/wasmcloud_host".to_string());
 /// # }
 /// ```
 pub async fn download_wasmcloud_for_os_arch_pair<P>(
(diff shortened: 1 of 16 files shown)


crates/wash-lib/tests/parser/files/withcargotoml/Cargo.toml
```diff
@@ -5,4 +5,7 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[lib]
+crate-type = ["cdylib", "rlib"]
```
brooksmtownsend, line 9: note: this change to the test artifact lets people run `cargo install --git <this repo>` again. `cargo` was picking this crate up as a binary you could install

src/up/mod.rs
```diff
@@ -392,17 +393,18 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
-            if !cmd.nats_opts.connect_only {
-                stop_nats(install_dir).await?;
-            }
             if let Some(mut child) = wadm_process {
                 child.kill().await?;
             }
+            //TODO: only kill NATS if there are no running hosts
+            if !cmd.nats_opts.connect_only {
+                stop_nats(install_dir).await?;
+            }
```
brooksmtownsend: don't forget this, whoever opened this PR

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -158,50 +157,36 @@ where
...
+    // to pipe the response body into a file. I'm not sure if there's a better way to do this.
+    let download_response = reqwest::get(url.clone()).await?;
+    if download_response.status() != StatusCode::OK {
+        return Err(anyhow!(
+            "Failed to download wasmCloud host from {}. Status code: {}",
+            url,
+            download_response.status()
+        ));
+    }
+    let wasmcloud_host_burrito = reqwest::get(url).await?.bytes().await?.to_vec();
```
brooksmtownsend: @thomastaylor312 Line 170 is the one that feels weird, in tandem with having to copy it with `.as_slice()`

connorsmith256: Is there a reason we're calling `reqwest::get(url)` twice?

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -245,32 +231,14 @@ where
...
-        // Stderr will include :nodedown if no other host is running, otherwise
-        // stdout will contain the PID
-        if !String::from_utf8_lossy(&output.stderr).contains(":nodedown") {
-            return Err(anyhow!(
-                "Another wasmCloud host is already running on this machine with PID {}",
-                String::from_utf8_lossy(&output.stdout)
-            ));
-        }
-    }
-
```
connorsmith256, line 264: What happens now if you call the burrito app when one is already running?

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -289,17 +257,8 @@ where
...
-    let lib_dir = versioned_dir.join("lib");
-    let releases_dir = versioned_dir.join("releases");
-    let file_checks = vec![
-        metadata(versioned_dir),
-        metadata(bin_dir),
-        metadata(bin_file.clone()),
-        metadata(lib_dir),
-        metadata(releases_dir),
-    ];
+    let file_checks = vec![metadata(bin_file.clone())];
```
connorsmith256: Under the assumption we're sticking to a single binary from here on out, you could drop the join_all iter stuff

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -309,11 +268,18 @@ where
...
 /// Helper function to determine the wasmCloud host release path given an os/arch and version
 fn wasmcloud_url(os: &str, arch: &str, version: &str) -> String {
-    format!("{WASMCLOUD_GITHUB_RELEASE_URL}/{version}/{arch}-{os}.tar.gz")
+    // NOTE(brooksmtownsend): I'm hardcoding `gnu` here because I'm not sure how to determine
+    // that programmatically. This essentially is what we had before (gnu only) but we do have a musl
+    // release that we should consider.
+    let os = os
+        .replace("macos", "darwin")
+        .replace("linux", "linux_gnu")
+        .replace("windows", "windows.exe");
```
connorsmith256, line 278: Ah yes, I love running windows.exe as my OS 😜

src/down/mod.rs
```diff
@@ -73,36 +131,69 @@ pub(crate) async fn handle_down(
...
+    } else if all {
+        let host_stops = hosts
+            .iter()
+            .map(|host| async {
+                let host_id = &host.id;
+                let _ = client
+                    .stop_host(host_id, None)
+                    .await
+                    // TODO: deal with this error
+                    .map_err(|e| anyhow!(e));
```
connorsmith256: Can we TODONE this and line 176?

src/up/mod.rs
```diff
@@ -281,13 +282,18 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
 
-    // Avoid downloading + starting NATS if the user already runs their own server. Ignore connect_only
-    // if this server has a remote and credsfile as we have to start a leafnode in that scenario
+    let nats_client = nats_client_from_wasmcloud_opts(&cmd.wasmcloud_opts).await;
     let nats_opts = cmd.nats_opts.clone();
-    let nats_bin = if !cmd.nats_opts.connect_only
+    // Avoid downloading + starting NATS if the user already runs their own server and we can connect.
+    // Ignore connect_only if this server has a remote and credsfile as we have to start a leafnode in that scenario
+    let nats_bin = if (!cmd.nats_opts.connect_only && nats_client.is_err())
         || cmd.nats_opts.nats_remote_url.is_some() && cmd.nats_opts.nats_credsfile.is_some()
```
connorsmith256: This condition is gnarly. The comment helps, but consider simplifying the expression so a comment isn't necessary

src/up/mod.rs
connorsmith256: 👀

src/up/mod.rs
```diff
@@ -364,7 +365,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
```
connorsmith256, line 408: I can't comment on line 358, and I know it's not technically related to the burrito stuff, but if you want to fix https://github.com/wasmCloud/wash/issues/492 while you're here.... it's been a thorn in our side for a while

src/up/mod.rs
```diff
@@ -560,6 +565,47 @@ async fn is_wadm_running(nats_opts: &NatsOpts, lattice_prefix: &str) -> Result<b
...
+/// # Arguments
+/// `supplied_port` - The port supplied by the user so we can check if it's open
+async fn ensure_open_port(supplied_port: Option<u16>) -> Result<u16> {
+    if let Some(port) = supplied_port {
+        tokio::net::TcpStream::connect((LOCALHOST, port))
+            .await
+            .map(|_tcp_stream| port)
+            .map_err(|e| anyhow!(e))
+    } else {
+        for i in 4000..=5000 {
```
connorsmith256: Nit, but if `DEFAULT_DASHBOARD_PORT` were to change, this would become a bug

connorsmith256 (changes requested):
Just nits and questions

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: Nope! Good catch

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: It launches another one, no conflicts on the burrito side (but it will crash if the port is allocated for the washboard)

src/down/mod.rs
brooksmtownsend: I think on line 176 it's handled since that one will bubble up, next commit here will have us logging for errors and not including the host IDs

src/up/mod.rs
brooksmtownsend: I believe in #527 we made the opinion here that if you use `wash up` to start a host + NATS, then start another host, and then kill the foregrounded wash up, then we shouldn't take the action to keep NATS around for the user

crates/wash-lib/src/start/wasmcloud.rs
connorsmith256: Hmm, can you share the error message? It might be okay. This might be an indirect way to make `wash up` idempotent now 😅 I just want to make sure it's not confusing to users

connorsmith256 (dismissed):
:shipit:

src/up/mod.rs
connorsmith256: Ah right. So should we remove the TODO?

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: Well... right now it actually "just works" and launches another host on port 4001

connorsmith256: Ah okay, this is captured separately so no worries: https://github.com/wasmCloud/wash/issues/602

connorsmith256 (dismissed):

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -1,47 +1,45 @@
...
+pub const WASMCLOUD_HOST_BIN: &str = "wasmcloud_host";
 #[cfg(target_family = "windows")]
-pub const WASMCLOUD_HOST_BIN: &str = "bin\\wasmcloud_host.bat";
+pub const WASMCLOUD_HOST_BIN: &str = "wasmcloud_host.exe";
 
-// Any version of wasmCloud under 0.57.0 uses distillery releases and is incompatible
-const MINIMUM_WASMCLOUD_VERSION: &str = "0.57.0";
+// Any version of wasmCloud under 0.63.0 uses Elixir releases and is incompatible
+// See https://github.com/wasmCloud/wasmcloud-otp/pull/616 for the move to burrito releases
+const MINIMUM_WASMCLOUD_VERSION: &str = "0.63.0";
```
thomastaylor312, line 26: Should this be 0.63.1?

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -158,50 +156,36 @@ where
...
+        if file_path.file_name().is_some() {
+            // Set permissions of executable files and binaries to allow executing
+            #[cfg(target_family = "unix")]
+            {
+                let mut perms = wasmcloud_file.metadata().await?.permissions();
+                perms.set_mode(0o755);
+                wasmcloud_file.set_permissions(perms).await?;
             }
         }
+        tokio::io::copy(&mut wasmcloud_host_burrito.as_slice(), &mut wasmcloud_file).await?;
```
thomastaylor312: You can actually use `bytes_stream` and consume the stream rather than all of this as_slice stuff if you want, but this does work.

If you want to use `bytes_stream` you'll want the [`StreamReader`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html) type from the tokio util crate

crates/wash-lib/src/start/wasmcloud.rs
```diff
@@ -345,7 +306,7 @@ mod test {
     use reqwest::StatusCode;
     use std::{collections::HashMap, env::temp_dir};
     use tokio::fs::{create_dir_all, remove_dir_all};
-    const WASMCLOUD_VERSION: &str = "v0.60.0";
+    const WASMCLOUD_VERSION: &str = "v0.63.0";
```
thomastaylor312, line 315: Don't forget to bump here too

src/up/config.rs
```diff
@@ -11,8 +11,9 @@ pub(crate) const DEFAULT_NATS_PORT: &str = "4222";
 // wadm configuration values
 pub(crate) const WADM_VERSION: &str = "v0.4.0";
 // wasmCloud configuration values, https://wasmcloud.dev/reference/host-runtime/host_configure/
-pub(crate) const WASMCLOUD_HOST_VERSION: &str = "v0.62.1";
+pub(crate) const WASMCLOUD_HOST_VERSION: &str = "v0.63.0";
```
thomastaylor312: Version here too

thomastaylor312 (changes requested):
One really minor change around the default wasmcloud version. The rest is all optional

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: It should be 0.63.0 for this PR, but I will bump the default to 0.63.1

brooksmtownsend: to clarify: the minimum _burrito_ release is 0.63.0, but the patch that includes something nice for wadm is 0.63.1 so it makes more sense to use that

src/up/config.rs
brooksmtownsend: bumping here

crates/wash-lib/src/start/wasmcloud.rs
brooksmtownsend: here is ok I think because it's a test, right?

thomastaylor312 (approved):

PR #559, opened 2023-05-17
Updates to dev container

lee-orr:
## Feature or Problem

This updates the dev container so it has access to all the required development tools (nextest, tinygo) as well as cargo watch, as well as resolving a permission issue with cargo/rust analyzer when using Microsoft's default dev container image as a base.

## Consumer Impact

Should have no consumer impact at all.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

I confirmed that I can load the dev container use it to build and run tests for wash.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,6 +9,7 @@
 		"ghcr.io/lee-orr/rusty-dev-containers/cargo-binstall:latest": {},
 		"ghcr.io/lee-orr/rusty-dev-containers/cargo-watch:latest": {},
 		"ghcr.io/lee-orr/rusty-dev-containers/cargo-nextest:latest": {},
+		"ghcr.io/devcontainers/features/go:1": {},
 		"ghcr.io/rchaganti/vsc-devcontainer-features/tinygo:1": {}
 	},
 	"workspaceMount": "source=${localWorkspaceFolder},target=/wash,type=bind,consistency=cached",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,6 +1,6 @@
 {
 	"name": "wash",
-	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
+	"image": "debian:bullseye",
 	"features": {
 		"ghcr.io/devcontainers/features/rust:1": {
 			"version": "latest"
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,11 +1,17 @@
 {
 	"name": "wash",
-	"image": "mcr.microsoft.com/devcontainers/base:bullseye",
+	"image": "debian:bullseye",
 	"features": {
```
brooksmtownsend: @lee-orr I was just considering this earlier today, what do you think about using the Rust base image so that we can skip that step when starting it up?

brooksmtownsend (changes requested):
I had a quick question around the base image, but otherwise this looks fantastic. @lee-orr make sure to sign your commit so we can keep the DCO check passing, you can find details on how to do so if you click **Details** next to the failing action

.devcontainer/devcontainer.json
lee-orr: Let me give it a test - I seem to remember having some issue with it when I tried in the rusty_dev_containers repo, but I may be remembering incorrectly

brooksmtownsend: Okay cool, if it doesn't work then that's all good, I just figured we could save ourselves a few seconds of building 😄

lee-orr: Do you know if this will work with regular go (i.e. no tiny go)? Looks like the lack of a sudo in rust: bullseye is causing issues with the existing tiny go devcontainer feature

brooksmtownsend: Bleh. Unfortunately at this point Go isn't quite at the point where it can work for our actors, though that's coming in the not-too-far future. I think it's probably better to leave it as bullseye then

brooksmtownsend: If you want to back that change out I can approve 👍🏻

lee-orr: I'm just adding tiny go to the now not aptly named rusty_dev_containers repo, and once that builds I'll confirm this all works

lee-orr: Looks like it works!

brooksmtownsend (approved):

Issue #538, opened 2023-05-01
[RFC] `wash` command structure rework


brooksmtownsend:
As `wash` matures and gathers more features and functionality, the list of subcommands has gotten larger. We've also been using the tool for a while now, and we've got a good idea for what feels good about `wash` and what doesn't. This issue also serves as an RFC, and is primarily a proposal for a "redesigned" `wash` experience. There will be breaking changes here, but I think that these changes will be well worth it and serve to create a better experience for wash users in general.

At a high level, here are a few things that I feel create friction with the current CLI:
1. The most common commands are not at top-level subcommands
    - When I think of my _most commonly used_ commands, they are: build, ctl, new, inspect, up, and soon to be `app`. The fact that the `ctl` commands are often 3-4 levels deep nested is a shame since they are used so often
3. Some uncommon commands are top-level subcommands
    - completions, ctx, gen, lint, validate are all things that aren't commonly used except for one-time use or advanced cases
5. Verbage is inconsistent across the CLI, especially with far nested commands
    - `wash ctl link query` vs `wash ctl get claims` vs `wash ctl start actor`, these mix verb/noun order, subcommands to get information, etc. Confusing.

I think the best way to propose this general change in structure is to show you what I would love for the help text to look like, all comments welcome. 
```
~/github.com/ ➜ wash -h
_________________________________________________________________________________
                               _____ _                 _    _____ _          _ _
                              / ____| |               | |  / ____| |        | | |
 __      ____ _ ___ _ __ ___ | |    | | ___  _   _  __| | | (___ | |__   ___| | |
 \ \ /\ / / _` / __| '_ ` _ \| |    | |/ _ \| | | |/ _` |  \___ \| '_ \ / _ \ | |
  \ V  V / (_| \__ \ | | | | | |____| | (_) | |_| | (_| |  ____) | | | |  __/ | |
   \_/\_/ \__,_|___/_| |_| |_|\_____|_|\___/ \__,_|\__,_| |_____/|_| |_|\___|_|_|
_________________________________________________________________________________

Interact and manage wasmCloud applications, projects, and runtime environments


Usage: wash [OPTIONS] <COMMAND>

Applications:
  app          Manage declarative applications and deployments (wadm) (experimental)
  call         Invoke a wasmCloud actor
  link         Link an actor and a provider component with configuration
  start        Start an actor or a provider component on a running wasmCloud host
  stop         Stop an actor component, provider component, or host
  get          Retrieve information about running components or the lattice

[...]

PR #569, opened 2023-05-22
feat(ux): flatten `wash ctl link` into `wash link`

vados-cosmonic:
## Feature or Problem

Flatten `wash ctl link` into `wash link`. 

Note that *both* ways of accessing this command will exist, until we are ready to go *only* to `wash link` (and update relevant documentation) and the other flattened commands.

See https://github.com/wasmCloud/wash/issues/538

## Release Information

`next`

## Consumer Impact

Customers will be able to use a more streamlined/convenient command structure.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration

An integration test for `wash link` was added which runs `wash link query`

### DIFF
--- a/crates/wash-lib/src/cli/link.rs
+++ b/crates/wash-lib/src/cli/link.rs
@@ -1,0 +1,191 @@
+use anyhow::{anyhow, Context, Result};
+use clap::Parser;
+use wasmcloud_control_interface::{CtlOperationAck, LinkDefinitionList};
+
+use crate::{
+    cli::{labels_vec_to_hashmap, CliConnectionOpts},
+    config::WashConnectionOptions,
+    id::{ModuleId, ServiceId},
+};
+
+#[derive(Parser, Debug, Clone)]
+pub struct LinkDelCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// Public key ID of actor
+    #[clap(name = "actor-id", value_parser)]
+    pub actor_id: ModuleId,
+
+    /// Capability contract ID between actor and provider
+    #[clap(name = "contract-id")]
+    pub contract_id: String,
+
+    /// Link name, defaults to "default"
+    #[clap(short = 'l', long = "link-name")]
+    pub link_name: Option<String>,
+}
+
+#[derive(Parser, Debug, Clone)]
+#[clap(
+    override_usage = "wash ctl link put --link-name <LINK_NAME> [OPTIONS] <actor-id> <provider-id> <contract-id> [values]..."
+)]
+pub struct LinkPutCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// Public key ID of actor
+    #[clap(name = "actor-id", value_parser)]
+    pub actor_id: ModuleId,
+
+    /// Public key ID of provider
+    #[clap(name = "provider-id", value_parser)]
+    pub provider_id: ServiceId,
+
+    /// Capability contract ID between actor and provider
+    #[clap(name = "contract-id")]
+    pub contract_id: String,
+
+    /// Link name, defaults to "default"
+    #[clap(short = 'l', long = "link-name")]
+    pub link_name: Option<String>,
+
+    /// Environment values to provide alongside link
+    #[clap(name = "values")]
+    pub values: Vec<String>,
+}
+
+#[derive(Parser, Debug, Clone)]
+pub struct LinkQueryCommand {
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+}
+
+#[derive(Debug, Clone, Parser)]
+pub enum LinkCommand {
+    /// Query established links
+    #[clap(name = "query")]
+    Query(LinkQueryCommand),
+
+    /// Establish a link definition
+    #[clap(name = "put")]
+    Put(LinkPutCommand),
+
+    /// Delete a link definition
+    #[clap(name = "del")]
+    Del(LinkDelCommand),
+}
+
+/// Converts error from Send + Sync error to standard anyhow error
+fn boxed_err_to_anyhow(e: Box<dyn ::std::error::Error + Send + Sync>) -> anyhow::Error {
+    anyhow!(e.to_string())
+}
+
+/// Query links for a given Wash instance
+///
+/// # Arguments
+///
+/// * `wco` - Options for connecting to wash
+///
+/// # Examples
+///
+/// ```no_run
+/// let ack = query_links(WashConnectionOptions::default()).await?;
+/// assert_eq!(ack.accepted, true);
+/// ```
+pub async fn query_links(wco: WashConnectionOptions) -> Result<LinkDefinitionList> {
+    wco.into_ctl_client(None)
+        .await?
+        .query_links()
+        .await
+        .map_err(boxed_err_to_anyhow)
+}
+
+/// Delete a single link
+///
+/// # Arguments
+///
+/// * `wco` - Options for connecting to wash
+/// * `actor_id` - The ID of the actor attached to the link
+/// * `contract_id` - The contract ID of the link
+/// * `link_name` - The link name of the link ('default')
+///
+/// # Examples
+///
+/// ```no_run
+/// let ack = delete_link(
+///   WashConnectionOptions::default(),
+///   "wasmcloud:httpserver",
+///   "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5", // wasmcloud.azurecr.io/echo:0.3.8
+///   "default",
+/// ).await?;
+/// assert_eq!(ack.accepted, true);
+/// ```
+pub async fn delete_link(
+    wco: WashConnectionOptions,
+    contract_id: &str,
+    actor_id: &ModuleId,
+    link_name: &str,
+) -> Result<CtlOperationAck> {
+    wco.into_ctl_client(None)
+        .await?
+        .remove_link(&actor_id.to_string(), contract_id, link_name)
+        .await
+        .map_err(boxed_err_to_anyhow)
+        .with_context(|| {
+            format!(
+                "Failed to remove link between {} and {} with link name {}",
+                actor_id, contract_id, link_name
+            )
+        })
+}
+
+/// Create ("put") a new link
+///
+/// # Arguments
+///
+/// * `wco` - Options for connecting to wash
+/// * `contract_id` - The contract ID of the link
+/// * `actor_id` - The ID of the actor attached to the link
+/// * `provider_id` - The ID of the provider attached to the link
+/// * `link_name` - The link name of the link ('default')
+///
+/// # Examples
+///
+/// ```no_run
+/// let ack = delete_link(
+///   WashConnectionOptions::default(),
+///   "wasmcloud:httpserver",
+///   "MBCFOPM6JW2APJLXJD3Z5O4CN7CPYJ2B4FTKLJUR5YR5MITIU7HD3WD5", // wasmcloud.azurecr.io/echo:0.3.8
+///   "VAG3QITQQ2ODAOWB5TTQSDJ53XK3SHBEIFNK4AYJ5RKAX2UNSCAPHA5M", // wasmcloud.azurecr.io/httpserver:0.17.0
+///   "default",
+///   vec!["KEY", "value"],
+/// ).await?;
+/// assert_eq!(ack.accepted, true);
+/// ```
+pub async fn create_link(
+    wco: WashConnectionOptions,
+    contract_id: &str,
+    actor_id: &ModuleId,
+    provider_id: &ServiceId,
+    link_name: &str,
+    link_values: &Vec<String>,
+) -> Result<CtlOperationAck> {
+    wco.into_ctl_client(None)
+        .await?
+        .advertise_link(
+            &actor_id.to_string(),
+            &provider_id.to_string(),
+            contract_id,
+            link_name,
+            labels_vec_to_hashmap(link_values.clone())?,
+        )
+        .await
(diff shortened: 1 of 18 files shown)


tests/common.rs
```diff
@@ -49,3 +56,122 @@ pub(crate) fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
...
+    test_dir: PathBuf,
+    nats_port: u16,
+    kill_cmd: String,
+    nats: Child,
+}
+
+impl TestWashInstance {
+    #[allow(unused)]
+    pub async fn create() -> Result<TestWashInstance> {
+        let epoch_secs = SystemTime::now()
```
thomastaylor312: It might be better to use a random string here instead. If multiple tests run at the same time, there is a chance they'll get the same number of seconds (since we aren't using millis here)

tests/common.rs
```diff
@@ -49,3 +56,122 @@ pub(crate) fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
...
+            .duration_since(UNIX_EPOCH)
+            .context("failed to retreive epoch")?
+            .as_secs();
+        let test_name = format!("test-{epoch_secs}");
+        let test_dir = test_dir_with_subfolder(test_name.as_str());
+        let log_path = test_dir.join("washup.log");
+        let stdout = std::fs::File::create(&log_path)
```
thomastaylor312: Nit (as this is a test): Should probably be using tokio file stuff rather than blocking io

tests/common.rs
```diff
@@ -49,3 +56,122 @@ pub(crate) fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
...
+        Ok(TestWashInstance {
+            test_dir,
+            kill_cmd: kill_cmd.to_string(),
+            nats,
+            nats_port,
+        })
+    }
+
+    #[allow(unused)]
+    pub async fn destroy(&mut self) -> Result<()> {
```
thomastaylor312: Let's make this an implementation of `Drop` instead of requiring the caller to manually destroy resources

thomastaylor312 (changes requested):
Logic looks good. Just a few fixes needed for tests

src/common/link_cmd.rs
```diff
@@ -0,0 +1,200 @@
...
+use crate::{
+    appearance::spinner::Spinner,
+    ctl::{ctl_client_from_opts, link_del_output, links_table, ConnectionOpts},
+    json,
+    util::{convert_error, validate_contract_id},
+    OutputKind,
+};
+
+#[derive(Parser, Debug, Clone)]
+pub(crate) struct LinkDelCommand {
```
thomastaylor312: Looking at this again, none of this has wash specific env vars, so this whole file should probably live in wash-lib! It will be easier to share/reuse from there

src/ctl/mod.rs
```diff
@@ -1135,7 +993,7 @@ async fn apply_manifest_providers(
     Ok(results)
 }
 
-async fn ctl_client_from_opts(
+pub(crate) async fn ctl_client_from_opts(
```
thomastaylor312: Nit: This function looks like it could be handy in wash-lib, but it isn't required to move for this PR

thomastaylor312 (changes requested):
Some additional comments I thought of around putting things in wash-lib

tests/common.rs
vados-cosmonic: Hey so the tests are marked `#[serial]` so this shouldn't happen, *but* I definitely agree with you on using the random name. I'll pull in `rand` and use that

- [x] use `rand` for random test name with the wash instance

tests/common.rs
vados-cosmonic: So this actually needed to be a regular file, for two reasons as far as I can see

- [`tokio::process::Command#stdout`](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.stdout) takes a regular `Stdio` which is `From<std::fs::File>`. I could do `into_std` here after making the file with the `tokio` version though.

- the `wash()` command is actually not async (most of the tests utilities aren't actually), so I'd need to actually change that to get properly async in the first place.

That said, I can change *both* of those bits to use the `tokio` variants -- been wanting to remove `wash()` anyway -- so I'll do that

- [x] use `tokio::fs::File` and remove use of `wash` (should be able to use cargo ENV var path)

tests/common.rs
vados-cosmonic: @thomastaylor312 So since there's an `tokio::process::Child` in the Struct, this means I have to do `async` `Drop` -- there are some ways to do it, but [the decent solutions](https://stackoverflow.com/questions/71541765/rust-async-drop) look like they involve a copy + spawn...

Do you have a good pattern that you usually use that I can look at?

[EDIT] - managed to avoid this by using `kill_on_drop` for the `Child`

src/common/link_cmd.rs
vados-cosmonic: Oh great, will move this there!

- [x] move `link_cmd` to `wash-lib`

src/ctl/mod.rs
vados-cosmonic: - [x] move `ctl_client_from_opts` into `wash-lib`

tests/common.rs
vados-cosmonic: OK, I reworked it a bit and just made sure that `kill_on_drop` was set for the command -- removed the use of `wash()`, but I *did not* update the rest of the code there to avoid this PR getting too big (they're all still `#[test]`s).

src/common/link_cmd.rs
vados-cosmonic: Hey @thomastaylor312 actually looking at this, this code has a dependency on the `wasmcloud_control_interface` crate -- that's not in `wash-lib` -- is there any reason behind that? Does `wash-lib` *intentionally* not use `wasmcloud_control_interface`?

vados-cosmonic: Just as an update here, `wasmcloud_control_interface` is OK to add as a dep!

crates/wash-lib/src/cli/link.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// * `wco` - Options for connecting to wash
+///
+/// # Examples
+///
+/// ```no_run
+/// let ack = query_links(WashConnectionOptions::default()).await?;
+/// assert_eq!(ack.accepted, true);
+/// ```
+pub async fn query_links(wco: WashConnectionOptions) -> Result<LinkDefinitionList> {
```
thomastaylor312, line 96: Note to ourselves for later: It kinda feels like these methods (like query_links and `delete_link`) could be on the impl of `WashConnectionOptions`, but since they are currently exclusively used for just CLI commands I think this current way of doing things is just fine

crates/wash-lib/src/config.rs
```diff
@@ -5,6 +5,12 @@ use std::{
     path::PathBuf,
 };
 
+use crate::context::WashContext;
```
thomastaylor312: Nit: Move crate imports to a section below external imports

crates/wash-lib/src/config.rs
```diff
@@ -42,3 +48,249 @@ pub fn model_cache_dir() -> IoResult<PathBuf> {
...
+        });
+
+        let ctl_port = self.ctl_port.unwrap_or_else(|| {
+            self.ctx
+                .as_ref()
+                .map(|c| c.ctl_port.to_string())
+                .unwrap_or_else(|| DEFAULT_NATS_PORT.to_string())
+        });
+
+        let ctl_jwt = if self.ctl_jwt.is_some() {
```
thomastaylor312: Nit(as I know this is code you didn't write): I think these similar blocks can all be simplified to something like this (not actually tested):

```rust
self.ctl_jwt.or_else(|| self.ctx.as_ref().map(|c| c.ctl_jwt.clone())).unwrap_or_default()
```

Generally it is more idiomatic to not use `is_some` unless it is a guard clause or you aren't using the data in the option

crates/wash-lib/src/config.rs
```diff
@@ -42,3 +48,249 @@ pub fn model_cache_dir() -> IoResult<PathBuf> {
...
+                .as_ref()
+                .map(|c| c.ctl_credsfile.clone())
+                .unwrap_or_default()
+        };
+
+        let nc =
+            create_nats_client_from_opts(&ctl_host, &ctl_port, ctl_jwt, ctl_seed, ctl_credsfile)
+                .await?;
+
+        let timeout = tokio::time::Duration::from_millis(self.timeout_ms);
```
thomastaylor312, line 223: This timeout is for control interface timeout and not nats timeout, do we actually need to return it? It looks like we ignore it everywhere in the rest of the code

src/app/mod.rs
```diff
@@ -196,7 +190,8 @@ async fn undeploy_model(cmd: UndeployCommand) -> Result<DeployModelResponse> {
 
 async fn deploy_model(cmd: DeployCommand) -> Result<DeployModelResponse> {
     let lattice_prefix = cmd.opts.lattice_prefix.clone();
-    let (client, _timeout) = nats_client_from_opts(cmd.opts).await?;
+    let wco: WashConnectionOptions = cmd.opts.try_into()?;
```
thomastaylor312: Nit: Might be better to just oneline this rather than having a var you immediately consume

src/common/link_cmd.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::{
+    appearance::spinner::Spinner,
+    ctl::{link_del_output, links_table, ConnectionOpts},
+    json,
+    util::validate_contract_id,
+    OutputKind,
+};
+
+#[derive(Parser, Debug, Clone)]
+pub(crate) struct LinkDelCommand {
```
thomastaylor312: I actually think these commands can be moved over to wash-lib as well. That is what I originally meant, but didn't convey very well. You can see another example of this here: https://github.com/wasmCloud/wash/blob/main/crates/wash-lib/src/cli/claims.rs

tests/common.rs
```diff
@@ -49,3 +57,138 @@ pub(crate) fn test_dir_file(subfolder: &str, file: &str) -> PathBuf {
...
+        Ok(TestWashInstance {
+            test_dir,
+            kill_cmd: kill_cmd.to_string(),
+            nats,
+            nats_port,
+        })
+    }
+
+    #[allow(unused)]
+    pub fn destroy(&mut self) -> Result<()> {
```
thomastaylor312: I think this can be removed now that drop is in place. Definitely better than allowing unused here

thomastaylor312 (changes requested):
This is almost there. Clarified one thing I missed before that needs to be address, but the rest of the things are nits you can address or ignore at your leisure

crates/wash-lib/src/config.rs
vados-cosmonic: Woops yeah, will move!

- [x] move crate import

crates/wash-lib/src/config.rs
vados-cosmonic: yeah I agree -- I didn't change the previous code but since I'm doing another pass anyway, will add this in -- was weird to me as well... Half the setup was one pattern and then the rest did this pattern

Note that it looks like they actually *want* the default of an option there -- the code *does want* the option (or `None` -- I think that's the default being used there)

- [x] refactor `.is_some` usage

crates/wash-lib/src/config.rs
vados-cosmonic: Yep saw that, I'm not sure why this code is like this to be honest -- it's like it returns a timeout for the caller to use if they want to, not sure why this is necessary. 

Open to removing it if you'd like

src/app/mod.rs
vados-cosmonic: Yeah that's what I did initially, it turns out to require a pretty ugly type hint:

```rust
    let (client, _timeout) = <ConnectionOpts as TryInto<WashConnectionOptions>>::try_into(cmd.opts)?.into_nats_client().await?;
```

src/common/link_cmd.rs
vados-cosmonic: Oh sure -- will move these too:

- [x] move commands into wash-lib

crates/wash-lib/src/cli/link.rs
vados-cosmonic: Yeah I'd think that should go on something like a `WashInstance`, which maybe has a constructor that takes a `WashConnectionOptions` -- agreed though, would be nice to have these on some consistent object (especially so we can re-use connections).

tests/common.rs
vados-cosmonic: Sure -- `drop` actually uses this (so `allow(unused)` could have been removed probably), but I just moved all this to `drop`

src/app/mod.rs
vados-cosmonic: I've gone ahead and changed all of them to this (though it `fmt`s to a more lines)

connorsmith256 (dismissed):
LGTM!

vados-cosmonic:
@connorsmith256 one more small refactor while I was looking -- removed the `link.rs` file completely!

crates/wash-lib/src/config.rs
thomastaylor312: Let's just remove it, but not gonna block on it here

src/app/mod.rs
thomastaylor312: You can revert it then, I thought it would be able to type infer here

thomastaylor312 (approved):

PR #580, opened 2023-05-30
feat(ux): flatten wash reg push/pull into wash push/pull

vados-cosmonic:
## Feature or Problem

Flatten `wash reg push` into `wash push`
Flatten `wash reg pull` into `wash pull`

See:
- https://github.com/wasmCloud/wash/issues/538

## Release Information

`next`

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration

Altered the existing integration tests to use the new subcommands

PR #585, opened 2023-05-31
feat(ux): flatten ctl link query claims inventory

vados-cosmonic:
### ~~:warning: This PR should be merged *after* https://github.com/wasmCloud/wash/pull/580~~

Merge `wash ctl link query`, `wash ctl get claims` and `wash ctl get inventory` into a single `wash get` command

See:
- https://github.com/wasmCloud/wash/issues/538

## Release Information

`next`

## Consumer Impact

Users should be able to more quickly and conveniently access information about the cluster in fewer keystrokes with `wash get` rather than `wash ctl link query` (and similar commands).

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration

Integration tests were added for the new top level commands (see `integration_get.rs`), and the existing tests continue to pass for older CLI commands.

### DIFF
--- a/crates/wash-lib/src/cli/output.rs
+++ b/crates/wash-lib/src/cli/output.rs
@@ -2,10 +2,11 @@ use std::collections::HashMap;
 use wasmbus_rpc::core::ActorLinks;
 
 use serde::Deserialize;
+use wasmcloud_control_interface::{GetClaimsResponse, Host, HostInventory};
 
 /// JSON Output of the `wash start` command
 #[derive(Debug, Deserialize)]
-pub struct StartCommandJsonOutput {
+pub struct StartCommandOutput {
     pub actor_id: Option<String>,
     pub actor_ref: Option<String>,
 
@@ -20,7 +21,7 @@ pub struct StartCommandJsonOutput {
 
 /// JSON Output representation of the `wash stop` command
 #[derive(Debug, Deserialize)]
-pub struct StopCommandJsonOutput {
+pub struct StopCommandOutput {
     pub host_id: Option<String>,
     pub result: String,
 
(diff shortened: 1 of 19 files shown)


crates/wash-lib/src/cli/output.rs
```diff
@@ -37,7 +38,28 @@ pub struct StopCommandJsonOutput {
 
 /// JSON output representation of the `wash link query` command
 #[derive(Debug, Deserialize)]
-pub struct LinkQueryOutput {
+pub struct LinkQueryJsonOutput {
```
connorsmith256: Supernit: I don't think the `Json` prefix is needed. While we only serialize to JSON today, I don't see anything here at the data representation layer that necessitates coupling to JSON. In a world where `wash get links -o toml` was a thing, these structs wouldn't need to change

connorsmith256 (dismissed):

crates/wash-lib/src/cli/output.rs
vados-cosmonic: Yeah that's true -- I'm open to changing them all to `...Output`

- [x] Rename `JsonOutput` to `Output`

vados-cosmonic:
Looks like there is a flaky test -- not any of the new added ones, but while I'm here I'd like to investigate it

connorsmith256 (dismissed):
LGTM. I can merge as-is or hold if you're still looking into the flaky test 👍

vados-cosmonic:
Yeah I'm still sus about [this one failure](https://github.com/wasmCloud/wash/actions/runs/5136995124/jobs/9244474304) -- I want to see if I can reproduce it in my own fork after running it a few times.

vados-cosmonic:
Yeah so a bunch of tests are [leaky](https://nexte.st/book/leaky-tests.html) (`nextest` is awesome!):

```
        LEAK [  26.120s] wash-cli::integration_get integration_get_claims
        LEAK [  16.044s] wash-cli::integration_get integration_get_host_inventory
        LEAK [  16.559s] wash-cli::integration_get integration_get_hosts
        LEAK [  13.734s] wash-cli::integration_get integration_get_links
        LEAK [  13.701s] wash-cli::integration_link integration_link
        LEAK [  17.404s] wash-cli::integration_start integration_start_actor
         LEAK [  19.604s] wash-cli::integration_start integration_start_provider
        LEAK [  11.569s] wash-cli::integration_stop integration_stop_host
        LEAK [  14.827s] wash-cli::integration_stop integration_stop_provider        
```

Something related to starting/stopping the processes. I think it might be `wash()` starting a `std::process::Command` instead of a tokio one... so that might be a somewhat deep change, because `wash()` is used in non-async contexts all over the place

vados-cosmonic:
Finally found the issue -- `nextest` does not seem to support interop with `serial_test`... I've been using `#[serial]`, and I'm fairly certain it does nothing right now when we run with `nextest`. The fix was to modify how we were using `nextest`'s serial functionality a little bit so we can more easily use it. No more leaky tests!

Seems to have exposed some issues with other tests so dealing with that...

brooksmtownsend (approved):
Nice!

PR #589, opened 2023-06-02
Update dep to support new provider archive

autodidaddict:
"Simple" version bump to take the newest version of provider-archive (which can serialize/deserialize provider schemas)


autodidaddict:
This turned out to be a larger can of worms than I expected. Closing this until we actually implement the full feature for wash being able to embed JSON schemas in `par` files.

PR #610, opened 2023-06-09
feat: add wash dev command

vados-cosmonic:
## Feature or Problem

Add `wash dev` command for local dev loops

## Release Information

`next`

## Consumer Impact

Users developing will be able to start one command to automatically work on actors

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Verified manually

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -1,0 +1,376 @@
+use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::{path::PathBuf, sync::Arc};
+
+use anyhow::{anyhow, bail, Context, Result};
+use clap::Parser;
+use console::style;
+use notify::{event::EventKind, Event as NotifyEvent, RecursiveMode, Watcher};
+use tokio::task::JoinHandle;
+use tokio::time::{timeout, Duration};
+use tokio::{select, sync::mpsc};
+use wash_lib::generate::emoji;
+use wash_lib::{
+    actor::{scale_actor, start_actor, StartActorArgs},
+    build::{build_project, SignConfig},
+    cli::dev::run_dev_loop,
+    cli::CommandOutput,
+    config::downloads_dir,
+    id::{ModuleId, ServerId},
+    parser::get_config,
+};
+use wasmcloud_control_interface::Host;
+
+use crate::{
+    down::{handle_down, DownCommand},
+    up::{handle_up, NatsOpts, UpCommand, WadmOpts, WasmcloudOpts, DOWNLOADS_DIR},
+};
+
+#[derive(Debug, Clone, Parser)]
+pub struct DevCommand {
+    #[clap(flatten)]
+    pub(crate) nats_opts: NatsOpts,
+
+    #[clap(flatten)]
+    pub(crate) wasmcloud_opts: WasmcloudOpts,
+
+    #[clap(flatten)]
+    pub(crate) wadm_opts: WadmOpts,
+
+    /// ID of the host to use for `wash dev`
+    /// if one is not selected, `wash dev` will attempt to use the single host in the lattice
+    #[clap(long = "host-id", name = "host-id", value_parser)]
+    pub host_id: Option<ServerId>,
+
+    /// Path to code directory
+    #[clap(name = "code-dir", long = "work-dir", env = "WASH_DEV_CODE_DIR")]
+    pub code_dir: Option<PathBuf>,
+
+    /// Whether to leave the host running after dev
+    #[clap(
+        name = "leave-host-running",
+        long = "leave-host-running",
+        env = "WASH_DEV_LEAVE_HOST_RUNNING",
+        default_value = "false",
+        help = "Leave the wasmCloud host running after stopping the devloop"
+    )]
+    pub leave_host_running: bool,
+
+    /// Run the host in a subprocess (rather than detached mode)
+    #[clap(
+        name = "use-host-subprocess",
+        long = "use-host-subprocess",
+        env = "WASH_DEV_USE_HOST_SUBPROCESS",
+        default_value = "false",
+        help = "Run the wasmCloud host in a subprocess (rather than detached mode)"
+    )]
+    pub use_host_subprocess: bool,
+}
+
+/// Utility struct for holding a wasmCloud host subprocess.
+/// This struct ensures that the join handle is aborted once the
+/// subprocess is dropped.
+struct HostSubprocess(Option<JoinHandle<()>>);
+
+impl HostSubprocess {
+    fn into_inner(mut self) -> Option<JoinHandle<()>> {
+        self.0.take()
+    }
+}
+
+impl Drop for HostSubprocess {
+    fn drop(&mut self) {
+        if let Some(handle) = self.0.take() {
+            handle.abort();
+        }
+    }
+}
+
+/// Handle `wash dev`
+pub async fn handle_command(
+    cmd: DevCommand,
+    output_kind: wash_lib::cli::OutputKind,
+) -> Result<CommandOutput> {
+    // Check if host is running
+    let pid_file = downloads_dir()?.join(DOWNLOADS_DIR);
+    let existing_instance = tokio::fs::metadata(pid_file).await.is_ok();
+
+    let mut host_subprocess: Option<HostSubprocess> = None;
+
+    // Start host if it's not already running
+    if !existing_instance {
+        eprintln!(
+            "{} {}{}",
+            emoji::WARN,
+            style("No running wasmcloud host detected (PID file missing), ").bold(),
+            style("starting a new host...").bold()
+        );
+        // Ensure that file loads are allowed
+        let mut wasmcloud_opts = cmd.wasmcloud_opts.clone();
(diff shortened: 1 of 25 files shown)


vados-cosmonic:
This still needs at least one E2E test, but it'll be a bit complicated to test, so putting this up while I work on that.

crates/wash-lib/src/config.rs
```diff
@@ -17,6 +17,7 @@ const WASH_DIR: &str = ".wash";
 const DOWNLOADS_DIR: &str = "downloads";
 pub const DEFAULT_NATS_HOST: &str = "127.0.0.1";
 pub const DEFAULT_NATS_PORT: &str = "4222";
+pub const DEFAULT_NATS_PORT_INT: u16 = 4222;
```
connorsmith256: I only see this used in one spot, where we convert it to a string. Is this needed? 🤔

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+    #[clap(name = "code-dir", long = "work-dir", env = "WASH_DEV_CODE_DIR")]
+    pub code_dir: Option<PathBuf>,
+}
+
+/// Handle `wash dev`
+pub async fn handle_command(
+    cmd: DevCommand,
+    output_kind: wash_lib::cli::OutputKind,
+) -> Result<CommandOutput> {
+    // Check if wash is running
```
connorsmith256: Nit: these should say "host" rather than "wash"

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+    let pid_file = downloads_dir()?.join(DOWNLOADS_DIR);
+    let wash_instance_running = tokio::fs::metadata(pid_file).await.is_ok();
+
+    // Start wash if it's not already running
+    if !wash_instance_running {
+        log::info!(
+            "[info] no running wasmCloud host detected (PID file missing), start a new instance..."
```
connorsmith256:
```suggestion
            "[info] no running wasmCloud host detected (PID file missing), starting a new host..."
```

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+                detached: true,
+                nats_opts: cmd.nats_opts,
+                wasmcloud_opts,
+                wadm_opts: cmd.wadm_opts,
+            },
+            output_kind,
+        )
+        .await?;
+        log::info!("started wasmCloud instance");
+        eprintln!("🛠  successfully started wasmCloud instance");
```
connorsmith256: Do we need both of these print statements?

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+        .context("failed to create ctl client")?;
+
+    // Get host information
+    let hosts = ctl_client
+        .get_hosts()
+        .await
+        .or_else(|e| bail!("failed to retrieve hosts from lattice: {e}"))?;
+    let host: Host = match &hosts[..] {
+        [] => bail!("0 hosts detected, is wasmCloud running?"),
+        [h] => h.clone(),
```
connorsmith256, line 226: I never thought to try destructuring a vec in Rust. Neat!

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+            }
+        }
+    };
+
+    // Resolve project configuration from the current path
+    let current_dir = std::env::current_dir()?;
+    let project_path = cmd.code_dir.unwrap_or(current_dir);
+    let project_cfg = get_config(Some(project_path.clone()), Some(true))?;
+
+    // Build the project (equivalent to `wash build`)
```
connorsmith256, line 247: I think you alluded to this as a refactor opportunity when we talked. Wanna file an issue so we could do something like `handle_build` in the future?

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+        .ok_or_else(|| anyhow!("failed to do thing"))?;
+    }
+
+    // Set up a oneshot channel to remove
+    let (stop_tx, mut stop_rx) = mpsc::channel::<()>(1);
+    let (reload_tx, mut reload_rx) = mpsc::channel::<()>(1);
+
+    // // Set up a handler for Ctrl + c
+    // ctrlc::set_handler(async move || stop_tx.send(())
+    //     .expect("failed to set Ctrl-C handler, please file a bug issue https://github.com/wasmCloud/wash/issues/new/choose")?;
```
connorsmith256: Assuming the tokio ctrl-c handler worked, can we remove this?

src/dev.rs
```diff
@@ -0,0 +1,285 @@
...
+                pause_watch.store(false, Ordering::SeqCst);
+                eprintln!("👀 watching for file changes (press Ctrl+c to stop)...");
+            },
+            _ = stop_rx.recv() => {
+                pause_watch.store(true, Ordering::SeqCst);
+                log::debug!("processing stop signal...");
+                eprintln!("🛑 received Ctrl + c, stopping devloop...");
+
+                eprintln!("⏳ stopping wasmCloud instance...");
+                handle_down(DownCommand{}, output_kind).await?;
```
connorsmith256: Hmm, how about the case where the user runs `wash dev` after a host has been started, and wants to keep it running after ctrl-c to exit the dev loop?

src/dev.rs
connorsmith256, line 1: Thoughts on moving some of this to `wash-lib`?

vados-cosmonic: So I wanted to, but even simple things like `DevCommand` can't really be moved simply because it re-uses `WasmcloudOpts` which has end-project details (defaults, etc)...

`run_dev_loop` can be taken out though, so I did that!

- [x] move `run_dev_loop` to `wash-lib`

src/dev.rs
vados-cosmonic: Ah that's reasonable -- I was thinking that mostly people would want to clean up the dev loop entirely afterwards (it's running with settings they didn't necessarily pick, etc). Will add an option though to leave it running

- [x] Add option to leave host running

src/dev.rs
vados-cosmonic: Yup! I was a bit worried the tokio one wouldn't work properly but it's just fine.

- [x] remove `ctrlc` handler

src/dev.rs
vados-cosmonic: Ahh so I actually ended up doing it here... But I am open to making a polish ticket!

src/dev.rs
vados-cosmonic: yeah, it's a super fun language feature!

src/dev.rs
vados-cosmonic: weirdly enough, yes i think so -- the logs could be going to a file or something else, but the stderr is what users are supposed to see... it's a bit hard to reconcile how it's done across wash...

thoughts on what we should do here? I'm ok removing the `log::info` here, assuming that whoever is reading the logs will have some other way to know the wasmcloud instance was successfully started

src/dev.rs
vados-cosmonic: Ahh thanks, nice catch

- [x] fix grammar on host detected message

src/dev.rs
vados-cosmonic: Yeah that's true

- [x] Fix `wash` mentions (should be `host`)

crates/wash-lib/src/config.rs
vados-cosmonic: Nope it can definitely be removed! 

- [x] remove `DEFAULT_NATS_PORT_INT`

src/dev.rs
vados-cosmonic: Ahh thanks, nice catch

- [x] fix grammar on host detected message

src/dev.rs
connorsmith256: I'd stick with whatever we've done elsewhere in interactive flows (e.g. `wash up`). Looks like that's probably just `println`, or sometimes macros from the `log` crate, but not both

vados-cosmonic: OK since the user is the prime receiver here I'll get rid of the `log::` message

- [x] remove log messages

vados-cosmonic:
Hey @connorsmith256 one more thing -- I forgot to put `wash dev` behind the `WASH_EXPERIMENTAL` flag, so I've added that

vados-cosmonic:
Looks like it takes a *really* long time for `wash` to set itself up to run in CI... 

`wash` is getting downloaded and the host starts running but the test quits before it has time to finish. Going to give it a bunch of time since the tests pass locally (and the check for finished happens every 5 seconds, not waiting the whole time, so it could *theoretically* finish early)

[EDIT] - Hmnn, something *must* be wrong in CI at this point -- Will need to turn on some visibility (and probably make that an option like `--show-logs`) to figure out what `wash up` is up to.

connorsmith256 (dismissed):
will merge once tests pass 🤞

connorsmith256 (approved):

PR #612, opened 2023-06-09
feat(capture): Adds a new experimental `wash capture` command

### LINKED ISSUE #601: [FEATURE] `wash capture`
Something that would be really helpful for debugging is to have a tool that could capture a "coredump" at a lattice level and be able to replay the last X minutes of events. This tool could become quite awesome and complex, but I would love to try a simple experiment first:

We should add something to wash (Let's call it `wash capture` for now) where it sets up a stream to capture all the invocations in a lattice for the last hour (configurable obviously). The stream can drop all the old invocations automatically for us. Then, when a user encounters a bug, they could just type `wash capture` and it would dump all the host inventories and all of the stream messages into a tarball for debugging. Then you can use the inventory to manually set up a similar environment (we could automate it in the future!) and then replay all of the events.

```plain
wash capture --enable
# When you have a bug
wash capture
Outputting capture to wash-capture-<timestamp>.tar.gz

# Now you need to debug
wash replay wash-capture-<timestamp>.tar.gz

# If for some reason you don't want to capture the sliding window of events
wash capture --disable
```

This experiment would be the most dead simple thing to start (DoD: Set up stream, dump inventories and stream on request), but you can see all the cool features this could enable like starting and stopping a replay.

This feature should most definitely be behind the experimental flag

thomastaylor312:
This one is very experimental, so I didn't even add it to the top level help text, but it is all manually tested and good to go. `wash capture` allows you to have a sliding window of captured events that you can dump to disk to debug any issues that occur

## Related Issues

Closes #601

## Release Information

0.18 if we get it in in time

## Consumer Impact

None, as it is a new feature

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

Unit tests to check the proper creation of the output tarball and that it can be read properly

### Acceptance or Integration

All manual verification

### Manual Verification

Manually verified all operations locally by enabling, disabling, capturing, and replaying

### DIFF
--- a/crates/wash-lib/src/cli/capture.rs
+++ b/crates/wash-lib/src/cli/capture.rs
@@ -1,0 +1,331 @@
+use std::{path::PathBuf, time::Duration};
+
+use anyhow::Result;
+use async_nats::jetstream::{
+    consumer::{pull::Config as ConsumerConfig, AckPolicy, DeliverPolicy},
+    stream::Config,
+};
+use clap::{Parser, Subcommand};
+use futures::TryStreamExt;
+use tokio::io::{stdin, stdout, AsyncReadExt, AsyncWriteExt};
+use tokio::time::Instant;
+use wasmbus_rpc::core::Invocation;
+
+use super::{CliConnectionOpts, CommandOutput};
+use crate::config::WashConnectionOptions;
+use crate::{
+    capture::{ReadCapture, WriteCapture},
+    id::{ModuleId, ServiceId},
+    spier::{ObservedInvocation, ObservedMessage},
+};
+
+pub const CAPTURE_STREAM_NAME: &str = "wash-capture";
+
+#[derive(Debug, Parser, Clone)]
+pub struct CaptureCommand {
+    /// Enable wash capture. This will setup a NATS JetStream stream to capture all invocations
+    #[clap(name = "enable", long = "enable", conflicts_with = "disable")]
+    pub enable: bool,
+
+    /// Disable wash capture. This will removed the NATS JetStream stream that was setup to capture
+    /// all invocations
+    #[clap(name = "disable", long = "disable", conflicts_with = "enable")]
+    pub disable: bool,
+
+    /// The length of time in minutes to keep messages in the stream.
+    #[clap(name = "window_size", long = "window-size", default_value = "60")]
+    pub window_size_minutes: u64,
+
+    #[clap(flatten)]
+    pub opts: CliConnectionOpts,
+
+    /// Replay through a stream of captured invocations
+    #[clap(subcommand)]
+    pub replay: Option<CaptureSubcommand>,
+}
+
+#[derive(Debug, Subcommand, Clone)]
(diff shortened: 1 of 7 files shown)


crates/wash-lib/src/cli/capture.rs
```diff
@@ -0,0 +1,331 @@
...
+    ))
+}
+
+pub async fn disable(
+    ctx: async_nats::jetstream::Context,
+    lattice_id: &str,
+) -> Result<CommandOutput> {
+    ctx.delete_stream(stream_name(lattice_id))
+        .await
+        .map_err(|e| anyhow::anyhow!("{e:?}"))?;
```
connorsmith256, line 210: Should we explicitly handle the case where disable is run before enable (so the stream never existed)? Just wondering how ugly of an error message we'll return in this case

crates/wash-lib/src/cli/capture.rs
```diff
@@ -0,0 +1,331 @@
...
+    let filename = format!(
+        "{}.{}.washcapture",
+        chrono::Local::now().to_rfc3339(),
+        lattice_id
+    );
+    let mut capture = WriteCapture::start(inventory, &filename).await?;
+
+    loop {
+        tokio::select! {
+            _ = expiry.tick() => {
```
connorsmith256, line 262: Am I reading this right? Users only have 1s after enabling capture to receive a message before we stop?

connorsmith256 (commented):
LGTM. Two questions (for one of which I assume I misinterpreted)

crates/wash-lib/src/cli/capture.rs
thomastaylor312: I want to come back once JS stuff has concrete errors so I can tell which error case it is

crates/wash-lib/src/cli/capture.rs
thomastaylor312: No, it is reading from the beginning of the stream, so as long as _something_ was running in there lattice there will be at least a few health check events. There are two interrupt possibilities. In a busy lattice, as soon as we see an invocation that occurs after when we started capture, we'll stop. In a less busy lattice, the last message might have been 30s ago, so this helps break it as the stream will play through really quickly (as fast as it can deliver) and will stop once there are no more messages

connorsmith256: I misunderstood the order of events. This happens _after_ the invocation, because the user types `wash capture` after things have happened. And we stop processing from the stream once there's been a 1s gap after the last delivery, which means we've "caught up" to now

connorsmith256 (approved):

PR #614, opened 2023-06-13
Release v0.18.0

brooksmtownsend:
## Feature or Problem
This PR updates the release actions for better release notes and fixes a few final issues for v0.18.0.

The issues:
- We didn't properly gate the CLI options in wash-lib's registry module, which is now fixed with an internal `mod`
- When the NATS pidfile was empty, we instead just sent the `stop` signal, which would stop even unmanaged NATS servers. Now, we return an error which will be reflected in the `wash down` output.

~I've also begun migrating our https://github.com/wasmCloud/homebrew-wasmcloud and https://github.com/wasmCloud/chocolatey-wash repositories here. I foresee some iteration especially with editing files via `sed` (which I may use templates to make less hacky looking) so in order to test via workflow dispatch and in a separate PR I'm just committing some base actions and the metadata from those repos.~

## Related Issues
~Begins work on #615~

## Release Information
v0.18.0

## Consumer Impact
Refer to the release notes for v0.18.0 for the full consumer impact.

## Testing
Tested individual issues manually to confirm resolution

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

PR #619, opened 2023-06-21
fix(ci): flaky tests

vados-cosmonic:
## Feature or Problem

This test fixes some tests that were flaky in CI and removes waits.

Tests have only become *slightly* faster (on the order of seconds).

Some additional fixes were made to the nix tests, but it's not clear why they're failing (now that all deps are present, it runs for 25 minutes...)

## Release Information

`next`

## Consumer Impact

Developers can more readily trust failing test runs to mean actual regressions.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -42,7 +42,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/install-nix
-      with: 
+      with:
         cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - uses: ./.github/actions/build-nix
       with:
@@ -66,11 +66,12 @@ jobs:
           test: ./result/bin/wash --version
 
     name: wash-${{ matrix.config.target }}
-    runs-on: macos-12
+    # NOTE(vados-cosmonic): Mac have OOMed in GitHub CI on macos-12 at least once
+    runs-on: macos-12-xl
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/install-nix
-      with: 
+      with:
         cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - uses: ./.github/actions/build-nix
       with:
@@ -125,7 +126,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/install-nix
-      with: 
+      with:
         cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - run: nix build -L .#checks.x86_64-linux.${{ matrix.check }}
 
@@ -134,7 +135,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: ./.github/actions/install-nix
-      with: 
+      with:
         cachixAuthToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - run: nix fmt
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -148,3 +148,6 @@ weld-codegen = "0.7.0"
 which = "4.4.0"
 chrono = "0.4.24"
 clap_complete = "4.3.0"
+
+[profile.dev.build-override]
+opt-level = 3
```
rvolosatovs: Why is this necessary? Considering that this option can slow down builds quite a bit and may not necessarily provide any performance benefits, this definitely requires a comment.
Also, why is this only done in `dev`? Why would we care about performance in `dev` profile?

For reference https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level

vados-cosmonic: Ah this was a retest of this option from early on, it was one of the first things I tried locally -- will remove this!

- [x] remove dev build override

connorsmith256 (dismissed):
LGTM! One item for consideration: it looks like we mostly call `wait_until_process_has_count` on `beam.smp` with a predicate where we check for a count of 0. In one case we check for `nats-server` with a count of 1. If these are the only reasonable use cases, I'd consider wrapping these blocks in functions like `wait_for_host_to_stop` and `wait_for_nats_to_start`. Then you could remove the comments on these blocks, since the function call is self-describing :)

vados-cosmonic:
Hey @connorsmith256 yeah that's reasonable, will add a helper function that wraps that re-used functionality at a higher level. 

One thing I don't want to do is to make it seem like we're actually waiting for the host in an intelligent way though -- might make people think we're *not* just grepping for processes (essentially), since we don't know the PIDs to wait on specifically necessarily.

While I'm in here I'm going to mark the macOS nix build as optional since sometimes it OOMs (and there's not much we can do about that right now)

connorsmith256 (dismissed):

vados-cosmonic:
One more rebase (I refactored and made the nix MacOS builds optional for now) -- @connorsmith256 if you wouldn't mind taking another look (and possibly merging!)

.github/workflows/nix.yml
```diff
@@ -67,10 +67,12 @@ jobs:
 
     name: wash-${{ matrix.config.target }}
     runs-on: macos-12
+    # NOTE(vados-cosmonic): Mac builds can OOM in GitHub CI
+    continue-on-error: true
```
rvolosatovs: The only place where I have seen this fail was the install step, exactly once. Rather than silently ignoring errors here and completely breaking the assumptions of this CI check, the issue should be fixed in `build-nix` as was discussed offline - use plain bundled `install` binary rather than using GNU coreutils one.
If you don't want to do this in this PR, then please file an issue, but IMO, the proposed change is not a solution for the problem

vados-cosmonic: And like we discussed offline, this is actually more complicated because `nix` did provide us a way to be cross-platform (in particular, we could ignore windows' lack of `install`) -- if you have some confidence that the error was only temporary/unlikely to reoccur, then I'll re-enable this step!

if not, I'd prefer to file an issue and have someone who is more experienced with `nix` take a look, and we can debug that way.

That said, there's another option here -- I'm going to mark `continue-on-error` only for *non* release branches. This way regular work can continue, but during a release, it has to pass.

vados-cosmonic:
Squashed once more, fixed the release check. 

To make it clear here, I'm expecting release branches [like this one](https://github.com/wasmCloud/wash/pull/614) to contain `release` somewhere (ex. `release/v0.18.0`) -- if this isn't standardized/process it won't work

connorsmith256 (approved):

PR #629, opened 2023-06-22
fix(up): Allows multiple hosts to run without sharing data

thomastaylor312:
## Feature or Problem

I found out when running some blobby tests that if you spin up multiple hosts, the NATS servers are separate, but they actually use the same data directory by default for jetstream. This means that two different locally running hosts _technically_ have the same streams and data available, which could lead to conflicts.

This segments it off into different data directories depending on the port the nats server is listening on. Technically there are still bugs when running two different nats servers as they write to the same log file, but we can solve that one later

## Release Information

This should probably be a patch release for 0.18 and for wash-lib. If we decide to do so, I'll follow up with a version bump

## Consumer Impact

Right now this is limited mostly to testing scenarios, so the current bug isn't super bad. But it is definitely annoying for users using wash for integration testing

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)

None

### Acceptance or Integration

None

### Manual Verification

I started two `wash up` processes with different NATS ports and validated that they were running with different store directories and information

### DIFF
--- a/crates/wash-lib/src/start/mod.rs
+++ b/crates/wash-lib/src/start/mod.rs
@@ -59,7 +59,7 @@ pub async fn wait_for_server(url: &str, service: &str) -> Result<()> {
     loop {
         // Magic number: 10 + 1, since we are starting at 1 for humans
         if wait_count >= 11 {
-            anyhow::bail!("Ran out of retries waiting for host to start");
+            anyhow::bail!("Ran out of retries waiting for {service} to start");
         }
         match tokio::net::TcpStream::connect(url).await {
             Ok(_) => break,
--- a/crates/wash-lib/src/start/nats.rs
+++ b/crates/wash-lib/src/start/nats.rs
@@ -215,7 +215,7 @@ leafnodes {{
             r#"
 jetstream {{
     domain={}
-    store_dir={}
+    store_dir={:?}
 }}
 {}
 "#,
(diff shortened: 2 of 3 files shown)


brooksmtownsend (dismissed):
Won't hold this back on it, but if you wanted to go ahead and change the NATS log file to be a combination of `nats_{port}.log` then that would help the log side of things too

thomastaylor312:
There are a bunch of things (like the wadm and wasmcloud files) that would need to change too, so we can do that as a follow up as it is a little less of a problem)

ricochet (dismissed):

thomastaylor312:
Ugh...looks like the fix might not work on windows. Will have to dig into it some more

brooksmtownsend:
> There are a bunch of things (like the wadm and wasmcloud files) that would need to change too

Feels like we may be better off setting up hosts to have their own directories in ~/.wash so that we run into as few problems as possible. Also worth looking more at wasmCloud/wasmCloud#846 to better define when we want to run multiple hosts on one machine

brooksmtownsend (approved):

PR #631, opened 2023-06-26
fix(up): the wadm pidfile is not removed when wadm is not started

### LINKED ISSUE #634: [BUG] Unsuccessful exit when using `wash up --disable-wadm`
**Describe the bug**

I use `wash up --disable-wadm` to start washcloud.
when I use ctrl-c to exit the program, it fails because `--disable-wadm` does not create the wadm process and the `wadm.pid` file, and removing the pidfile on exit will cause it to fail

**To Reproduce**
1. Run command 'wash up --disable-wadm'
2. . Run other command 'ctrl-c'
3. . See error
```bash
10:17:02.302 [info] Access WasmcloudHostWeb.Endpoint at http://localhost:4000

^C>>> ⢀⠀

No such file or directory (os error 2)
```

**Expected behavior**

wash up exits successfully when it receives `ctrl-c`

**Screenshots**

 - OS: [e.g. Linux, MacOS] MacOS
 - Shell [e.g. bash, zsh, powershell] zsh
 - wash Version [e.g. 0.2.0] 0.18.0

**Additional context**

Iceber:
## Feature or Problem

* Stops the wadm process and removes the wadm pidfile if other actions fail
* When wash does not start the wadm process, no pidfile files that may exist are removed at any time.

The pidfile file may have been created by another `wash up`.

Also, this pr ignores the case that the pidfile does not exist when it is removed, and the file should not cause a failure if it does not exist

## Related Issues

fix: https://github.com/wasmCloud/wash/issues/634
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -513,8 +513,10 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
             "CTRL+c received, stopping wasmCloud, wadm, and NATS...".to_string(),
         );
 
-        // remove wadm pidfile, the process is stopped automatically by CTRL+c
-        remove_wadm_pidfile(&install_dir).await?;
+        if wadm_process.is_some() {
+            // remove wadm pidfile, the process is stopped automatically by CTRL+c
+            remove_wadm_pidfile(&install_dir).await?;
+        }
 
         spinner.finish_and_clear();
     }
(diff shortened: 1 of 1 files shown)


brooksmtownsend (approved):
Great fix @Iceber ! Thank you for filing the issue and then taking care of it 🚀

PR #643, opened 2023-07-05
[feat] Add Detached Washboard

lachieh:
## Related Issues
- wasmcloud/wasmcloud#321

## Feature or Problem
This work comes out of [this wasmCloud RFC](https://github.com/wasmCloud/wasmCloud/issues/321) which aims to make the wasmCloud dashboard UI (AKA the washboard) detached from the host and able to connect to any lattice.

### Technology decisions
We discussed various frontend frameworks like Svelte and Solid, but ultimately decided on React because of community size and support. Here are some of the packages chosen alongside React:
- UI
  - [Tailwind](https://tailwindcss.com/) with [cva](https://cva.style) for utility-first approach with smallest stylesheet footprint
  - [Radix](https://radix-ui.com/) with [shadcn/ui](https://ui.shadcn.com/) for a large, accessibility focused, theme-able component library
  - [React Router](https://reactrouter.com/en/main) for best routing experience
  - [React Hook Form](https://react-hook-form.com/) with [zod](https://zod.dev/) for form handling and validation
- State management
  - [RxJs](https://rxjs.dev/) for event-based data handling (pairs nicely with NATS subscriptions)
  - [NATS.ws](https://github.com/nats-io/nats.ws) for communicating with lattice

### Features

The aim is to be as close to feature parity with the existing OTP host dashboard. Here is a breakdown/comparison of each dashboard

✅ - supported
⛔️ - not supported
⏱️ - planned
🔨 - in progress

| | OTP Host | Detatched |
|-|:-:|:-:|
| **Hosts** | | |
| View Status | ✅ | ✅ |
| Stop Host | ⛔️ | ⏱️ | 
| **Providers** | | |
| View Status | ✅ | ✅ |
| Deploy from file | ✅ | ⛔️ <sup>1</sup>|
| Deploy from registry | ✅ | ⏱️ |
| Remove | ✅ | ⏱️ |
| **Actors** | | |
| View Status | ✅ | ✅ |
| Deploy from file | ✅ | ⛔️ <sup>1</sup>|
| Deploy from file (hot-reload) | ✅ | ⛔️ <sup>1</sup>|
| Deploy from registry | ✅ | ⏱️ |
| Remove | ✅ | ⏱️ |
| **Links** | | |
| View Status | ✅ | ✅ |
| Define | ✅ | ⏱️   |
| Remove | ✅ | ⏱️ |
| **Lattice** | | |
| Connect to lattice by URL | ⛔️<sup>2</sup> | ✅ |
| Connect using credentials | ⛔️<sup>2</sup> | ⏱️ |

##### Notes
1. sending actor/provider bytes over NATS is not (yet?) supported
2. OTP-Host Washboard is built into host

## Considerations

This PR assumes only one method of distribution—bundling assets and serving via the `wash` cli tool—but this doesn't have to be the final solution. In fact, since the app itself compiles to static web assets, multiple options could be implemented with little effort maximizing flexibility. Alternates include:
- separately built binary that is only downloaded when `wash ui` is called for the first time
- as above, but packaged as a wash cli plugin
- bundled as a wasmCloud Actor component
- bundled as a Tauri application
- provided as static files only


package.json
connorsmith256, line 1: Is this needed at the top-level, or should it be part of `packages/package.json`?

packages/washboard/package.json
connorsmith256, line 1: Curious if we need `packages/washboard` instead of just `washboard` as a top-level directory?

packages/washboard/public/.gitkeep
connorsmith256, line 1: What goes in this dir?

packages/washboard/src/assets/.gitkeep
connorsmith256, line 1: Why is this needed?

packages/washboard/src/lib/utils.ts
```diff
@@ -0,0 +1,6 @@
+import {type ClassValue, clsx} from 'clsx';
+import {twMerge} from 'tailwind-merge';
+
+export function cn(...inputs: ClassValue[]): string {
```
connorsmith256, line 11: Nit: idk what `cn` means. Am I out of touch?

EDIT: classname?

packages/washboard/src/assets/.gitkeep
lachieh: 🗑️✅

packages/washboard/public/.gitkeep
lachieh: nada, 🗑️✅

packages/washboard/src/lib/utils.ts
lachieh: yah, added ts-doc comment

packages/washboard/package.json
lachieh: Was thinking that the `LatticeService` class may make it's way out into it's own npm package also in which case this makes it easy to add multiple npm packages/projects in future. Happy to move, though, if it feels weird.

package.json
lachieh: yeah, should be part of subfolder 🗑️✅

packages/washboard/package.json
lachieh: I will say that this structure is typical for npm/node monorepos

connorsmith256 (dismissed):

packages/washboard/package.json
connorsmith256: Thanks for the explanation 👍

src/ui/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+        .and(warp_embed::embed(&Asset))
+        .boxed();
+
+    let cors = warp::cors()
+        .allow_any_origin()
+        .allow_methods(vec!["GET", "POST"])
+        .allow_headers(vec!["Content-Type"]);
+
+    println!("Washboard running on http://localhost:3030");
+    println!("Hit CTRL-C to stop");
```
brooksmtownsend: nit: can we send this to stderr with `eprintln!`? That way if people are running in JSON output mode it won't break jq

src/main.rs
```diff
@@ -293,6 +299,7 @@ async fn main() {
         }
         CliCommand::Stop(stop_cli) => common::stop_cmd::handle_command(stop_cli, output_kind).await,
         CliCommand::Up(up_cli) => up::handle_command(up_cli, output_kind).await,
+        CliCommand::Ui(ui_cli) => ui::handle_command(ui_cli, output_kind).await,
```
brooksmtownsend: Make sure to gate this behind the experimental flag, if you mention that in the README

brooksmtownsend (changes requested):
One request, one nit that will probably never apply

src/main.rs
lachieh: broader question, should it be experimental? I could go either way

lachieh: Added it anyway to get it in. Easy enough to remove if necessary.

brooksmtownsend (dismissed):

build.rs
```diff
@@ -0,0 +1,54 @@
...
+};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-changed=./packages/washboard/src");
+    println!("cargo:rerun-if-changed=./packages/washboard/public");
+    println!("cargo:rerun-if-changed=./packages/washboard/package.json");
+    println!("cargo:rerun-if-changed=./packages/washboard/vite.config.js");
+    println!("cargo:rerun-if-changed=./packages/washboard/tsconfig.json");
+    println!("cargo:rerun-if-changed=./packages/washboard/tailwind.config.js");
+    println!("cargo:rerun-if-changed=./packages/washboard/postcss.config.cjs");
```
brooksmtownsend: @thomastaylor312 is this special syntax to actually not re-run this build if the static assets stay the same? Want to make sure that we aren't recomputing the npm commands each time you build

connorsmith256 (dismissed):

.github/workflows/nix.yml
```diff
@@ -1,5 +1,8 @@
 name: nix
 
+env:
+  WASH_CI: true
```
rvolosatovs: `nix` does not have access to any env variables by default, so pretty sure this is not effective

.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     tags:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
 
+env:
+  WASH_CI: true
```
rvolosatovs: see above

build.rs
```diff
@@ -0,0 +1,59 @@
...
+};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-env-changed=WASH_CI");
+    let running_in_ci = std::env::var("WASH_CI").is_ok();
+
+    // check if packages/washboard/dist/index.html exists, fail if running_in_ci, warn if not
+    let index_html_path = std::path::Path::new("./packages/washboard/dist/index.html");
+    match index_html_path.exists() {
+        false if running_in_ci => {
```
rvolosatovs: I don't think it's good developer experience to have to set `WASI_CI` to build UI changes.
Why not use something more like `WASH_BUILD_UI` ?

build.rs
```diff
@@ -0,0 +1,59 @@
...
+            // TODO: change this to fetch the latest release of the UI after we start publishing releases
+            // run npm install and npm run build if in CI
+            handle_output(
+                Command::new("npm")
+                    .args(["install", "--prefix", "packages/washboard"])
+                    .output()?,
```
rvolosatovs: Why not just let the command write to inherited stdio directly?

build.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+        false => {
+            println!("cargo:warning=Wash UI assets not found. Run `make build-ui` from the root of the repo or the `wash ui` command will not work correctly");
+            Ok(())
+        }
+        true => Ok(()),
+    }
+}
+
+fn handle_output(
```
rvolosatovs: This would be redundant if we just used the defaults above

build.rs
```diff
@@ -0,0 +1,59 @@
...
+                Command::new("npm")
+                    .args(["run", "build", "--prefix", "packages/washboard"])
+                    .output()?,
+                "npm run build",
+            )?;
+
+            Ok(())
+        }
+        false => {
+            println!("cargo:warning=Wash UI assets not found. Run `make build-ui` from the root of the repo or the `wash ui` command will not work correctly");
```
rvolosatovs: This does not work the way you think it does - you can't just log stuff using `println`. At least this should be using `eprintln`, which will only be printed on failure pretty much

build.rs
```diff
@@ -0,0 +1,59 @@
...
+    io::Write,
+    process::{Command, Output},
+};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo:rerun-if-env-changed=WASH_CI");
+    let running_in_ci = std::env::var("WASH_CI").is_ok();
+
+    // check if packages/washboard/dist/index.html exists, fail if running_in_ci, warn if not
+    let index_html_path = std::path::Path::new("./packages/washboard/dist/index.html");
```
rvolosatovs: This script should rerun each time any of the UI source or the generated assets change using https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath

build.rs
```diff
@@ -0,0 +1,59 @@
...
+                "npm run build",
+            )?;
+
+            Ok(())
+        }
+        false => {
+            println!("cargo:warning=Wash UI assets not found. Run `make build-ui` from the root of the repo or the `wash ui` command will not work correctly");
+            Ok(())
+        }
+        true => Ok(()),
```
rvolosatovs: We should either check the env var or check the existence of the file.
This is confusing, why not just always attempt to build if the env var is set?

connorsmith256 (approved):

connorsmith256:
Merging with failing windows build because it's failing everywhere

PR #644, opened 2023-07-06
fix(up): fix stop nats after starting wasmcloud host failed

Iceber:
## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -484,7 +484,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
             if let Some(child) = wadm_process {
                 stop_wadm(child, &install_dir).await?;
             }
-            if !cmd.nats_opts.connect_only {
+            if nats_bin.is_some() {
                 stop_nats(install_dir).await?;
             }
             return Err(e);


src/up/mod.rs
```diff
@@ -484,7 +484,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
             if let Some(child) = wadm_process {
                 stop_wadm(child, &install_dir).await?;
             }
-            if !cmd.nats_opts.connect_only {
+            if nats_bin.is_some() {
                 stop_nats(install_dir).await?;
```
brooksmtownsend, line 488: @Iceber are you seeing incorrect behavior when running `wash` currently? The current implementation ensures that wash will stop NATS as long as the flag `--nats-connect-only` isn't supplied. This change would attempt to stop NATS as long as it's installed, which would cause some unintended behavior.

brooksmtownsend (changes requested):

src/up/mod.rs
Iceber: I think we should probably decide to stop nats based on wash up having run nats.

When  `--nats-connect-only`  is not specified, but the nats client has access, wash up will not start nats, and then we will incorrectly stop nats on the same host that are started by other `wash up` when handling errors.

We can use two `wash up`s on the same host, but note that for the second `wash up` we need to have `start_wasmcloud_host` return the error.

Even if `wash up` is not running on the same host, the nats client can access the extra nats server, but it will still fail to stop because there are no nats running locally.
https://github.com/wasmCloud/wash/blob/b841d2124f4ea8ba1bfd65f7b2ad55fdb027ed13/src/down/mod.rs#L218-L220

This error will override the error returned by `start_wasmcloud_host` and be returned to the user

---

Also even if `--nats-connect-only` is specified and `remote url` is specified at the same time, this will not report an error at runtime, but will not stop the running nats server when `start_wasmcloud_host` returns an error (note: different wash up on the same host running nats at the same time will have an exception, this should be considered a special case bug) 
https://github.com/wasmCloud/wash/blob/b841d2124f4ea8ba1bfd65f7b2ad55fdb027ed13/src/up/mod.rs#L379-L383

I hope I haven't overlooked certain situations :rofl:

Iceber: @brooksmtownsend Hi, is there anything else that needs to be done for this pr？

brooksmtownsend: Hi @Iceber, I think I see what you're getting at, which is that `--nats-connect-only` isn't exactly the most descriptive way to say if NATS is running. After checking back over the code, it looks like I was making an incorrect assumption that `nats_bin` was the path to the binary even if we didn't launch it, but it looks like we only have that variable be `Some` if we _do_ launch NATS. In this case, I think this PR is good to go 😄

brooksmtownsend (approved):

PR #645, opened 2023-07-06
feature(up): ensure that the wasmcloud-host can be stopped gracefully

Iceber:
## Feature or Problem

the `wash up` can gracefully stop wasmcloud-host when it receives ctrl-c, for example by sending SIGTERM to it.

This allows the host to broadcast that it has stopped, and ensures that wasmcloud stops gracefully

In the previous logic, when terminating `wash up` with ctrl-c, nats and wadm would both receive the SIGINT signal and they would be shut down gracefully, but `wash up` uses SIGKILL to stop wasmcloud.

On unix, because wasmcloud was not stopped gracefully due to the use of SIGKILL, none of the providers it started were stopped and became orphaned processes

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/src/up/mod.rs
+++ b/src/up/mod.rs
@@ -476,7 +476,7 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
     let version = wasmcloud_opts.wasmcloud_version.clone();
 
     let host_env = configure_host_env(nats_opts, wasmcloud_opts).await;
-    let wasmcloud_child = match start_wasmcloud_host(
+    let mut wasmcloud_child = match start_wasmcloud_host(
         &wasmcloud_executable,
         std::process::Stdio::null(),
         stderr,
@@ -511,13 +511,15 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
 
     spinner.finish_and_clear();
     if !cmd.detached {
-        run_wasmcloud_interactive(wasmcloud_child, host_port, output_kind).await?;
+        run_wasmcloud_interactive(&mut wasmcloud_child, host_port, output_kind).await?;
 
         let spinner = Spinner::new(&output_kind)?;
         spinner.update_spinner_message(
-            // wasmCloud and NATS both exit immediately when sent SIGINT
+            // wadm and NATS both exit immediately when sent SIGINT
             "CTRL+c received, stopping wasmCloud, wadm, and NATS...".to_string(),
         );
+
+        stop_wasmcloud(wasmcloud_child).await?;
 
         if wadm_process.is_some() {
             // remove wadm pidfile, the process is stopped automatically by CTRL+c
@@ -601,7 +603,7 @@ async fn start_nats(install_dir: &Path, nats_binary: &Path, nats_opts: NatsOpts)
 
 /// Helper function to run wasmCloud in interactive mode
 async fn run_wasmcloud_interactive(
-    mut wasmcloud_child: Child,
+    wasmcloud_child: &mut Child,
     port: u16,
     output_kind: OutputKind,
 ) -> Result<()> {
@@ -646,7 +648,27 @@ async fn run_wasmcloud_interactive(
     if let Some(handle) = handle {
         handle.abort()
     };
+    Ok(())
+}
 
+#[cfg(all(unix))]
+async fn stop_wasmcloud(mut wasmcloud_child: Child) -> Result<()> {
+    use nix::sys::signal::{kill, Signal};
+    use nix::unistd::Pid;
+
+    if let Some(pid) = wasmcloud_child.id() {
+        // Send the SIGTERM signal to ensure that wasmcloud is graceful shutdown.
+        kill(Pid::from_raw(pid as i32), Signal::SIGTERM)?;
+
+        // TODO(iceber): the timeout for the SIGTERM could be added in the future,
+        // but it doesn't look like it's needed yet.
+        wasmcloud_child.wait().await?;
+    }
+    Ok(())
+}
+
+#[cfg(target_family = "windows")]
+async fn stop_wasmcloud(mut wasmcloud_child: Child) -> Result<()> {
     wasmcloud_child.kill().await?;
     Ok(())
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ notify = "6.0.1"
 rust-embed = "6.8.1"
 warp = "0.3.5"
 warp-embed = "0.4.0"
+nix = { workspace = true }
 
 [dev-dependencies]
 reqwest = { workspace = true, features = ["json", "rustls-tls"] }
@@ -75,7 +76,6 @@ futures = { workspace = true }
 # serial_test ensures serial execution when running with cargo, '<test name>_serial' works with nextest
 serial_test = { workspace = true }
 rand = "0.8.5"
-nix = { version = "0.26.2", default-features = false, features = [ "signal" ] }
 
 [[bin]]
 bench = true
@@ -151,3 +151,4 @@ weld-codegen = "0.7.0"
 which = "4.4.0"
 chrono = "0.4.26"
 clap_complete = "4.3.2"
+nix = { version = "0.26.2", default-features = false, features = ["signal"] }


src/up/mod.rs
```diff
@@ -506,14 +506,16 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
...
+        run_wasmcloud_interactive(&mut wasmcloud_child, host_port, output_kind).await?;
 
         let spinner = Spinner::new(&output_kind)?;
         spinner.update_spinner_message(
-            // wasmCloud and NATS both exit immediately when sent SIGINT
+            // wadm and NATS both exit immediately when sent SIGINT
             "CTRL+c received, stopping wasmCloud, wadm, and NATS...".to_string(),
         );
 
+        stop_wasmcloud(wasmcloud_child).await?;
```
Iceber, line 522: Place the `stop_wasmcloud` outside the `run_wasmcloud_interactive` to explicity stop wasmcloud, and the user can see the spinner output first

src/up/mod.rs
```diff
@@ -641,7 +643,28 @@ async fn run_wasmcloud_interactive(
...
+    use nix::sys::signal::{kill, Signal};
+    use nix::unistd::Pid;
+
+    // Send the SIGTERM signal to ensure that wasmcloud is graceful shutdown.
+    kill(
+        Pid::from_raw(wasmcloud_child.id().unwrap() as i32),
+        Signal::SIGTERM,
+    )?;
+
+    // TODO(iceber): the timeout for the SIGTERM cloud be added in the future,
```
brooksmtownsend: nit: typo in `could`

src/up/mod.rs
```diff
@@ -641,7 +643,28 @@ async fn run_wasmcloud_interactive(
...
+#[cfg(all(unix))]
+async fn stop_wasmcloud(mut wasmcloud_child: Child) -> Result<()> {
+    use nix::sys::signal::{kill, Signal};
+    use nix::unistd::Pid;
+
+    // Send the SIGTERM signal to ensure that wasmcloud is graceful shutdown.
+    kill(
+        Pid::from_raw(wasmcloud_child.id().unwrap() as i32),
+        Signal::SIGTERM,
+    )?;
```
brooksmtownsend: Instead of the `unwrap` here, I would prefer an `if let Some(id)` statement to ensure we don't panic, even if it's unlikely the PID is null

src/up/mod.rs
```diff
@@ -506,14 +506,16 @@ pub(crate) async fn handle_up(cmd: UpCommand, output_kind: OutputKind) -> Result
 
     spinner.finish_and_clear();
     if !cmd.detached {
-        run_wasmcloud_interactive(wasmcloud_child, host_port, output_kind).await?;
+        run_wasmcloud_interactive(&mut wasmcloud_child, host_port, output_kind).await?;
 
         let spinner = Spinner::new(&output_kind)?;
         spinner.update_spinner_message(
-            // wasmCloud and NATS both exit immediately when sent SIGINT
+            // wadm and NATS both exit immediately when sent SIGINT
```
brooksmtownsend, line 518: We'll have to double check this behavior, I _think_ wadm, NATS, and wasmCloud all get terminated when CTRL+c is received 🤔 Were you able to verify that your stop_wasmcloud function sends the SIGTERM and that's what stops wasmCloud?

brooksmtownsend (changes requested):
A couple nits, and just one question

src/up/mod.rs
Iceber: I found that wadm and nats automatically terminate after receiving ctrl-c, but wasmcloud doesn't, it needs to call  `kill` or `send a signal`
This is actually caused by 
https://github.com/wasmCloud/wash/blob/7265d9e9bde1192de847f2bcaf8e508b31a6f682/crates/wash-lib/src/start/wasmcloud.rs#L248-L251

Although I'm not sure why a process group is created for wasmcloud here, this way I can send some signals(SIGKILL, SIGTEM, etc...) to decide how to terminate wasmcloud

Iceber: So for this comment, it's true that only wamd and nats will be exited

connorsmith256: I believe @brooksmtownsend is right and all three child processes are terminated after a CTRL-C. For wasmCloud this is the relevant line (see the "Wait for the user to send Ctrl+C" message above it):
https://github.com/wasmCloud/wash/blob/87033c94bf4f27b7936b5e1191f929562a23554e/src/up/mod.rs#L645

Iceber: > I found that wadm and nats automatically terminate after receiving ctrl-c, but wasmcloud doesn't, it needs to call kill or send a signal

@connorsmith256 Sorry, my reply above may not have explained it well.

Yes, all three processes started by wash up will eventually be terminated, but they are not terminated in the same way.
wadm and nats will be **terminated immediately** after wash up receives the ctrl-c signal, while
wasmcloud is not terminated by the **ctrl-c**, but by calling `wasmcloud.kill()` or by using [`stop_wasmcloud`](https://github.com/wasmCloud/wash/pull/645/files#diff-dcf780801376c219d3b5d7f2f9a5a2745c5ccb124b8c621281e13d3c775ed9f3R650)

I moved `stop_wasmcloud` under `spinner.message("CTRL+c received, stopping wasmCloud, wadm, and NATS...".to_string())` to explicitly stop wasmcloud, because there is a short interval between sending the signal to wasmcloud and waiting for wasmcloud to exit, and it would be nice to print spinner.message before then.

After `stop_wasmcloud` is moved under `spinnner.message`, **only wadm and nats are **immediately stopped** by the **ctrl-c** before this comment**, while wasmcloud is still waiting for the stop_wasmcloud call.
so I changed the comment to `wadm and NATS both exit immediately when sent SIGINT`

src/up/mod.rs
Iceber: updated, thanks

Iceber:
> On unix, because wasmcloud was not stopped gracefully due to the use of SIGKILL, none of the providers it started were stopped and became orphaned processes

@brooksmtownsend @connorsmith256 i updated the descritpiton of pr. This looks like a serious bug, SIGKILL is a dangerous shutdown signal.

brooksmtownsend:
> > On unix, because wasmcloud was not stopped gracefully due to the use of SIGKILL, none of the providers it started were stopped and became orphaned processes
> 
> @brooksmtownsend @connorsmith256 i updated the descritpiton of pr. This looks like a serious bug, SIGKILL is a dangerous shutdown signal.

@Iceber I remember when initially creating this feature, I couldn't find a great way to spawn a child process and then intercept the CTRL+c because it's sent immediately down to the children of the foreground. I'm not sure if you know a way to do this, since the mechanics of the SIGKILL getting sent to children too isn't something we can control

Iceber:
> @Iceber I remember when initially creating this feature, I couldn't find a great way to spawn a child process and then intercept the CTRL+c because it's sent immediately down to the children of the foreground. I'm not sure if you know a way to do this, since the mechanics of the SIGKILL getting sent to children too isn't something we can control

CTRL-c is actually sending a SIGINT signal to the process group/session. So `nats` and `wadm` receive SIGINT signals, whereas `wasmcloud` is created in a new process group, and it doesn't receive SIGINT signals, so if we don't call `wasmcloud.kill` to shut it down, it will become an orphaned process, but `wasmcloud.kill` sends SIGKILL signals, which will make the wasmcloud to lose some of its termination operations. So I chose to use SIGTERM instead of SIGKILL signal, and the provider subprocesses created by wasmcloud should be managed by wasmcloud itself, and it is true that when `wasmcloud-host` receives SIGTERM/SIGINT it will terminate all provider processes.

In my opinion, wash up doesn't need to manage the sub-processes created by wasmcloud, it's managed by wasmcloud, what wash up can do is to send a graceful shutdown signal to wasmcloud.

connorsmith256:
@Iceber Ah, thank you for the clarification! I think I understand now. I agree we should send SIGTERM to the wasmcloud host, which will give it time to gracefully shut down, including emitting the `host_stopped` event.

However, there's one tricky part of making this work in practice: the NATS server process will be killed immediately after a CTRL-C, before we gracefully shut down the wasmcloud host with SIGTERM. But since the NATS server will be dead already, the host won't be able to emit the `host_stopped` event.

I haven't thought much about this, but perhaps the solution needs to involve moving the NATS server to a new process group as well, so it's not immediately killed with a CTRL-C. Then we could control the shutdown logic and send a SIGTERM to the host before the NATS server process

Iceber:
@connorsmith256 I'm not sure if I'm overlooking something.
A wash up starts nats and does not connect to external nats, when the wash up receives a ctrl-c, it kills the nats, and any other connections to the nats will not work, it doesn't seem to matter whether or not the wasmcloud of the terminated wash up sends a successful host_stopped on shutdown, it just needs to do other things, such as recycle its provider subprocess. It just needs to do other things, such as reclaiming its provider subprocesses.

connorsmith256:
@Iceber There are two important steps for a wasmCloud host to shut down gracefully:
1) stop provider processes
2) emit a `host_stopped` event

Your refactor improves the current behavior by replacing a SIGKILL with a SIGTERM, which allows the host to perform step 1). However, we kill the local NATS server process before 2)

The second step is important in environments where there are multiple wasmCloud hosts running. E.g. if someone is using wadm to distribute an application across hosts, and pressing CTRL-C causes a host to shut down without emitting the `host_stopped` event, then it will take a couple of minutes for wadm to detect the host has become unavailable and attempt to reschedule the application

That being said, fixing 1) is still an improvement! I'm happy to merge this PR as-is, but I don't think we should close https://github.com/wasmCloud/wasmcloud-otp/issues/643 until we also fix 2) (thank you for filing that issue, by the way)

@brooksmtownsend I suggest we move https://github.com/wasmCloud/wasmcloud-otp/issues/643 back to the wash repo, since fixing 1) and 2) above are wash-specific behavior. If the washboard doesn't expire hosts which haven't emitted heartbeats, that feels like a separate bug. Thoughts?

connorsmith256:
@Iceber I think all this needs is a rebase and we can merge

Iceber:
@connorsmith256 @brooksmtownsend I am sorry, I was working on an important feature for another project the other day and I can now continue working on wasmcloud.

Especially in leaf nats mode, we really should be needing to shut down nats last, and maybe we really need to create a new process group for it.

There is also a signal handling issue, does wash up need to be running in a container, `wash up` doesn't handle SIGTERM signals, it only handles SIGINT.

And when `wasm up` receives SIGTERM/SIGINT instead of CTRL-C, it doesn't shut down nats and wadm.

connorsmith256 (approved):
As discussed, this doesn't entirely solve the problem, but it is a step in the right direction!

Thank you for your contribution @Iceber 🙌

brooksmtownsend (approved):

PR #651, opened 2023-07-07
fix typo im Completions.md

voigt:
## Feature or Problem

I just noticed that there is a typo in the docs about adding completions for bash. Once the completions are added to `.bashrc` the `.bashrc` needs to be sourced. Not the `.zshrc` :)


brooksmtownsend (approved):
but zshrc is cooler!!!!!!!!

Great catch @voigt 😄

Issue #640, opened 2023-07-05
[FEATURE] Support for WASI Preview 2 components


vados-cosmonic:
**Is your feature request related to a problem? Please describe.**

As the ecosystem moves from WASI preview1 to WASI preview2 compnents, `wash build` is not yet capable of building preview2 components (most language ecosystems can't either, just yet). 

At this point, most ecosystems can (and do, via `wash build`) generate preview1 components just fine -- but preview1 need to [be "adapted" to WASI preview2](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-preview1-component-adapter).

The easiest place to make this happen automatically (and in a way that `wash` can control) is in the Rust ecosystem. Rust projects being built that target `wasi32-wasi` are preview1, not preview2 -- Rust has actually [renamed `wasm32-wasi` to `wasm32-wasi-preview1` to reflect this reality](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-preview1-component-adapter). 

As Rust has the most robust ecosystem for dealing with preview2 components, it makes sense to leverage the [`wasm-tools` crates](https://github.com/bytecodealliance/wasm-tools/) here and add a step that does the necessary adaptation so that users can more easily gravitate to WASI preview2.

This also goes hand in hand with the work being done on the [Rust host in `wasmcloud/wasmcloud`](https://github.com/wasmCloud/wasmCloud/tree/main/crates/host/src), as that host will be expecting to deal with components more often than modules once things settle.

**Describe the solution you'd like**

Enable `wash build` to support WASI preview2 components, without requiring the user to do manual work.

I think this could work in mostly the following steps:

- [x] Add a way to indicate that the user *wants* to build for WASI preview2 (in `washcloud.toml`)
- [x] Figure out a way to download and manage the adapter modules (these can/do change over time)
- [x] Modify wash to adapt preview1 components into preview2 components after building them (something like `build_s.preview1.wasm` and `build_s.preview2.wasm` might make sense)

**Describe alternatives you've considered**

Allowing users to run `wasm-tools compose` manually to generate components.

**Additional context**

vados-cosmonic:
Hey this is done and completed, so closing!

PR #656, opened 2023-07-10
feat: update versions, save component adapters in-tree

vados-cosmonic:
## Feature or Problem

Build the [WASI preview1 component adapter](https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-preview1-component-adapter) inside `wash`

Note that the adapter is *not* built inside `wash-lib` because it requires the unstable `bindeps` feature. 

It seemed somewhat undesirable to require nightly Rust for the `wash` build when it actually doesn't require it directly, so I instead require that the toolchain is *installed*, and build the adapter in a temporary directory and output the path so `wash` can use it (see sections of `build.rs`).

Of course, nightly Rust as a requirement for everyone who uses `wash-lib` is pretty painful, so the code to do the adapting is going to need to go into `wash` (not `wash-lib`).

All this is gated behind the `wasi-p1-adapter` flag which can be enabled when building `wash`.

## Related Issues

#640 

## Release Information

`next`

## Consumer Impact

Consumers will be able to build the p1 adapter locally

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Verified locally


build.rs
```diff
@@ -0,0 +1,217 @@
+use std::{
```
rvolosatovs: Maybe add a reference to the original at the top of the file?

build.rs
```diff
@@ -0,0 +1,217 @@
...
+            .map(|(pkg, paths)| (pkg.as_str(), paths.as_slice()))
+    }
+}
+
+async fn build_wasi_p1_adapter() -> Result<()> {
+    println!("cargo:rerun-if-changed=./crates/wash-wasi-preview1-adapter");
+
+    // Ensure that cargo nightly is supported
+    Command::new("cargo")
+        .args(["+nightly", "version"])
```
rvolosatovs: Let's not explicitly depend on rustup here, just remove the `+nightly`

build.rs
```diff
@@ -0,0 +1,217 @@
...
+        .output()
+        .await
+        .context("cargo nightly version failed, is cargo nightly installed?")?;
```
rvolosatovs: This should not be necessary, instead we should select the right version in `rust-toolchain`

.github/workflows/rust_ci.yml
```diff
@@ -30,6 +31,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
+      - run: rustup toolchain install nightly --profile minimal
```
rvolosatovs: This should not be necessary, instead we should just use `rust-toolchain.toml`

rvolosatovs (changes requested):

rvolosatovs (commented):
The built adapter is never actually used anywhere, so what's the purpose of building it at all?

build.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    // Build a temporary directory for building the WASI preview1 adapter
+    let tmp_dir = std::env::temp_dir().join("wash-wasi-preview1-adapter");
+    let crate_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("./crates/wash-wasi-preview1-adapter")
+        .canonicalize()?;
+
+    // Copy the code from the inner crate to out to the temporary directory
+    fs::create_dir_all(&tmp_dir).await?;
```
rvolosatovs: Why is all of this required?

build.rs
```diff
@@ -0,0 +1,217 @@
...
+        copy_file(crate_path.join("src/lib.rs"), tmp_dir.join("src/lib.rs"),),
+    )
+    .context("failed to copy over files")?;
+
+    // Build the preview1 adapter
+    build_wasi_preview1_adapter(&tmp_dir, &tmp_dir).await?;
+
+    // Ensure that the wasm got built
+    let output_wasm_path = tmp_dir.join("wasi-preview1-component-adapter.wasm");
+    if fs::try_exists(&output_wasm_path).await.is_err() {
```
rvolosatovs: This check is redundant

rvolosatovs (changes requested):

build.rs
vados-cosmonic: Sure! 

- [x] Add reference to testing crates

build.rs
vados-cosmonic: Sure, I'll remove this if `rustup-toolchain.toml` clears all the other issues as mentioned in the other comments

[EDIT] ah, this may not be good when I think of the implications of it -- [see this comment](https://github.com/wasmCloud/wash/pull/656#discussion_r1258578139). 

`wash` itself doesn't need nightly rust, so it seems weird to require it for the whole project. Requiring nightly for a dependency behind a flag is more reasonable.

If we're requiring nightly rust for the whole project that seems like a change that should be discussed, IMO

build.rs
vados-cosmonic: Please see the description of this ticket -- `wash` itself doesn't actually require nightly, so it doesn't seem reasonable to make the whole project nightly via `rust-toolchain`.

.github/workflows/rust_ci.yml
vados-cosmonic: The CI build was failing, which is why this was added -- please see also [the comment above](https://github.com/wasmCloud/wash/pull/656#discussion_r1258578139) with regards to `rust-toolchain.toml`

build.rs
vados-cosmonic: Please see the description of this PR

build.rs
vados-cosmonic: Sure, will remove, and will update the function to output the actual path it writes to.

- [x] remove check for output wasm
- [x] update fn to return output wasm path

vados-cosmonic:
Oh also, to address [this comment](https://github.com/wasmCloud/wash/pull/656#pullrequestreview-1522514579) -- this is a stepping stone on the way to using the adapter in #640

build.rs
vados-cosmonic: There is also a [discussion on the wasmcloud slack now](https://wasmcloud.slack.com/archives/C04RVMJJH17/p1689010717433159), please put feedback there!

crates/wash-lib/build.rs
```diff
@@ -0,0 +1,212 @@
...
+        copy_file(
+            crate_path.join("rust-toolchain.toml"),
+            tmp_dir.join("rust-toolchain.toml"),
+        ),
+        copy_file(crate_path.join("src/lib.rs"), tmp_dir.join("src/lib.rs"),),
+    )
+    .context("failed to copy over files")?;
+
+    // Build the preview1 adapter
+    let output_wasm_path = build_wasi_preview1_adapter(&tmp_dir, &tmp_dir)
```
connorsmith256: This name is confusing, given the function we're in is `build_wasi_p1_adapter`

connorsmith256 (commented):
Looks like the build is still failing, with the super-helpful `No such file or directory (os error 2)`

crates/wash-lib/build.rs
vados-cosmonic: changed this to match the name!

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/wasmtime"]
+	path = vendor/wasmtime
+	url = https://github.com/rvolosatovs/wasmtime
```
rvolosatovs, line 3: I thought we agreed on using a filtered subtree for this?

docs/DEVELOPMENT.md
```diff
@@ -57,6 +66,29 @@ To build continuously (thanks to [`cargo-watch`](https://crates.io/crates/cargo-
...
 ```
 
+### Building the WASI preview1 component adapters
+
+If you find it necessary to rebuild the preview1 components adapters, you must use `git submodule` (as [`wasmtime`][wasmtime] does). After checking out this repository:
+
+```console
+git submodule update --init # initialize the vendor/wasmtime submodule
+cd vendor/wasmtime
+git submodule update --init # intiialize the submodules inside wasmtime
```
rvolosatovs, line 76:
```suggestion
git submodule update --init # intialize the submodules inside wasmtime
```

Cargo.toml
```diff
@@ -63,8 +63,8 @@ warp = "0.3.5"
 warp-embed = "0.4.0"
 
 [dev-dependencies]
```
rvolosatovs, line 65: I think that dependency updates are out-of-scope of this PR and belong in a separate one

README.md
```diff
@@ -220,7 +220,6 @@ Perform validation checks on .smithy models, ensuring that your interfaces are v
 See [Completions](./Completions.md) for instructions for installing
 autocomplete for your shell.
 
-
```
rvolosatovs, line 223: How is this change relevant to the PR?

docs/DEVELOPMENT.md
```diff
@@ -13,14 +13,23 @@ This document serves as a guide and reference for people looking to develop `was
 
 ## Development Prerequistes
 
-To contribute to `wash`, you just need [Rust](https://rustup.rs/) installed. To run any `wash` tests, you need to install [`nextest`](https://nexte.st/index.html). With a Rust toolchain already installed, you can simply install this with:
+To contribute to `wash`, you just need [Rust](https://rustup.rs/) installed.
+
+Due to the use of the as-of-yet unstable [`bindeps` feature][bindeps], `wash` also requires [nightly rust][rust-nightly] to
```
rvolosatovs, line 18: Does this really still apply? I do not see bindeps being used

flake.nix
```diff
@@ -92,6 +92,9 @@
           "snap"
           "tools"
 
+          "vendor/wasm/wasi_snapshot_preview1.command.wasm"
```
rvolosatovs, line 95: I don't understand why these are in-tree, but excluded from the build

crates/wash-lib/Cargo.toml
```diff
@@ -19,6 +19,7 @@ start = ["semver"]
 parser = ["config", "semver", "serde", "serde_json"]
 cli = ["clap", "term-table", "console", "dialoguer", "heck", "ignore", "indicatif", "path-absolutize"]
 nats = ["async-nats", "wadm"]
+wasi-p1-adapter = []
```
rvolosatovs, line 22: What does this feature actually do? Isn't it completely redundant?

rvolosatovs (changes requested):

rvolosatovs:
I really do not think we should be committing compiled binaries into the tree

vados-cosmonic:
Closing in favor of https://github.com/wasmCloud/wash/pull/663 -- going to download the binary at runtime and use from `wash_lib`

PR #663, opened 2023-07-14
feat: support adapting p2 components

vados-cosmonic:
## Feature or Problem

This PR adds functionality to `wash` and `wash-lib` to enable adapting WASI preview 2 components.

Since we want the functionality in `wash-lib` (for easy re-use in `cosmo`) yet do not necessarily want the adapters to be included in `wash-lib`, the integration works this way:

- adapter binary contents are packed into `wash`
- `wash-lib` accepts a a reference to a slice of bytes that represents the adapter, through various `build_xxxx` functions that work to build the actor.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ atelier_core = { workspace = true }
 bytes = { workspace = true }
 cargo_atelier = { workspace = true }
 clap = { workspace = true, features = ["derive", "env"] }
+clap_complete = { workspace = true }
 cloudevents-sdk = { workspace = true }
 console = { workspace = true }
 dirs = { workspace = true }
@@ -29,7 +30,9 @@ envmnt = { workspace = true }
 futures = { workspace = true }
 indicatif = { workspace = true }
 log = { workspace = true }
+nix = { workspace = true }
 nkeys = { workspace = true }
+notify = "6.0.1"
 oci-distribution = { workspace = true, features = ["rustls-tls"] }
 once_cell = { workspace = true }
 provider-archive = { workspace = true }
@@ -37,6 +40,7 @@ regex = { workspace = true }
 remove_dir_all = { workspace = true }
 rmp-serde = { workspace = true }
 rmpv = { workspace = true }
+rust-embed = "6.8.1"
 sanitize-filename = { workspace = true }
 semver = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
@@ -48,34 +52,30 @@ term-table = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 toml = { workspace = true }
-which = { workspace = true }
 wadm = { workspace = true }
-wash-lib = { workspace = true, features = ["cli"] }
-wascap = { workspace = true }
-weld-codegen = { workspace = true }
-wasmcloud-control-interface = { workspace = true }
-wasmbus-rpc = { workspace = true }
-wasmcloud-test-util = { workspace = true }
-clap_complete = { workspace = true }
-notify = "6.0.1"
-rust-embed = "6.8.1"
 warp = "0.3.5"
 warp-embed = "0.4.0"
-nix = { workspace = true }
+wascap = { workspace = true }
+wash-lib = { workspace = true, features = ["cli"] }
+wasmbus-rpc = { workspace = true }
+wasmcloud-control-interface = { workspace = true }
+wasmcloud-test-util = { workspace = true }
+weld-codegen = { workspace = true }
+which = { workspace = true }
 
 [dev-dependencies]
+assert-json-diff = { workspace = true }
+chrono = { workspace = true }
+futures = { workspace = true }
+nix = { version = "0.26.2", default-features = false, features = [ "signal" ] }
+rand = "0.8.5"
 reqwest = { workspace = true, features = ["json", "rustls-tls"] }
+scopeguard = { workspace = true }
+serial_test = { workspace = true } # serial_test works with cargo, '<test name>_serial' works with nextest
+sysinfo = { workspace = true }
 tempfile = { workspace = true }
 test-case = { workspace = true }
 test_bin = { workspace = true }
-assert-json-diff = { workspace = true }
-scopeguard = { workspace = true }
-sysinfo = { workspace = true }
-chrono = { workspace = true }
-futures = { workspace = true }
-# serial_test ensures serial execution when running with cargo, '<test name>_serial' works with nextest
-serial_test = { workspace = true }
-rand = "0.8.5"
 
 [[bin]]
 bench = true
@@ -95,8 +95,10 @@ atelier_core = "0.2"
 bytes = "1.4"
 cargo_atelier = "0.2"
 cargo_toml = "0.15.3"
+chrono = "0.4.26"
 claims = "0.7.1"
 clap = "4"
+clap_complete = "4.3.2"
 cloudevents-sdk = "0.7.0"
 command-group = "1.0.8"
 config = "0.13.1"
@@ -110,6 +112,7 @@ heck = "0.4"
 ignore = "0.4"
 indicatif = "0.17.5"
 log = "0.4"
+nix = { version = "0.26.2", default-features = false, features = ["signal"] }
 nkeys = "0.3.0"
 oci-distribution = { version = "0.9.4", default-features = false, features = ["rustls-tls"] }
 once_cell = "1.18"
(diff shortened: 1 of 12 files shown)


rvolosatovs (changes requested):
Since https://github.com/wasmCloud/wasmCloud/pull/378 https://github.com/wasmCloud/wasmcloud now contains `wasmcloud_component_adapters` crate, which exports the adapters as const byte slices, meaning that there is no need to build the adapters in this repository and also no need to check-in the Wasm files. Could you add (git) dependency on that crate please and use the consts instead?

rvolosatovs:
I have filed https://github.com/wasmCloud/wash/pull/670, which contains all the necessary Nix changes to make this work.

Cargo.toml
```diff
@@ -140,15 +144,14 @@ tokio-stream = "0.1"
 tokio-tar = "0.3"
 tokio-util = "0.7.8"
 toml = "0.7.6"
-wadm = "0.4.0"
+wadm = "0.4.1"
 walkdir = "2.3"
 wascap = "0.11.0"
 wash-lib = { version = "0.9.0", path = "./crates/wash-lib" }
 wasmbus-rpc = "0.14.0"
+wasmcloud-component-adapters = { git = "https://github.com/wasmCloud/wasmCloud", rev = "a339ede" }
```
rvolosatovs: Do we really need the rev here?

src/build.rs
```diff
@@ -89,6 +93,24 @@ pub(crate) fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
     }
 }
 
+/// Retrieve bytes for WASI preview2 adapter given a project configuration
+pub(crate) async fn get_wasi_preview2_adapter_bytes(
+    config: &ProjectConfig,
+) -> Result<Option<Vec<u8>>> {
```
rvolosatovs: When would this ever return a `Ok(None)`? What's the purpose of the `Option` here?

crates/wash-lib/src/build.rs
```diff
@@ -76,17 +92,42 @@ pub fn build_actor(
     language_config: &LanguageConfig,
     common_config: &CommonConfig,
     signing_config: Option<SignConfig>,
+    adapter_bytes: &Option<impl AsRef<[u8]>>,
```
rvolosatovs:
```suggestion
    adapter_bytes: impl AsRef<[u8]>,
```

src/build.rs
```diff
@@ -89,6 +93,24 @@ pub(crate) fn handle_command(command: BuildCommand) -> Result<CommandOutput> {
...
+    match &config.project_type {
+        TypeConfig::Actor(ActorConfig {
+            wasi_preview2_adapter_path: Some(path),
+            ..
+        }) => tokio::fs::read(&path)
+            .await
+            .map(Option::from)
+            .with_context(|| format!("failed to read wasm bytes from [{}]", path.display())),
+        _ => Ok(Some(
+            wasmcloud_component_adapters::WASI_PREVIEW1_REACTOR_COMPONENT_ADAPTER.into(),
```
rvolosatovs: We should make a distinction here whether the component is a "reactor" or "command". I suppose it's fine to leave it out for now, but there should be an issue filed about that

Cargo.toml
vados-cosmonic: any reason we shouldn't be explicit on which version we're pulling in?

crates/wash-lib/src/build.rs
vados-cosmonic: Sometimes, the adapter bytes are not needed (i.e. all the other non-p2-component builds)

src/build.rs
vados-cosmonic: We don't have anything that uses command, as far as I know -- we should be able to add that once we have a use case

Cargo.toml
rvolosatovs: Having the `rev` here means we cannot simply `cargo update` anymore and would have to manually check the `rev` and copy-paste it here, which complicates and slows down maintenance

src/build.rs
vados-cosmonic: Ah was for easy use where this function is used, I can remove the option here and add it @ the use site.

~~- [ ] Remove `Option` from `get_wasi_preivew2_adapter_bytes`~~

vados-cosmonic: Looking at this more -- what I'm missing is the `None` case (which represents all the other non-component builds) -- so I'll add that instead

vados-cosmonic: addressed -- missing case (where you have `None` is there now

Cargo.toml
rvolosatovs: @vados-cosmonic could you switch to https://github.com/rvolosatovs/wasmcloud-component-adapters (at least temporarily?)
We should move that repo to `wasmCloud` org, but I did not have permissions to do that.

The issue here is that in `cargo vendor` use cases, build can only refer to files within the crate and `flake.lock` in this case is outside of it. Seems that a separate repo is the only reasonable way to handle this case.

vados-cosmonic: Yeah I think we should be able to make that other repo really quick, so I'm fine waiting for that to happen then switching

brooksmtownsend: @rvolosatovs can you give me ownership over your repository? I can initiate the move then

crates/wash-lib/src/build.rs
```diff
@@ -49,11 +56,20 @@ pub struct SignConfig {
 /// # Arguments
 /// * `config`: [ProjectConfig] for required information to find, build, and sign an actor
 /// * `signing`: Optional [SignConfig] with information for signing the project artifact. If omitted, the artifact will only be built
-pub fn build_project(config: &ProjectConfig, signing: Option<SignConfig>) -> Result<PathBuf> {
+/// * `adapter_bytes`: Optional [&[u8]] bytes that represent a wasm component adapter that should be used, if present.
+pub fn build_project(
+    config: &ProjectConfig,
+    signing: Option<SignConfig>,
+    adapter_bytes: &Option<impl AsRef<[u8]>>,
```
connorsmith256: Since the adapter bytes are retrieved based off the `config`, does this need to be a separate arg, or can we have `build_actor` be responsible for calling `get_wasi_preview2_adapter_bytes`? That would simplify the call chain

I assume wash-lib could contain the bytes(?)

connorsmith256 (dismissed):
> yet do not necessarily want the adapters to be included in wash-lib

Can you help me understand reasons for/against including the adapters in wash-lib?

crates/wash-lib/src/build.rs
vados-cosmonic: This was brought up during a previous review which I agree with -- if we can we should avoid `wash-lib` holding the lib, because then every downstream will need to pull in the blobs (which is large, for a regular lib), the PR was written with the intent to avoid pulling in the lib (which required those contortions).

For example `tokio`weighs in at 694kb, and wash lib currently weighs in at 85kb. I'm not sure it's something people will care about that much but given that the adapters are ~100+kb each, that would be a significant increase

Also, it's very much the case that a downstream of `wash-lib` could have their own way of fetching/using adapter bytes (and not want to incur the cost of pulling in those extra bytes for an adapter they don't use)

Cargo.toml
```diff
@@ -48,34 +52,32 @@ term-table = { workspace = true }
...
 warp = "0.3.5"
 warp-embed = "0.4.0"
-nix = { workspace = true }
+wascap = { workspace = true }
+wash-lib = { workspace = true, features = ["cli"] }
+wasmbus-rpc = { workspace = true }
+wasmcloud-component-adapters = { workspace = true }
+wasmcloud-control-interface = { workspace = true }
+#wasmcloud-test-util = { workspace = true }
+wasmcloud-test-util = { git = "https://github.com/vados-cosmonic/wasmcloud-test.git", branch = "refactor/remove-smithy-bindgen-dep" }
```
connorsmith256: btw, v0.10 is available now

vados-cosmonic: Oh thanks for the reminder, will use that

crates/wash-lib/src/build.rs
vados-cosmonic: after some discussion, the added weight to `wash-lib` doesn't seem to be a huge deal so will refactor to add that

connorsmith256 (dismissed):

connorsmith256 (approved):

PR #727, opened 2023-08-24
fix: docker build and push flow

vados-cosmonic:
## Feature or Problem

Based on the work in [this PR](https://github.com/wasmCloud/wash/pull/720), this PR attempts to update the `docker` build and release process to build for both arm64 and amd64. 

The build process partially mirrors `wadm` but breaks with it in using GHA matrix configuration for publishing the images on different platforms.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/.github/workflows/docker_check.yml
+++ b/.github/workflows/docker_check.yml
@@ -1,4 +1,4 @@
-name: Docker
+name: Docker build check
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/docker_check.yml
```diff
@@ -34,13 +131,14 @@ jobs:
...
+
+      - name: Docker build
         uses: docker/build-push-action@v4
         with:
           file: "Dockerfile"
-          target: release-alpine
-          tags: |
-            wasmcloud/wash:alpine
+          context: .
+          platforms: ${{ matrix.arch.docker_platform }}
```
brooksmtownsend: Am I missing this variable somewhere?

.github/workflows/docker_check.yml
```diff
@@ -7,16 +7,113 @@ on:
...
+            extension: ""
+            targetPath: "target/release/"
+          - os: "windows-latest"
+            arch: "amd64"
+            extension: ".exe"
+            targetPath: "target/release/"
+          - os: "macos-latest"
+            arch: "aarch64"
+            extension: ""
+            targetPath: "target/aarch64-apple-darwin/release/"
```
brooksmtownsend: Do we need to build these ones for the docker check? I know windows containers exist, but I wasn't thinking we had to add that quite yet (or as a part of this PR)

.github/workflows/release_docker.yml
```diff
@@ -7,34 +8,191 @@ on:
...
+            extension: ""
+            targetPath: "target/release/"
+          - os: "windows-latest"
+            arch: "amd64"
+            extension: ".exe"
+            targetPath: "target/release/"
+          - os: "macos-latest"
+            arch: "aarch64"
+            extension: ""
+            targetPath: "target/aarch64-apple-darwin/release/"
```
brooksmtownsend: Same comment re: target building

Dockerfile
```diff
@@ -1,57 +1,20 @@
...
 
-ENV RUSTFLAGS=-Ctarget-feature=-crt-static
-RUN cargo build --release
+# Copy application binary from disk
+COPY --chown=$USERNAME ${BIN} /usr/local/bin/wash
 
-FROM alpine as release-alpine
-WORKDIR /home/user
-RUN apk add --no-cache bash curl libgcc libressl-dev ca-certificates musl-dev
+USER $USERNAME
```
brooksmtownsend: I think building the image in this way will _only_ build it for a single docker architecture, where the wadm implementation conditionally copies either the arm64/amd64 binary based on the `uname -a` in the container, right?

.github/workflows/release_docker.yml
```diff
@@ -45,23 +203,23 @@ jobs:
...
-          tags: |
-            wasmcloud/wash:alpine
-          labels: ${{ steps.meta.outputs.labels }}
-      - name: Build and push - scratch
         id: docker_build
         uses: docker/build-push-action@v4
         with:
+          build-args: |
+            BIN=/artifacts/wash-${{ env.RELEASE_VERSION }}-${{ matrix.config.binary_suffix }}
+          context: .
```
brooksmtownsend: I think here we should pass two binaries, one for amd64 and one for arm64 so the container can properly buildx build a multi-platform image

brooksmtownsend (changes requested):
By and large this looks like the correct implementation / transfer of wadm things! Just left a few comments to make this a true multiplatform image and maybe a conditional compile or two we could remove

.github/workflows/release_docker.yml
vados-cosmonic: Sure I can remove these other targets -- cut it down to just the linux builds.

- [x] remove mac, windows builds

.github/workflows/docker_check.yml
vados-cosmonic: No I can certainly remove these -- similar to the `release_docker.yml`

Dockerfile
vados-cosmonic: Yep, but then we are using this `Dockerfile` across a matrix, so both images will get built, tagged and pushed (just separately).

.github/workflows/release_docker.yml
vados-cosmonic: yeah the attempt here was to enable building across the matrix so we could support containers on the other platforms (you can do this with the `docker manifest` command)

Since we're removing the other platforms for now I can revert this to the same way WADM does and we can revisit when adding support for the other platforms.

- [x] Use linux platform only multi-platform Dockerfile

Dockerfile
vados-cosmonic: yeah that's how it used to work -- but I realized that only worked for linux based platforms so wanted to build with the matrix and combine later -- but given that we're not going to support the other platforms yet, will revert this!

.github/workflows/docker_check.yml
vados-cosmonic: Ah this should be removed

- [x] remove unused `matrix.arch.docker_platform`

brooksmtownsend (dismissed):
Awesome stuff @vados-cosmonic ! Sorry, I introduced a merge conflict by merging #804 but if you can rebase then it's good to go 😄

vados-cosmonic:
Done @brooksmtownsend , thanks for the quick review!

brooksmtownsend (approved):

PR #731, opened 2023-08-29
fix: workspaces setup

yordis:
## Context

Since the setup is a workspace https://github.com/wasmCloud/wash/blob/a74b50297496578e5e6c0ee806304a3ff05cd073/packages/package.json#L7, the location of the `package-lock.json` should be at the root of the workspace. The file was created when I tried to do `npm install`


lachieh (approved):
LGTM

connorsmith256 (approved):

PR #743, opened 2023-08-30
yordis/feat add env var for websocket host

yordis:


brooksmtownsend (commented):
This looks fine to me, would like @lachieh's opinion just to make sure I'm not missing anything

packages/washboard/src/lattice/lattice-service.ts
```diff
@@ -67,7 +67,7 @@ class LatticeService {
   private connection?: NatsConnection;
 
   private config = {
-    latticeUrl: 'ws://localhost:4001',
+    latticeUrl: import.meta.env.VITE_NATS_WEBSOCKET_URL,
```
lachieh: Let's include a default if the env is not set
```suggestion
    latticeUrl: import.meta.env.VITE_NATS_WEBSOCKET_URL ?? 'ws://localhost:4001',
```

lachieh (changes requested):

packages/washboard/src/lattice/lattice-service.ts
autodidaddict: You TypeScript people, always bragging about your `??` like you're so cool 😁

lachieh: I always love finding places to use the elvis operator `?:`, although in JS/TS it's `??=`

autodidaddict: Almost as cool as the Rust toilet bowl : `|_|()`

lachieh (approved):

connorsmith256 (approved):

PR #799, opened 2023-09-13
fix(sign): move generate key message to info log

### LINKED ISSUE #698: Omit/clarify "No keypair found" message generated when signing a compiled module
Several users have expressed confusion at the following message that wash emits when generating a new keypair:
```
{"keygen":"true","path":"/Users/user/.wash/keys/derp_module.nk","status":"No keypair found"}
```

This is an informational/debug message that shouldn't be emitted by default. Or, if we think it _should_, we should replace the serialized JSON — which tends to look "scary" — with a very clear message. Something like:

> "Info: No existing keypair for module found. Wash generated a new one at /path/to/key"

I'm open to different wording/solutions. I think it would be easy to improve the status quo

brooksmtownsend:

## Feature or Problem
This PR moves the `println` inside of wash-lib to an `info` level log that can be controlled by the user. This will avoid printing the message about generating keypairs to stdout every time, and it also adjusts the messaging there to make the log sound less error-ey

## Related Issues
Fixes #698

## Release Information
wash-lib v0.10.2

## Consumer Impact
Consumers of wash-lib no longer need to address rogue generate messages on stdout.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4345,7 +4345,7 @@ dependencies = [
 
 [[package]]
 name = "wash-lib"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "anyhow",
  "async-compression",
--- a/crates/wash-lib/Cargo.toml
+++ b/crates/wash-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wash-lib"
-version = "0.10.1"
+version = "0.10.2"
 authors = ["wasmCloud Team"]
 categories = ["wasm", "wasmcloud"]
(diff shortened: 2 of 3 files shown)


connorsmith256 (approved):
❤️

PR #844, opened 2023-10-02
Added par support to wash-lib

aish-where-ya:
## Feature or Problem
Feat: Move `wash par` to `wash-lib`

## Release Information
next

## Consumer Impact
Internal code refactoring.
No impact to existing functionality

## Testing
unit and manual tests

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Unit Test(s)
NA

### Acceptance or Integration
NA

### Manual Verification
manually verified

### DIFF
--- a/crates/wash-lib/src/cli/par.rs
+++ b/crates/wash-lib/src/cli/par.rs
@@ -42,7 +42,7 @@ pub async fn create_provider_archive(
     Ok(par)
 }
 
-pub async fn insert_provider_archive(
+pub async fn insert_provider_binary(
     arch: String,
     binary_bytes: &[u8],
     mut par: ProviderArchive,
--- a/crates/wash-lib/src/cli/par.rs
+++ b/crates/wash-lib/src/cli/par.rs
@@ -1,11 +1,6 @@
-use std::{fs::File, io::prelude::*, path::PathBuf};
-
-use crate::cli::{extract_keypair, OutputKind};
-use anyhow::{anyhow, bail, Context, Result};
-use nkeys::KeyPairType;
+use anyhow::{anyhow, Context, Result};
 use provider_archive::ProviderArchive;
-
-const GZIP_MAGIC: [u8; 2] = [0x1f, 0x8b];
+use std::path::PathBuf;
 
 pub struct ParCreateArgs {
     pub capid: String,
@@ -13,86 +8,27 @@ pub struct ParCreateArgs {
     pub revision: Option<i32>,
     pub version: Option<String>,
     pub schema: Option<PathBuf>,
-    pub issuer: Option<String>,
-    pub subject: Option<String>,
     pub name: String,
-    pub directory: Option<PathBuf>,
     pub arch: String,
-    pub binary: String,
-    pub destination: Option<String>,
-    pub compress: bool,
-    pub disable_keygen: bool,
-}
-pub struct ParInsertArgs {
-    pub archive: String,
-    pub arch: String,
-    pub binary: String,
-    pub directory: Option<PathBuf>,
-    pub issuer: Option<String>,
-    pub subject: Option<String>,
-    pub disable_keygen: bool,
 }
 
-#[allow(clippy::too_many_arguments)]
-pub async fn handle_par_create(
+pub async fn create_provider_archive(
     ParCreateArgs {
-        // claims related options
         capid,
         vendor,
         revision,
         version,
         schema,
-        issuer,
-        subject,
         name,
-        // par related options
-        directory,
         arch,
-        binary,
-        destination,
-        compress,
-        disable_keygen,
     }: ParCreateArgs,
-    output_kind: OutputKind,
-) -> Result<String> {
+    binary_bytes: &[u8],
+) -> Result<ProviderArchive> {
     let mut par = ProviderArchive::new(&capid, &name, &vendor, revision, version);
 
-    let mut f = File::open(binary.clone())?;
-    let mut lib = Vec::new();
-    f.read_to_end(&mut lib)?;
+    par.add_library(&arch, binary_bytes)
+        .map_err(convert_error)?;
 
-    let issuer = extract_keypair(
-        issuer,
-        Some(binary.clone()),
-        directory.clone(),
-        KeyPairType::Account,
-        disable_keygen,
-        output_kind,
-    )?;
-    let subject = extract_keypair(
-        subject,
-        Some(binary.clone()),
-        directory,
-        KeyPairType::Service,
-        disable_keygen,
-        output_kind,
-    )?;
-
-    par.add_library(&arch, &lib).map_err(convert_error)?;
-
-    let extension = if compress { ".par.gz" } else { ".par" };
-    let outfile = match destination {
-        Some(path) => path,
-        None => format!(
-            "{}{}",
-            PathBuf::from(binary.clone())
-                .file_stem()
-                .unwrap()
-                .to_str()
-                .unwrap(),
-            extension
-        ),
-    };
     if let Some(ref schema) = schema {
         let bytes = std::fs::read(schema)?;
         par.set_schema(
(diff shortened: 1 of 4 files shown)


crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub binary: String,
+    pub directory: Option<PathBuf>,
+    pub issuer: Option<String>,
+    pub subject: Option<String>,
+    pub disable_keygen: bool,
+}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn handle_par_create(
+    ParCreateArgs {
```
brooksmtownsend: Since this is in the library I like the idea of calling this something more human-parseable
```suggestion
#[allow(clippy::too_many_arguments)]
pub async fn create_provider_archive(
    ParCreateArgs {
```

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+
+pub struct ParCreateArgs {
+    pub capid: String,
+    pub vendor: String,
+    pub revision: Option<i32>,
+    pub version: Option<String>,
+    pub schema: Option<PathBuf>,
+    pub issuer: Option<String>,
+    pub subject: Option<String>,
+    pub name: String,
```
brooksmtownsend: In the future I think we could consider taking these fields out and accepting a Claims struct with provider information in it and then archive options... but there's just so many options for this command I'm fine with the way this is now 👍🏻

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+        compress,
+        disable_keygen,
+    }: ParCreateArgs,
+    output_kind: OutputKind,
+) -> Result<String> {
+    let mut par = ProviderArchive::new(&capid, &name, &vendor, revision, version);
+
+    let mut f = File::open(binary.clone())?;
+    let mut lib = Vec::new();
+    f.read_to_end(&mut lib)?;
```
brooksmtownsend: Instead of opening the file here, let's have the user supply the binary as a `&[u8]` so the CLI can deal with opening the files and the library can just deal with the bytes themself

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+        output_kind,
+    )?;
+    let subject = extract_keypair(
+        subject,
+        Some(binary.clone()),
+        directory,
+        KeyPairType::Service,
+        disable_keygen,
+        output_kind,
+    )?;
```
brooksmtownsend: Same comment here re: let's just take the issuer/subject as function arguments and let the CLI load the key

brooksmtownsend: If making this refactor lets you get rid of more arguments, then that's great too!

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+    par.write(&outfile, &issuer, &subject, compress)
+        .await
+        .map_err(|e| anyhow!("{}", e))
+        .with_context(|| {
+            format!(
+                "Error writing PAR. Please ensure directory {:?} exists",
+                PathBuf::from(outfile.clone()).parent().unwrap(),
+            )
+        })?;
+    Ok(outfile)
```
brooksmtownsend: Instead of writing to the file directly, I think it would be nice to simply return the `ProviderArchive` struct and let the CLI user (in our case, wash) write the provider archive out to a file

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+        None => format!(
+            "{}{}",
+            PathBuf::from(binary.clone())
+                .file_stem()
+                .unwrap()
+                .to_str()
+                .unwrap(),
+            extension
+        ),
+    };
```
brooksmtownsend: If you can refactor this function to return the ProviderArchive struct, then you shouldn't need destination or outfile anymore (wash itself can handle those)

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+    }: ParInsertArgs,
+    output_kind: OutputKind,
+) -> Result<()> {
+    let mut buf = Vec::new();
+    let mut f = File::open(archive.clone())?;
+    f.read_to_end(&mut buf)?;
+
+    let mut par = ProviderArchive::try_load(&buf)
+        .await
+        .map_err(convert_error)?;
```
brooksmtownsend: I think this function would be great accepting the mutable provider archive instead of the archive file

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+        output_kind,
+    )?;
+    let subject = extract_keypair(
+        subject,
+        Some(binary.clone().to_owned()),
+        directory,
+        KeyPairType::Service,
+        disable_keygen,
+        output_kind,
+    )?;
```
brooksmtownsend: Same comment re: accepting issuer / subject as arguments

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    let mut f = File::open(binary.clone())?;
+    let mut lib = Vec::new();
+    f.read_to_end(&mut lib)?;
+
```
brooksmtownsend: If we accept this binary as a `&[u8]` as well, we can simply insert it into the provider archive and then finish

brooksmtownsend (changes requested):
One nit, one comment that may not be applicable anymore, but _mostly_ ergonomic changes that feel like they would make the functions here a whole lot simpler if we pushed a little responsibility on the CLI-side. Essentially, I think finding and loading keypairs and files should be the responsibility of the CLI, and the library can just properly take that information and turn it into the appropriate `ProviderArchive` structs that the CLI can then write out

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,179 @@
...
+        .with_context(|| {
+            format!(
+                "Error writing PAR. Please ensure directory {:?} exists",
+                PathBuf::from(outfile.clone()).parent().unwrap(),
+            )
+        })?;
+    Ok(outfile)
+}
+
+pub async fn handle_par_insert(
```
brooksmtownsend: With the refactors we can rename this too
```suggestion
pub async fn insert_provider_binary(
```

aish-where-ya:
I think I managed to address all the review comments here. Pulling stuff out from the library into the CLI definitely helped in reducing some of the function arguments and also removing the function argument struct for the `insert_provider_binary`, thus reducing a lot of clutter.

crates/wash-lib/src/cli/par.rs
```diff
@@ -0,0 +1,59 @@
...
+                .with_context(|| "Unable to parse JSON from file contents".to_string())?,
+        )
+        .map_err(convert_error)
+        .with_context(|| format!("Error parsing JSON schema from file '{:?}'", schema))?;
+    }
+
+    Ok(par)
+}
+
+pub async fn insert_provider_archive(
```
brooksmtownsend: I think this should be named `insert_provider_binary` to be specific

brooksmtownsend (commented):
Looks really clean 👍🏻 One tiny request to rename function but looks good to me

crates/wash-lib/src/cli/par.rs
aish-where-ya: oopsies

aish-where-ya: done

brooksmtownsend (approved):

Issue #512, opened 2023-04-06
Make Tinygo Target Configurable


jordan-rash:
Would like to make the below line configurable.  tinygo will take either a `wasm` or `wasi` target.  The `wasm` target is what we have always used, but it does pull in all the JS stuff.  I am just starting the test the `wasi` target, so I don't know what we want to make the default at this point.  Probably `wasm` to reduce confusion? 🤷🏼 

https://github.com/wasmCloud/wash/blob/f7b34341e177ce72a3a50366e476b752a9e9d54d/crates/wash-lib/src/build.rs#L198

jordan-rash:
cc: @brooksmtownsend && @mattwilkinsonn 

I assume this is a "Good First Issue", but I'm not sure if what discussions need to be had on defaults

brooksmtownsend:
Yeah @jordan-rash I think we want to make use of https://github.com/wasmCloud/wash/blob/f7b34341e177ce72a3a50366e476b752a9e9d54d/crates/wash-lib/src/parser/mod.rs#L50 instead of hardcoded `wasm` there!

edit: this will just go in the wasmcloud.toml
```toml
[actor]
wasm_target = "wasm|wasi"
```

jordan-rash:
I noticed that the tinygo build function only takes in the common_config and the tinygo_config, while the rust build function takes in the actor_config (where that setting lives) as a third parameter.  Is there a reason that the actor_config was left out?  

Also, it looks like that setting defaults to wasm32_unknown_unknown.  If we were to use it, we would have to make the default conditional on the language.  I wonder if it would be easier to move the target setting to the rust_config / tinygo_config sections??

(Sorry for the lack of formatting, I'm on my phone)

brooksmtownsend:
Looks like you're probably right in both cases, we should also put in actor config and let the default be dictated by the language or overrided

jordan-rash:
Also need to address the scheduler flag

Right now, most the automation I use compiles like so 
`tinygo build -target=wasi -scheduler=none -o output_path .`

PR #848, opened 2023-10-03
refactor: embed component metadata

vados-cosmonic:
## Feature or Problem

This refactors the existing branch implementing Go WASI p2 components with code to embed component metadata.

## Related Issues

https://github.com/wasmCloud/wash/issues/512

## Consumer Impact

Consumers will be able to build WASI preview2 components from Go with `wash build`

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### Manual Verification

Verified manually locally w/ the [wasmcon golang codebase](https://github.com/cosmonic/wasmcon-workshop/tree/main/hello-cosmo-go)


crates/wash-lib/src/build.rs
```diff
@@ -95,14 +97,21 @@ pub fn build_actor(
...
+                .as_ref()
+                .context("`wit_world` must be specified in wasmcloud.toml for creating preview1 or preview2 components")?,
+            &actor_wasm_path,
+            &actor_wasm_path,
+        )?;
+    };
 
-    // If the actor has been configured as WASI Preview2, adapt it
-    if actor_config.wasm_target == WasmTarget::WasiPreview2
-        || actor_config.wasm_target == WasmTarget::TinyGoWasiPreview1
```
ricochet, line 104: So essentially if it has not been configured for Preview2 we do not adapt. This avoids the situation @jordan-rash pointed out where you need to target wasi to build wasm-only (no js glue) tiny-go modules.

crates/wash-lib/src/build.rs
```diff
@@ -330,6 +339,76 @@ pub(crate) fn get_wasi_preview2_adapter_bytes(config: &ActorConfig) -> Result<Ve
     Ok(wasmcloud_component_adapters::WASI_PREVIEW1_REACTOR_COMPONENT_ADAPTER.into())
 }
 
+/// Embed the WIT inside a given WASM binary, in-place.
```
ricochet: OK this is the nittiest of nits:
https://www.w3.org/TR/wasm-core/intro/introduction.html#wasm

A contraction of “WebAssembly”, not an acronym, hence not using all-caps.

crates/wash-lib/src/build.rs
vados-cosmonic: Yup, that's absolutely right -- so right now, while we embed component metadata whether you target p1 or p2, we only *adapt* if you target p2 (to take you from p1 to p2).

The `wasmcloud.toml` in the wasmcon demo when modified looks like this:

```toml
name = "hello-cosmo-go"
language = "tinygo"
type = "actor"
version = "0.1.0"

[actor]
claims = ["wasmcloud:httpserver", "wasmcloud:keyvalue", "wasmcloud:builtin:logging"]
wasm_target = "wasm32-wasi-preview2"
wit_world = "hello-cosmo"
```

ricochet (approved):

crates/wash-lib/src/build.rs
vados-cosmonic: Absolutely will fix, thank you, thanks for spotting this

- [x] remove WASM usage in favor of WebAssembly

crates/wash-lib/src/build.rs
```diff
@@ -330,6 +339,72 @@ pub(crate) fn get_wasi_preview2_adapter_bytes(config: &ActorConfig) -> Result<Ve
...
+            wit_dir.display()
+        );
+    };
+
+    // Resolve the WIT directory packages & worlds
+    let mut resolve = wit_parser::Resolve::default();
+    let (package_id, _paths) = resolve
+        .push_dir(&wit_dir)
+        .context("failed to add WIT deps directory")?;
+    log::info!("successfully loaded WIT @ [{}]", wit_dir.display());
```
connorsmith256, line 363: Nit: debug level? I don't think users need to see this every time they `wash build`

(Same with the other logs in this function)

connorsmith256 (changes requested):

crates/wash-lib/src/build.rs
vados-cosmonic: Ah so the default level is *above* info -- it's error (i.e. someone has to do `RUST_LOG=info` to see these). 

Also, I think info is the right level for these messages -- they don't touch on the really deep "how", and they mostly say when something happened successfully/happy path. 

In my mind debug messages generally cover the before, during and end of some operation (so we can debug) where as informational stuff just lets you know general outcomes from time to time.

connorsmith256: I don't think I realized until this moment that info wasn't the default level 😮 my bad!

PR #862, opened 2023-10-05
[FEAT] Update wash to use max_concurrent control interface

brooksmtownsend:
## Feature or Problem
This PR is updating `wash` to use the latest control interface that supports max scale instead of counts. I verified that this version of the control interface works great with wasmCloud 0.78.0 as the breaking changes there are either backwards compatible over-the-wire or no longer involve the host (e.g. querying links and claims from the bucket directly always).

I got a _little_ carried away and made a few more changes in this PR regarding the `ctl` command in this PR. This sets us up to be able to deprecate `ctl` entirely instead of keeping it stable for a subset of commands:
1. Added top level `wash scale actor` command to eventually replace `wash ctl scale actor`
2. Added top level `wash update actor` command to eventually replace `wash ctl update actor`
3. Removed `wash ctl apply manifest` entirely, as it's functionality that is replaced by wadm and uses a manifest format that's bespoke for this individual command.
The above changes were mostly moving code to `wash-lib` and then exposing another top level command, e.g. mostly refactoring.

## Related Issues
Depends on https://github.com/wasmCloud/control-interface-client/pull/58 (this PR's checks can be re-run after that merges)
RFC https://github.com/wasmCloud/wasmCloud/issues/696

## Release Information
wash-lib v0.12.0
wash-cli v0.21.0

## Consumer Impact
As designed, this version of `wash` will remain completely compatible with older versions of the host and work for newer released versions.

## Testing

Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows


ghost (changes requested):
Are you able to add integration tests for the new commands?

brooksmtownsend:
> Are you able to add integration tests for the new commands?

Good call, they aren't new commands but we should definitely have an integration test for it at least basic. I've tested with a few manual cases so I'll just turn that into an integration test :)

crates/wash-lib/src/actor.rs
```diff
@@ -51,6 +51,7 @@ pub async fn start_actor(
         .context("Failed to get lattice event channel")?;
 
     // Start the actor
+    #[allow(deprecated)]
     let ack = ctl_client
         .start_actor(host_id, actor_ref, count, None)
```
connorsmith256, line 56: Can we update this to use scale under the hood?

crates/wash-lib/src/actor.rs
```diff
@@ -133,7 +132,7 @@ pub async fn stop_actor(
         .map_err(boxed_err_to_anyhow)?;
 
     let ack = client
-        .stop_actor(host_id, actor_id, count, annotations)
+        .stop_actor(host_id, actor_id, annotations)
```
connorsmith256, line 135: Can we also update this to use scale under the hood?

src/up/config.rs
```diff
@@ -11,7 +11,7 @@ pub(crate) const DEFAULT_NATS_PORT: &str = "4222";
 // wadm configuration values
 pub(crate) const WADM_VERSION: &str = "v0.6.0";
 // wasmCloud configuration values, https://wasmcloud.dev/reference/host-runtime/host_configure/
-pub(crate) const WASMCLOUD_HOST_VERSION: &str = "v0.78.0-rc6";
+pub(crate) const WASMCLOUD_HOST_VERSION: &str = "v0.78.0";
```
connorsmith256, line 14: Should we be hesitant about bumping to 0.79.0-rc3?

connorsmith256 (commented):
Looks great, just a couple nits, and 👍 to updating tests

crates/wash-lib/src/actor.rs
brooksmtownsend: Oh yes, this was leftover from before scale Some(0) == stop. Will do

src/up/config.rs
brooksmtownsend: Don't believe so, but I wanted to first ensure all of this worked for the current version of the host and then go from there to 0.79

crates/wash-lib/src/actor.rs
brooksmtownsend: I wanted to leave this as `start` so that previous versions of the host would maintain the same behavior, since the host has enough information to "start" more actors

I could alternatively here query the hosts inventory and increment that number before sending scale

crates/wash-lib/src/actor.rs
brooksmtownsend: Actually, I don't know if I want to do this quite yet. Scale accepts an actor reference but stop accepts an actor ID, so it would be a decently breaking change to do this

brooksmtownsend: Ideally what I would like to do is make it so that scale can take an actor ID instead of an image reference if scaling down but that introduces weird edge cases

crates/wash-lib/src/actor.rs
connorsmith256: Ah okay, replying to this and the other thread here: I guess I'm wondering what will happen when we actually remove support for start/stop from the control interface. At that point wash _has_ to change its logic. Are you saying we just wait a while until we can reasonably assume all hosts will be running with the new scale logic, and update the wash logic at that time?

brooksmtownsend: Yep that's my thought
1. wash works the same as before while hosts have old operations <-- you are here
2. wash removes old operations
3. host removes support for old operations

connorsmith256: Sounds good 👍

connorsmith256 (dismissed):

brooksmtownsend:
Failing tests because of a DNS error... will try again tomorrow

brooksmtownsend:
> Failing tests because of a DNS error... will try again tomorrow

Failing tests because I didn't exclude them from the nix suite :)

Cargo.toml
```diff
@@ -151,10 +151,10 @@ wadm = "0.6.0"
 walkdir = "2.4"
 wascap = "0.11.2"
 wasm-encoder = "0.33.2"
-wash-lib = { version = "0.11", path = "./crates/wash-lib" }
-wasmbus-rpc = "0.14"
 wasmcloud-component-adapters = { version = "0.2.4", default-features = false }
-wasmcloud-control-interface = "0.27"
+wash-lib = { version = "0.12", path = "./crates/wash-lib" }
+wasmbus-rpc = "0.15.0"
+wasmcloud-control-interface = "0.29"
```
connorsmith256: Can we bump to 0.30 or are there breaking changes?

connorsmith256 (dismissed):
Approved with a suggestion :)

connorsmith256 (approved):

Cargo.toml
brooksmtownsend: done!

brooksmtownsend:
@mattwilkinsonn added integration tests and ready for re-review

ghost (approved):

PR #867, opened 2023-10-05
fix(ui): update imports after package update and move to root

lachieh:
## Feature or Problem
After the svgr plugin got major version bump from [dependabot](https://github.com/wasmCloud/wash/pull/815), it broke the import style for svgs. This resolves the build by changing the import path.

NPM workspaces have also been causing issues, so this also moves the washboard up to the root and removes the use of NPM workspaces. 

## Related Issues
None

## Release Information
`next`

## Consumer Impact
Since the static assets were not rebuilt, this is not an issue for any current version of `wash` but it would have been a blocker for any contributions to the washboard ui.

## Testing
Tested build both ui and wash builds locally and confirmed working.


connorsmith256 (approved):

connorsmith256:
@lachieh LGTM but need DCO :)

PR #897, opened 2023-10-16
fix: downgrade wit-parser

ricochet:
Without this change, `cargo install` may fail.

wit-component 0.14.6 has been yanked. This is the version I had.

There isn't a combination of the latest released
versions that works without pulling conflicting versions
of wit-parser. That's why we go back a version of wit-parser.

Signed-off-by: Bailey Hayes <behayes2@gmail.com>


ricochet:
Fallout from https://github.com/bytecodealliance/wasm-tools/pull/1250

brooksmtownsend (commented):
This looks good to me, if you'd like we can go ahead and add a patch to this for wash-lib and wash-cli so we can yank the non-compiling version

brooksmtownsend (approved):
Works with a cargo install

connorsmith256 (approved):
tested e2e by building a tinygo component from this branch and ensuring it could instantiate with version v0.79.0 of the host 👍

Issue #620, opened 2023-06-21
[BUG] wash build appears to ignore key directory setting from wasmcloud.toml


autodidaddict:
**Describe the bug**
When I supply the `key_directory` field (and for giggles, I also tried `keys_directory`) in the `[actor]` section, `wash build` will still either reuse or generate a new key in `~/.wash/keys`. It appears to the end user that wash build is ignoring this field in the toml file.

**To Reproduce**
1. Put a key in `./keys/my_actor_module.nk` where `my_actor` corresponds to the project name in `Cargo.toml`
2. Add `key_directory=./.keys` to the `wasmcloud.toml` file
3. Run `wash build`, you will see that it has generated a new key and _not_ reused the one from the `keys` directory.

**Expected behavior**
I expect it to honor the `key_directory` setting

**Screenshots**
This is the shell output, note that it's creating a new key:
```
{"keygen":"true","path":"/Users/kevin/.wash/keys/foo_aggregate_module.nk","status":"No keypair found"}

Actor built and signed and can be found at "/Users/kevin/code/(blah blah)"
```

**Environment**
 - OS: macOS
 - Shell zsh
 - wash Version 0.18.0

connorsmith256:
From what I can tell, `key_directory` is parsed from the actor config, but isn't used anywhere :( 

https://github.com/wasmCloud/wash/blob/67d9b91085ed74c64ef8bc56de48e54e4a760a64/crates/wash-lib/src/parser/mod.rs#L48

The `--keys-directory` option on `wash build` does appear to be used, FWIW

connorsmith256:
Closed by #926

PR #926, opened 2023-10-19
fix: `wash build` key directory setting behavior

ahmedtadde:
## Feature or Problem
Despite setting `key_directory` field in the `[actor]` section of `wasmcloud.toml`, `wash build` will still either reuse or generate a new key in `~/.wash/keys`. It appears to the end user that `wash build` command is ignoring the provided configuration.

## Related Issues
#620 

## Release Information
v0.21.1

## Consumer Impact
improved correctness of the `wash` cli; behaves as expected.

## Testing
Built on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [ ] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [x] aarch64-darwin
- [ ] x86_64-windows

### Acceptance or Integration
added test cases to ensure key directory configuration is respected and adheres to common configuration precedence rule `config file field < env var < cli arg`

### Manual Verification
ran `wash build` on `hello` actor template repo with different key directory configuration.

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -15,7 +15,7 @@ use wit_component::{ComponentEncoder, StringEncoding};
 
 use crate::{
     cli::{
-        claims::{sign_file, ActorMetadata, SignCommand},
+        claims::{sign_file, ActorMetadata, GenerateCommon, SignCommand},
         OutputKind,
     },
     parser::{
@@ -181,6 +181,11 @@ fn sign_actor_wasm(
             call_alias: actor_config.call_alias.clone(),
             issuer: signing_config.issuer,
             subject: signing_config.subject,
+            common: GenerateCommon {
+                disable_keygen: signing_config.disable_keygen,
+                directory: signing_config.keys_directory,
+                ..Default::default()
+            },
             tags,
             ..Default::default()
         },
--- a/src/build.rs
+++ b/src/build.rs
@@ -53,12 +53,15 @@ pub(crate) async fn handle_command(command: BuildCommand) -> Result<CommandOutpu
     let config = get_config(command.config_path, Some(true))?;
 
     match config.project_type {
-        TypeConfig::Actor(ref _actor_config) => {
+        TypeConfig::Actor(ref actor_config) => {
             let sign_config = if command.build_only {
                 None
             } else {
                 Some(SignConfig {
-                    keys_directory: command.keys_directory,
+                    keys_directory: command
+                        .keys_directory
+                        .clone()
+                        .or(Some(actor_config.key_directory.to_path_buf())),
                     issuer: command.issuer,
                     subject: command.subject,
                     disable_keygen: command.disable_keygen,
@@ -69,7 +72,7 @@ pub(crate) async fn handle_command(command: BuildCommand) -> Result<CommandOutpu
 
             let json_output = HashMap::from([
                 ("actor_path".to_string(), json!(actor_path)),
-                ("signed".to_string(), json!(command.build_only)),
+                ("signed".to_string(), json!(!command.build_only)),
             ]);
             Ok(CommandOutput::new(
                 if command.build_only {
--- a/tests/integration_build.rs
+++ b/tests/integration_build.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, Result};
-use std::fs::File;
+use std::{env, fs::File};
 use tokio::process::Command;
 
 mod common;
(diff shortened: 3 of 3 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -154,6 +154,13 @@ pub fn build_actor(
                 ..Default::default()
             },
         };
+
+        if config.keys_directory.is_some() {
+            sign_options.metadata.common.directory = config.keys_directory
+        }
+
+        sign_options.metadata.common.disable_keygen = config.disable_keygen;
```
brooksmtownsend: nit: consider moving this part up into when you create the SignCommand struct since it's not a conditional

src/build.rs
```diff
@@ -53,12 +53,15 @@ pub(crate) async fn handle_command(command: BuildCommand) -> Result<CommandOutpu
...
+        TypeConfig::Actor(ref actor_config) => {
             let sign_config = if command.build_only {
                 None
             } else {
                 Some(SignConfig {
-                    keys_directory: command.keys_directory,
+                    keys_directory: match &command.keys_directory {
+                        Some(dir) => Some(dir.clone()),
+                        None => Some((*actor_config.key_directory).to_path_buf()),
+                    },
```
brooksmtownsend: nit: this might be expressed simpler using the Option API, e.g.
```
command.keys_directory.clone().or_else(|| actor_config.key_directory.to_path_buf())
```
This will fall back to the actor config key dir if the command key dir isn't supplied.

brooksmtownsend: You may need to double check me on types here since I'm just winging it in GitHub

brooksmtownsend (dismissed):
Big fan of the new integration test! Added a few nits to consider when you look at the failing test but they are just nits.

crates/wash-lib/src/build.rs
```diff
@@ -151,9 +151,18 @@ pub fn build_actor(
...
+                common: GenerateCommon {
+                    disable_keygen: config.disable_keygen,
+                    ..Default::default()
+                },
                 ..Default::default()
             },
         };
+
+        if config.keys_directory.is_some() {
+            sign_options.metadata.common.directory = config.keys_directory
```
connorsmith256: Nit: could this be moved into the struct definition, to avoid the need for `mut`?
```
common: GenerateCommon {
    disable_keygen: config.disable_keygen,
    directory: config.keys_directory
    ..Default::default()
},
```

connorsmith256 (commented):
LGTM! I just have one nit

crates/wash-lib/src/build.rs
ahmedtadde: My understanding is that we want to do the override only when there is an actual value to override with. So, when it is None, we don't use that to potentially override a Some(...) value already set on the common object/strict.

Let me know if that made sense. And , whether my understanding is wrong.

connorsmith256: I might be missing something, but it looks like the `Default::default()` on `common` will set `directory` to `None` by default. My suggested change is to set `directory` to whatever `config.keys_directory` is. If the latter is `None`, then there's no change, and if it's `Some`, this line should have the same impact as your if statement

ahmedtadde: So, that was my original plan. But then I wasn't sure if one can assume that the directory under common would always be None in this case. I am not on my computer; I remember tracking down the struct for GenerateCommon and seeing that it is annotated with clap and the directory field uses the WASH_KEYS as the default fallback value. I think I (wrongly) assumed that the Default::default() involves using WASH_KEYS if it set and use None otherwise.


Anyway, sounds like I got it wrong on that front. I will make the suggested change once I come around to my computer.

connorsmith256: Ah, I see. Here's the bit you're referencing:
```
#[derive(Debug, Clone, Parser, Serialize, Deserialize, Default)]
pub struct GenerateCommon {
    /// Location of key files for signing. Defaults to $WASH_KEYS ($HOME/.wash/keys)
    #[clap(long = "directory", env = "WASH_KEYS", hide_env_values = true)]
    pub directory: Option<PathBuf>,
```
The default value derived from `Default` is `None`. When parsing with clap, clap will look for a `--directory` arg, or a `WASH_KEYS` flag. However, in this bit of code we're constructing this struct manually, not parsing with clap. Hope that helps!

ahmedtadde: [made the change](https://github.com/wasmCloud/wash/pull/926/commits/ea66be6b30c93224f17df190c9b9d0927392d83d)


admittedly, i am still a rust amateur and I am susceptible to do a derp like that. but now I know. thanks for the assist on this.

connorsmith256:
@ahmedtadde LGTM! but there's a minor merge conflict 🙏

ahmedtadde:
> @ahmedtadde LGTM! but there's a minor merge conflict 🙏

branch is now rebased with upstream/main.

connorsmith256 (approved):

PR #937, opened 2023-10-24
feat: mark components built with wash as experimental

vados-cosmonic:
## Feature or Problem

As the component model and WASI are still maturing, the components-first codebases built with `wash` should reflect the experimental nature of support to related tooling.

This commit marks both components as experimental at two levels -- a custom section in the Wasm metadata (as a custom section) and as a tag on the signed wasmCloud actor that is produced.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -24,9 +24,15 @@ use crate::{
     },
 };
 
+/// This tag indicates that a Wasm module uses experimental features of wasmCloud
+/// and/or the surrounding ecosystem.
+///
+/// This tag is normally embedded in a Wasm module as a custom section
+const WASMCLOUD_WASM_TAG_EXPERIMENTAL: &str = "wasmcloud.com/experimental";
+
 /// Configuration for signing an artifact (actor or provider) including issuer and subject key, the path to where keys can be found, and an option to
 /// disable automatic key generation if keys cannot be found.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct SignConfig {
     /// Location of key files for signing
     pub keys_directory: Option<PathBuf>,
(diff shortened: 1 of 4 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -131,8 +137,33 @@ pub fn build_actor(
...
+}
+
+/// Sign actor wasm configuration, if necessary
+fn sign_actor_wasm(
+    common_config: &CommonConfig,
+    actor_config: &ActorConfig,
+    signing_config: Option<SignConfig>,
+    actor_wasm_path: impl AsRef<Path>,
+) -> Result<PathBuf> {
     if let Some(config) = signing_config {
```
connorsmith256: Nit: I think this if statement should be outside this function (in `build_actor`), and `sign_actor_wasm` should take a `SignConfig` (not an `Option`). Kinda weird to call a function called `sign` but then conditionally not sign

crates/wash-lib/src/build.rs
```diff
@@ -380,14 +411,25 @@ fn embed_wasm_component_metadata(
...
+        },
+        wasm_encoder::CustomSection {
+            name: WASMCLOUD_WASM_TAG_EXPERIMENTAL.into(),
+            data: Cow::Borrowed(b"true"),
+        },
+    ];
+    for section in custom_sections {
+        wasm_bytes.push(section.id());
+        section.encode(&mut wasm_bytes);
+        log::info!(
```
connorsmith256: Just noticed this, but I feel like it should be debug instead of info

connorsmith256 (dismissed):
LGTM! Love the new tests. Just left a couple nits

crates/wash-lib/src/build.rs
vados-cosmonic: Yep, I think you're right

- [x] change log level to debug

crates/wash-lib/src/build.rs
vados-cosmonic: Ah you're right, will change this. 

- [x] remove option for signing config

vados-cosmonic:
@connorsmith256 thanks for the review, changes are in if you'd like to take a look!

connorsmith256 (approved):

PR #943, opened 2023-10-25
feat: generate golang code during wash build

vados-cosmonic:
## Feature or Problem

Components-first golang actors require that `go generate` be run, with wit-bindgen as the directive. While this is easy to do, it makes the build workflow (i.e. calling `wash build`) require more steps.

This commit adds support for running the golang wit-bindgen functionality as a part of `wash build`, so that users don't have to call wit-bindgen themselves, or add stanzas for generate to their code.

In the future, examples can be created that assume that the 'generated' folder is present, and import code as necssary.

## Testing

Built on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

Tested on platform(s)
- [x] x86_64-linux
- [ ] aarch64-linux
- [ ] x86_64-darwin
- [ ] aarch64-darwin
- [ ] x86_64-windows

### DIFF
--- a/crates/wash-lib/src/build.rs
+++ b/crates/wash-lib/src/build.rs
@@ -11,7 +11,10 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use wasm_encoder::{Encode, Section};
+use wit_bindgen_core::Files;
+use wit_bindgen_go::Opts as WitBindgenGoOpts;
 use wit_component::{ComponentEncoder, StringEncoding};
+use wit_parser::{Resolve, WorldId};
 
 use crate::{
     cli::{
@@ -110,7 +113,7 @@ pub fn build_actor(
             actor_config
                 .wit_world
                 .as_ref()
-                .context("`wit_world` must be specified in wasmcloud.toml for creating preview1 or preview2 components")?,
+                .context("missing `wit_world` in wasmcloud.toml ([actor] section) for creating preview1 or preview2 components")?,
             &actor_wasm_path,
             &actor_wasm_path,
         )?;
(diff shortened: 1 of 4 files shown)


crates/wash-lib/src/build.rs
```diff
@@ -316,6 +341,65 @@ fn build_tinygo_actor(
     Ok(common_config.path.join(wasm_file))
 }
 
+/// The folder that golang bindgen code will be placed in, normally
+/// from the top level golang project directory
+const GOLANG_BINDGEN_FOLDER_NAME: &str = "generated";
```
ricochet: `gen` is the convention I most often see. 

For example:
https://protobuf.dev/reference/go/go-generated/

crates/wash-lib/src/build.rs
```diff
@@ -366,6 +450,27 @@ pub(crate) fn get_wasi_preview2_adapter_bytes(config: &ActorConfig) -> Result<Ve
...
 }
 
+/// Build a [`wit_parser::Resolve`] from a provided directory
+/// and select a given world
+fn convert_wit_dir_to_world(
+    dir: impl AsRef<Path>,
+    world: impl AsRef<str>,
+) -> Result<(Resolve, WorldId)> {
+    // Resolve the WIT directory packages & worlds
+    let mut resolve = wit_parser::Resolve::default();
```
ricochet, line 460: Reading through the logic, I *think* we're OK but it would help to see a unit test where there are multiple world definitions.

ricochet: This is the case I'm thinking about: https://github.com/WebAssembly/wasi-cli/blob/main/wit/command.wit

command world includes the reactor world.

crates/wash-lib/src/build.rs
vados-cosmonic: Sure I can change this to `gen`

- [x] change `generated` to `gen`

crates/wash-lib/src/build.rs
vados-cosmonic: Done! Add a test that matches this expectation, and uses `include`

ricochet (approved):

PR #955, opened 2023-10-28
feat: add platform and arch bin releasing

yordis:


.github/workflows/release.yml
```diff
@@ -55,7 +133,13 @@ jobs:
         id: create_release
         uses: softprops/action-gh-release@v1
         with:
-          files: release/windows/wash.exe
```
yordis, line 58: Is `x86_64-pc-windows-gnu` binary the same as the one from `windows_build`?

.github/workflows/release.yml
```diff
@@ -30,8 +31,8 @@ jobs:
           args: --all-features
 
   windows_build:
```
yordis: is this build needed now?

connorsmith256:
Hey @yordis, as of today, wash has been moved to the [wasmCloud repo](https://github.com/wasmCloud/wasmCloud/tree/main/crates/wash-cli) 🎉 

I think the [workflow there](https://github.com/wasmCloud/wasmCloud/blob/main/.github/workflows/wasmcloud.yml) takes care of building wash for all platforms as well

If there are more improvements you want to make, let's open a PR against the main repo. I'm working on updating this repo to be a public archive :)

yordis:
I love that!

